PR #7, opened 2018-08-20
add part and section to request header

zemirco:
Hey,

I also started working on an OPC-UA implementation in Go. Then I saw your work. You're further ahead and so I decided to contribute instead of starting all over again. You're doing great so far!

This is my first PR. It's just to get started. I'd like to add more references from the official specification. Since the spec is huge and quite difficult to navigate I thought having direct links from the code would make it much easier to jump between source code and specification.

Is that OK for you? Then I would add more parts and sections to the other types.

Just out of interest? Is OPC UA and Go part of your job or are you working on this project in your spare time?

### DIFF
--- a/services/request-header.go
+++ b/services/request-header.go
@@ -14,6 +14,8 @@ import (
 )
 
 // RequestHeader represents a Request Header in each services.
+// Part 4
+// 7.28
 type RequestHeader struct {
 	AuthenticationToken datatypes.NodeID
 	Timestamp           time.Time


services/request-header.go
```diff
@@ -14,6 +14,8 @@ import (
 )
 
 // RequestHeader represents a Request Header in each services.
+// Part 4
```
wmnsk, line 17: In this way, GoDoc won't be so good in appearance point of view.

image

How do you think writing like this instead?

```go
// RequestHeader represents a Request Header in each services.
//
// Specicication: Part 4, 7.28
type RequestHeader struct {
```

image

zemirco:
Good point! I'm gonna close my current pull requests and open a new one. I messed up locally.

PR #10, opened 2018-08-20
add references to specification

zemirco:


wmnsk:
Looks nice :+1: 
Please keep on working on other parts!

zemirco:
That should be enough for now.

I would suggest adding more references in separate PRs. This one will get too big otherwise and will be open for too long.

PR #18, opened 2018-08-22
add wireshark format

zemirco:
Hey,

I've built myself a little helper function to compare the messages we're building to the messages you can see in Wireshark. It's quite useful for debugging.

For example at the beginning I was using the Prosys OPC UA Client https://downloads.prosysopc.com/opc-ua-client-downloads.php to connect to a server. I used Wireshark to record the messages and then started reverse engineering them.

It's also nice to check that the bytes in our messages are correct.

Here is a screenshot from Wireshark

screenshot from 2018-08-22 11-16-37

And here is the printed message using the new utility function.

screenshot from 2018-08-22 11-27-13

With the space at the right place (after 8 bytes) and the line break after 16 bytes it's very easy to see invalid bytes or different message lengths.

I simply added one line to the example.

```go
// Setup Hello
hello, err := uacp.NewHello(0, uint32(*sndBuf), uint32(*rcvBuf), uint32(*maxMsg), *url).Serialize()
if err != nil {
	log.Fatalf("Failed to serialize Hello: %s", err)
}

// new stuff here
fmt.Println(utils.Wireshark(0, hello))
```


wmnsk:
It looks so helpful.
Let me review this feature precisely tomorrow.

zemirco:
Sure, take your time! I added some comments to make it easier to understand.

wmnsk:
Thanks. It looks really good to me.

PR #19, opened 2018-08-22
add data type qualified name

zemirco:
I'm currently working on the `ReadRequest` service. It needs the type `ReadValueID`. This type needs the `QualifiedName` type.

Here is the first PR implementing the `QualifiedName` type. I tried to implement the same methods like you did with other types.

Let me know if you want anything changed / added / fixed.

### DIFF
--- a/datatypes/qualified-name.go
+++ b/datatypes/qualified-name.go
@@ -1,0 +1,69 @@
+package datatypes
+
+import "encoding/binary"
+
+// QualifiedName contains a qualified name. It is, for example, used as BrowseName.
+// The name part of the QualifiedName is restricted to 512 characters.
+//
+// Specification: Part 3, 8.3
+type QualifiedName struct {
+	NamespaceIndex uint16
+	Name           *String
+}
+
+// NewQualifiedName creates a new QualifiedName.
+func NewQualifiedName(index uint16, name string) *QualifiedName {
+	value := []byte(name)
+	length := -1
+
+	if len(value) != 0 {
+		length = len(value)
+	}
+
+	q := &QualifiedName{
+		NamespaceIndex: index,
+		Name: &String{
+			Value:  value,
+			Length: int32(length),
+		},
+	}
+	return q
+}
+
+// DecodeQualifiedName decodes given bytes into QualifiedName.
+func DecodeQualifiedName(b []byte) (*QualifiedName, error) {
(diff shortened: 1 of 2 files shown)


zemirco:
You're absolutely right. I also think twice before adding some third-party dependency :+1:.

However I think https://github.com/google/go-cmp is worth to be evaluated. Here is why I think it is great.

1. `reflect.DeepEqual` works indeed similarly. However when there is something wrong it doesn't tell you what is wrong. `go-cmp` does this. Let's take the following example from one of the tests from the PR.

    Test with `reflect.DeepEqual`.

    ```go
    func TestDecodeQualifiedName(t *testing.T) {
	    q, err := DecodeQualifiedName(foobar)
	    if err != nil {
	        t.Fatal(err)
	    }
	    expected := &QualifiedName{
		    NamespaceIndex: 1,
		    Name: &String{
                        // i added an extra "r" here
			    Value:  []byte("foobarr"),
			    Length: 6,
		    },
	    }
	    if !reflect.DeepEqual(q, expected) {
		    t.Error("oh noes")
	    }
    }
    ```

    Output

    > --- FAIL: TestDecodeQualifiedName (0.00s)
    > qualified-name_test.go:54: oh noes

    Same test with `go-cmp`.

    ```go
    func TestDecodeQualifiedName(t *testing.T) {
	    q, err := DecodeQualifiedName(foobar)
	    if err != nil {
	        t.Fatal(err)
	    }
	    expected := &QualifiedName{
		    NamespaceIndex: 1,
		    Name: &String{
                        // i added an extra "r" here
			    Value:  []byte("foobarr"),
			    Length: 6,
		    },
	    }
	    if diff := cmp.Diff(q, expected); diff != "" {
		    t.Error(diff)
	    }
    }
    ```

    Output

    > --- FAIL: TestDecodeQualifiedName (0.00s)
    >     qualified-name_test.go:56: {*datatypes.QualifiedName}.Name.Value[?->6]:
    >         -: <non-existent>
    >         +: 0x72

    As you can see the output is much nicer. No need to write "got: ..., want: ..." in your error message.

1. I think it's similar to the https://github.com/pkg/errors package which is on it's way to become the standard package for wrapping errors and getting proper stack traces.

1. It's from Google. It's not just any third party library.

I want to be a good citizen and since it's your package you can decide. I can change the tests if you want.

PR #20, opened 2018-08-27
add data type read value id

zemirco:
Here is the next data type `ReadValueID`. Since `ReadRequest` needs an array of `ReadValueID` I also added the type `ReadValueIDArray`. Very similar to your `StringArray`.

I also fixed some minor issues in `StringNodeID` and `QualifiedName`.

Thanks for accepting to use `go-cmp`.

### DIFF
--- a/datatypes/read-value-id.go
+++ b/datatypes/read-value-id.go
@@ -1,0 +1,213 @@
+package datatypes
+
+import "encoding/binary"
+
+// IntegerID is a UInt32 that is used as an identifier, such as a handle.
+// All values, except for 0, are valid.
+//
+// Specification: Part 4, 7.14
+type IntegerID uint32
+
+// Identifiers assigned to Attributes.
+//
+// Specification: Part 6, A.1
+const (
+	IntegerIDNodeID IntegerID = iota + 1
+	IntegerIDNodeClass
+	IntegerIDBrowseName
+	IntegerIDDisplayName
+	IntegerIDDescription
+	IntegerIDWriteMask
+	IntegerIDUserWriteMask
+	IntegerIDIsAbstract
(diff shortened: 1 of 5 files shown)


datatypes/read-value-id.go
```diff
@@ -0,0 +1,213 @@
...
+		return err
+	}
+	offset := r.NodeID.Len()
+
+	// attribute id
+	binary.LittleEndian.PutUint32(b[offset:offset+4], uint32(r.AttributeID))
+	offset += 4
+
+	// index range
+	if err := r.IndexRange.SerializeTo(b[offset:]); err != nil {
```
wmnsk: To avoid panic I always put `if xxx != nil {` block here even the spec says the field is mandatory. This may cause generating malformed byte stream, but the purpose is to be as flexible as possible (imagine fuzz-testing some OPC UA server with malformed packets).
Let me know if you have better idea :smiley:

datatypes/read-value-id.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	offset += r.IndexRange.Len()
+
+	// data encoding
+	return r.DataEncoding.SerializeTo(b[offset:])
+}
+
+// Len returns the actual length of ReadValueID in int.
+func (r *ReadValueID) Len() int {
+	return r.NodeID.Len() + 4 + r.IndexRange.Len() + r.DataEncoding.Len()
```
wmnsk: Same as `SerializeTo`. If struct field is nil it causes panic, which I want to avoid for flexibility.

wmnsk (commented):
Thanks for your work. Please check the inline comments.

wmnsk:
LGTM. Thanks!

PR #26, opened 2018-09-03
use node id interface for create session response

zemirco:
I've got `ActivateSessionRequest` working :tada: And I think I finally understood this whole `NodeID` thing. :smile: 

I've tried with two different OPC UA servers. Before pushing the PR for `ActivateSessionRequest` I had to make some changes to `CreateSessionResponse`. I got different responses from the servers. We used hard coded `NodeID` types for the `SessionID` and `AuthenticationToken` fields. That didn't work with one of the servers. We have to use the `NodeID` interface instead.

That's why I made the changes.

I will add the tests for `ActivateSessionRequest` and will then send the PR.

### DIFF
--- a/services/create-session-response.go
+++ b/services/create-session-response.go
@@ -19,9 +19,9 @@ import (
 type CreateSessionResponse struct {
 	TypeID *datatypes.ExpandedNodeID
 	*ResponseHeader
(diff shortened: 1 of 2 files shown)


zemirco:
Could you first merge this one? I thought having two separate ones would make review easier.

PR #30, opened 2018-09-03
Revise README

wmnsk:
Improved README.md a bit.

* Added goreport shield(badge)
* Updated TODO checklist
* Removed BMC link (this is no longer my lonely work)


zemirco:
Good idea. Maybe we could improve our "this is already done" and "this is missing at the moment" columns.

I like those summaries here:

https://github.com/node-opcua/node-opcua#supported-features
https://github.com/open62541/open62541/blob/master/FEATURES.md

They've grouped the requests/responses by Service. Then they've got transport, security, facets and profiles.

wmnsk:
I agree. Thanks for your proposal :smiley:

I'll work on that on another day and show you a sample.

PR #33, opened 2018-09-04
Implement UACP Client

wmnsk:
* Added `Conn` structure in UACP, with the methods that required to implement `net.Conn` interface.
* Added `ClientConfig` structure.
* Implemented `Dial()` function and some other methods.
* Added `client.go` in example.

This is still WIP - Methods implemented on `Conn` should be improved.

### DIFF
--- a/examples/client/client.go
+++ b/examples/client/client.go
@@ -3,6 +3,9 @@
 // found in the LICENSE file.
 
 // Command client provides a connection establishment of OPC UA Secure Conversation.
+//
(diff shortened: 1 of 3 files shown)


wmnsk:
Memo: Real packets exchanged between client and server in example.

[gopcua-client-server.pcapng.gz](https://github.com/wmnsk/gopcua/files/2347938/gopcua-client-server.pcapng.gz)

PR #38, opened 2018-09-04
Implement UACP Server

wmnsk:
* Added `Server` structure
* Implemented `Listen()` and `Accept()` methods
* Revised `client.go` and `conn.go`
* Added `server.go` in example/server
* Moved endpoint parser to `utils` package


wmnsk:
Memo: Real packets exchanged between client and server in example.

[gopcua-client-server.pcapng.gz](https://github.com/wmnsk/gopcua/files/2347938/gopcua-client-server.pcapng.gz)

PR #39, opened 2018-09-04
add activate session request

zemirco:
Here it is.

When this is merged I will continue working on `ReadRequest`.


zemirco:
forgot to include the fix for `ExtensionObject`. just a moment ...

wmnsk:
Thank you for the work!
It seems to work well and I want to merge asap, but I had two things in my mind.

* I feel a bit uncomfortable to put `ServiceTypeAnonymousIdentityToken` in ServiceType definitions in `services.go`, as it is not a type of service. I think we have two options to address this.

No. | Option | Pros | Cons
--- | --- | --- | --- 
1 | Put _all_ the NodeId in a file <br />(`node-id.go` or new file) | Simple and can be done easily with some script to convert NodeIds.csv to Golang constant | Somewhat tiresome to find the NodeId that users want,
2 | Keep ServiceTypeXxx as it is, and create new constant set in new file | Prefix helps users find the NodeId they want (thanks to the editor/IDE) | Somewhat confusing, as users need to check where each NodeId is defined.<br /> It might be hard to give proper prefix to some NodeIds.

* Do you plan to implement ActivateSessionResponse as well? Otherwise I'll work on that next.

zemirco:
You're right. I'm still having trouble understanding the whole protocol. `AnonymousIdentityToken` is not a service but it can be used inside an `ExtensionObject`. To decode an `ExtensionObject` we have to specify the `Identifier` somewhere.

This whole concept is very similar to the services where each service has a certain ID.

Regarding your two ideas. I would favor no. 1. I've removed the token from services and hard coded the value. I will change that once we have all NodeIDs available.

You can work on `ActivateSessionResponse`. I'll do `ReadRequest`.

wmnsk:
Looks good, but could you put the copyright on top of each file as I did in other files? Then I’ll merge this.
For NodeId definitions, I’ll create an issue and address it some other days.

PR #42, opened 2018-09-05
Improve GetPath() and add tests

wmnsk:
* Improved `GetPath()` in `utils` package. Now it raises error instead of panic when the input is invalid.
* Added tests for `ResolveEndpoint()` and `GetPath()`.

### DIFF
--- a/uacp/server.go
+++ b/uacp/server.go
@@ -74,7 +74,9 @@ func (l *Listener) Accept() (*Conn, error) {
 
 	switch msg := message.(type) {
 	case *Hello:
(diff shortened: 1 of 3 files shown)


wmnsk:
Mmm it seems net.ResolveTCPAddr fails on CCI container...

```shell-session
=== RUN   TestResolveEndpoint
--- FAIL: TestResolveEndpoint (0.11s)
	endpoint_test.go:72: {*net.DNSError}.Server:
			-: "127.0.0.11:53"
			+: ""
```

PR #47, opened 2018-09-05
add data type float

zemirco:


datatypes/float.go
```diff
@@ -0,0 +1,64 @@
...
+// found in the LICENSE file.
+
+package datatypes
+
+import (
+	"encoding/binary"
+	"math"
+)
+
+// All floating-point values shall be encoded with the appropriate IEEE-754 binary representation
```
wmnsk: Could you add a line starting from Float like other structs do?
Otherwise this comments won’t be parsed as GoDoc correctly.

wmnsk:
Nice work. Thanks!

PR #51, opened 2018-09-06
add node ids generation cmd

zemirco:
While working on the `ReadResponse` I came across the `Variant` data type. It needs all the node ids, e.g. for `Float` which is `10`.

Here is a PR adding all node ids to the package. They are auto generated from the `.csv` file. The script is located at `cmd/id/main.go`. It creates a new package `id/id.go`. I think having a very short package name makes it simple to use those ids.

Here is an example for the `UserIdentityToken` field from the `ActivateSessionRequest`

```go
UserIdentityToken: &datatypes.ExtensionObject{
	TypeID: &datatypes.ExpandedNodeID{
		NodeID: datatypes.NewFourByteNodeID(0, id.AnonymousIdentityToken_Encoding_DefaultBinary),
	},
	Length:       5,
	EncodingMask: 0x01,
	Body:         datatypes.NewByteString([]byte("0")),
}
```


wmnsk:
Looks good. Thanks!

PR #52, opened 2018-09-07
add data type variant

zemirco:


zemirco:
wow that looks wrong. i messed up

PR #53, opened 2018-09-07
add data type variant

zemirco:
Hey,

this adds the data type `Variant`. Currently `Boolean`, `LocalizedText` and `Float` can be decoded. It's similar to `Decode` from `services/service.go` where each Service is hard coded. I had to move the `Service` interface definition from the `services` package to the `datatypes` package to prevent circular references. I also removed the `String() string` method from the interface because I didn't want to implement that method for every type. I think it's more for debugging right? And as the proverb says

> The bigger the interface, the weaker the abstraction.

From https://go-proverbs.github.io/

I've written the tests so that additional types can be added easily.

I've branched from the previous PR so some changes are also included here. When the previous branch is merged those changes shouldn't be visible anymore. I needed the `id` package for various `ServiceTypes`.

For the tests I always try to take real bits from Wireshark. So far I haven't found any `Variant` where the fields `ArrayLength`, `ArrayDimensionsLength` and `ArrayDimensions` are used. That's why I left them out. If we have some real values and know how they should look we can add them to the implementation and to the tests.

What do you think?

### DIFF
--- a/datatypes/variant.go
+++ b/datatypes/variant.go
@@ -1,0 +1,102 @@
+package datatypes
+
+import (
+	"github.com/wmnsk/gopcua/errors"
+	"github.com/wmnsk/gopcua/id"
+)
+
+// Service is an interface to handle any kind of OPC UA Services.
+type Service interface {
+	DecodeFromBytes([]byte) error
+	Serialize() ([]byte, error)
+	SerializeTo([]byte) error
+	Len() int
+	// String() string
+	ServiceType() uint16
+}
+
+// Variant is a union of the built-in types.
+//
+// Specification: Part 6, 5.2.2.16
+type Variant struct {
+	EncodingMask          uint8
+	ArrayLength           *int32
+	Value                 Service
+	ArrayDimensionsLength *int32
+	ArrayDimensions       []*int32
+}
+
+// NewVariant creates a new Variant.
+func NewVariant(service Service) *Variant {
+	v := &Variant{
+		EncodingMask: uint8(service.ServiceType()),
+		Value:        service,
+	}
+	return v
+}
+
+// DecodeVariant decodes given bytes into Variant.
+func DecodeVariant(b []byte) (*Variant, error) {
+	v := &Variant{}
+	if err := v.DecodeFromBytes(b); err != nil {
+		return nil, err
+	}
+	return v, nil
+}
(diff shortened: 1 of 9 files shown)


wmnsk:
For `Service` interface, I'd like to leave it as it is, and implement new one for datatypes. Even OPC UA spec defines service types and data types in the same file: NodeId, for they are totally different in usage & development point of view.
How about creating `Data` interface and implement `DataType()` methods for each datatypes?

`String()` is not mandatory feature, so it's no problem removing it from interface. I was just wondering how we can implement it easier... Maybe with `reflect`, or something.

zemirco:
Sounds good to me. Will continue working on this PR tomorrow.

wmnsk:
Thank you for understanding. Looking forward to the update!

zemirco:
I rebased the branch. Please use `git pull --rebase` if you want to test it locally. not done yet....

zemirco:
OK I'm done. PR is ready for review. Sorry it took so long. The week was quite busy.

datatypes/variant_test.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func TestDecodeVariant(t *testing.T) {
+	for _, test := range variantTests {
+		t.Run(test.description, func(t *testing.T) {
+			v, err := DecodeVariant(test.in)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if diff := cmp.Diff(v, test.expected, opt); diff != "" {
```
wmnsk, line 63: Where's the `opt` came from...?

datatypes/variant.go
```diff
@@ -0,0 +1,92 @@
+package datatypes
```
wmnsk, line 5: Pls add copyright lines on the top.

datatypes/variant_test.go
```diff
@@ -0,0 +1,116 @@
+package datatypes
```
wmnsk, line 5: Pls add copyright lines on the top.

wmnsk (commented):
@zemirco Almost looks good. Please check my comments for some small things. Thanks.

datatypes/variant_test.go
wmnsk: OK, thanks!

PR #54, opened 2018-09-08
Update UACP state machine and add some tests

wmnsk:
To make it more convenient, UACP conn now handles UACP messages automatically and lets users(upper layer) focus on the data on the connection.


wmnsk:
It seems something's still unstable...
In my mind I need to work on the following items at least.

- [x] Should consider simplifying, as it's a bit too complicated. Especially, using *Client/*Server is tiresome and somewhat redundant.
- [x] ~Should monitor errors background continuously with another goroutine, to avoid process being blocked due to full channel capacity.~ Not actually required
- [x] Remote endpoint should be speficied each time `Dial()` is called, not in `Client` structure. This is to let a single `Client` to `Dial()` multiple endpoints.
- [x] Use `context` for goroutines invoked inside the package to avoid leaks.
- [x] Removed `Client` and `Server` as it's not really useful with it.
- [x] Removed TCP-specific funcs to be more flexible (no other L4 layer supported yet, though)

PR #72, opened 2018-09-17
add data type DataValueArray

zemirco:
This is the last data type we need for `ReadResponse`. When this is in I'll start working on it. Shouldn't be too hard since it's only `ResponseHeader`, `DataValueArray` and `DiagnosticInfoArray`.


wmnsk:
LGTM. Thanks!

PR #77, opened 2018-09-18
add read response service

zemirco:
:tada: I'm finally done and I'm able to read measurement values from our signal conditioner in pure Go. what a great day :champagne:


wmnsk:
Thanks, and great to hear it worked well!

Issue #66, opened 2018-09-15
Update networking tests


wmnsk:
`Conn`-related tests sometimes fail because new test case starts before the resources(like binded address) in previous test is not freed yet.
https://circleci.com/gh/wmnsk/gopcua/482

- [x] Consider how to address
- [ ] Update tests that requires OS resources

PR #78, opened 2018-09-18
add read response to service

zemirco:


wmnsk:
The failure of CCI tests was due to issue #66. Sorry for inconvenience.
And the update seems good. Thank you.

Not to miss these lines I’ll consider adding some tests that reminds developers, as even me oftem miss it :(

PR #112, opened 2018-10-12
Feature - UASC Security Policy Algorithms

dwhutchison:
I implemented an internal package that can be used for the encryption and decryption of the various security policies.

This package is meant to be used exclusively by the secure channel package to encrypt, decrypt, sign, and verify the contents of messages before and after they're sent on the wire, plus assist with packaging the messages by helping calculate the padding portion of the message.

The API is meant to be fairly non-intrusive.  You provide the policy URI you want to use and the local and remote keys (rsa keys for asymmetric, random nonces for symmetric).  The return struct has methods to encrypt, decrypt, sign, and verify arbitrary data with those keys.

A few notes:
- This PR does not actually integrate it into the main code; I'm working on updates to the UASC message code for full integration.  I noticed you were making similar changes so I wanted to offer this portion up to avoid duplicated effort.
- The code accepts arbitrary data, provided it's in the appropriately-sized blocks.  To avoid complicating this package, it doesn't need any knowledge of what it is encrypting and offloads the integration work to the calling package.
- Since the padding and signature is part of the UASC message, this means that the caller (the uasc package) needs to handle padding the message before encrypting, unpadding it after decryption, plus append and remove the signature.  The uasc code also needs to predict the final length of the message after encryption, but before signing because the length is part of the data that needs to be signed.  Asymmetric encryption increases the length of the payload beyond padding so this is a bit of a hassle.

Comments and suggestions welcome.


wmnsk:
Great work! The codes look really good to me :smiley: 

My work on `uasc` package and encryption was still in very early stage and much behind your work, so could you please continue working on integrating this with `uasc` package after merging this? Then I will work on another kind of stuff that does not conflict, like implementing other services, improving documents, etc.

securitypolicy/cryptoRsaPkc15.go
```diff
@@ -0,0 +1,106 @@
...
+			ciphertext = append(ciphertext, c...)
+			start = end
+			srcRemaining = len(src) - start
+		}
+
+		return ciphertext, nil
+	}
+}
+
+func signRsaPkc15(hash crypto.Hash, privKey *rsa.PrivateKey) func([]byte) ([]byte, error) {
```
wmnsk: Can you rename this as `signPKCS1v15` to be consistent with encryption/decryption funcs above?

securitypolicy/cryptoRsaPkc15.go
```diff
@@ -0,0 +1,106 @@
...
+
+		signature, err := rsa.SignPKCS1v15(rng, privKey, hash, hashed[:])
+		if err != nil {
+			return nil, err
+		}
+		return signature, err
+	}
+}
+
+func verifyRsaPkc15(hash crypto.Hash, pubKey *rsa.PublicKey) func([]byte, []byte) error {
```
wmnsk: Like sign, could you rename this as `verifyPKCS1v15`?

wmnsk (changes requested):

wmnsk:
Thank you!
Looking forward to the next PR!

magiconair:
@dwhutchison I've reworked the networking code a bit and would really like to integrate your security code. Are you still interested in working on a PR? If you have other commitments could you outline what would need to happen and be open to review the PR?

dwhutchison:
I’m definitely interested :). I actually just discovered your rewrite earlier this week so I’m going through and understanding the new code so I can finish what I started. I’ll be working on a PR shortly.

I really appreciate your work on the connection logic; that was a task I didn’t have time for and there were a few bugs that I couldn’t work around.

Dave Hutchison
________________________________

magiconair:
@dwhutchison this would be awesome. I can test this with our equipment and @kung-foo is also interested in this feature and he can probably test with other implementations. Please reach out if something is not clear in the networking layer. There are still some rough edges (bubble up errors) but this is starting to look solid now. Really cool that you want to add this 👍

PR #116, opened 2018-10-15
uacp: set SendBufferSize in HEL message

magiconair:
The UACP handshake fails when connecting to the open62541 server
and a SendBufferSize of 0. The server never responds with an ACK.
Setting the SendBufferSize fixes this.


wmnsk:
Good catch! Thanks :smiley:

Issue #115, opened 2018-10-15
Code Generation


magiconair:
Hi @wmnsk, I had started working on a pure go opcua implementation in April but didn't make much headway until recently but you've beaten me to the punch :) 

My code is roughly in the same state (encoder, decoder, uacp, session, ...) but I think having a larger and active community is more important and your project has gained some good traction. I have some experience from that with my own open-source projects.

The one big thing I can contribute is a code generator for all OPC/UA service objects and array types and I'm also thinking of writing a reflection based Encoder/Decoder to reduce the boilerplate even further.

The generated code looks like the one below and also generates all length encoded Array variants of objects like `[]*ReadValueId`.

Would you be interested in this? If yes, then we can discuss what the interfaces should look like since the generated code isn't a drop-in replacement for the current structures. I'll attach a PR with the code generator so that you can have a look.

```go
package service

import (
	"bytes"

	ua "github.com/northvolt/go-opcua/datatypes"
)

type CreateSubscriptionRequest struct {
	RequestHeader               *RequestHeader
	RequestedPublishingInterval float64
	RequestedLifetimeCount      uint32
	RequestedMaxKeepAliveCount  uint32
	MaxNotificationsPerPublish  uint32
	PublishingEnabled           bool
	Priority                    uint8
}

func NewCreateSubscriptionRequest() *CreateSubscriptionRequest {
	return &CreateSubscriptionRequest{
		RequestHeader: NewRequestHeader(),
	}
}

func (m *CreateSubscriptionRequest) ID() uint16 {
	return 787
}

func (m *CreateSubscriptionRequest) GetName() string {
	return "CreateSubscriptionRequest"
}

func (m *CreateSubscriptionRequest) Read(b *bytes.Buffer) error {
	var err error
	if m.RequestHeader, err = ReadRequestHeader(b); err != nil {
		return err
	}
	if m.RequestedPublishingInterval, err = ua.ReadDouble(b); err != nil {
		return err
	}
	if m.RequestedLifetimeCount, err = ua.ReadUInt32(b); err != nil {
		return err
	}
	if m.RequestedMaxKeepAliveCount, err = ua.ReadUInt32(b); err != nil {
		return err
	}
	if m.MaxNotificationsPerPublish, err = ua.ReadUInt32(b); err != nil {
		return err
	}
	if m.PublishingEnabled, err = ua.ReadBoolean(b); err != nil {
		return err
	}
	if m.Priority, err = ua.ReadByte(b); err != nil {
[...]

PR #117, opened 2018-10-15
gen: code generator for service objects

magiconair:
This is a code generator for OPC/UA service objects from a different
project. The generated code is **not compatible with gopcua yet** but
is here to have a discussion on whether this contribution would be
welcome and useful.

If you want to try the code generator you can run

```
cd cmd/service
go build
./service
```

See #115


magiconair:
I am going to close this in favor of #124 (which uses reflection) but we still have the option to generate encoder/decoder methods either by hand or automatically.

PR #120, opened 2018-10-23
Simplify service object constructors

magiconair:
The service object constructors all need a `TypeID` which is a four byte expanded node id in always all cases. This patch creates a helper function and uses it when creating service objects.


wmnsk:
Thanks. Simplified constructors look really helpful.

Memo:
Implementing helpers like `SetURI(uri string)` and `SetIndex(idx uint32)` might be nice to have for some users?

PR #121, opened 2018-10-23
Handle float32 NaN according to OPC/UA spec

magiconair:
From Part 6, 5.2.2.3 Float:

> The floating-point type supports positive and negative infinity and not-a-number (NaN). The IEEE specification allows for multiple NaN variants; however, the encoders/decoders may not preserve the distinction. Encoders shall encode a NaN value as an IEEE quiet-NAN (000000000000F8FF) or (0000C0FF). Any unsupported types such as denormalized numbers shall also be encoded as an IEEE quiet-NAN. Any test for equality between NaN values always fails.


wmnsk:
LGTM. Thank you!

PR #122, opened 2018-10-23
Simpler float test

magiconair:
As long as we are only testing float values within the range of precision
of the underlying datatype we should not require approximation. Any
deviation of the value from the bytes should be an error.


wmnsk:
Good catch. Thanks!

PR #124, opened 2018-10-24
reflection codec

### LINKED ISSUE #115: Code Generation

magiconair:
This patch provides a reflection based implementation of the OPC/UA
binary protocol encoding scheme.

Fixes #115


decode.go
```diff
@@ -0,0 +1,292 @@
...
+		return err
+	}
+	val.SetString(string(b))
+	return nil
+}
+
+func read(r io.Reader, n int) ([]byte, error) {
+	b := make([]byte, n)
+	p := 0
+	for p < n {
```
sbinet: this can be replaced with `io.ReadFull`

encode.go
```diff
@@ -0,0 +1,260 @@
...
+	_, err := w.Write([]byte(v))
+	return err
+}
+
+func (v *Bytes) DecodeOPCUA(r io.Reader) error {
+	b, err := ioutil.ReadAll(r)
+	if err != nil {
+		return err
+	}
+	*v = make([]byte, len(b))
```
sbinet: couldn't this be replaced with just:
```go
*v = Bytes(b)
```
?
this removes one possibly big allocation.

decode.go
magiconair: Yes :) Should have had another look at `io`.

encode.go
magiconair: I have refactored that code a bit and dropped the `Bytes` type when I started to integrate this with the existing code. You can now tag a `[]byte` with `opcua:"raw"` to just read the raw bytes and there is a fast path in the codec for that case. I'll have a look there. I'll push an update a bit later.

wmnsk:
Looks nice. I've avoided using `reflect` ever since I learned it affects performance, but it seems not too much, and there seems a lot of advantages over "manual" encoding/decoding.

In my understanding current commits are not complete, but before merging please;
* add GoDoc comments above exported API?
* cleanup the commented-out "print debugging" lines

magiconair:
yes, this isn't ready yet. The other pull requests I've submitted are a prerequisite for this so I'm working towards making this useful.

magiconair:
This is still work in progress but the `datatypes` package has been refactored to use the reflection codec and most hand-written codec code has been removed. All arrays are handled by the codec. Types like `string`, `bool` and `float` have been replaced with their native Go equivalents. 

I've moved the `DiagnosticInfo` into `datatypes` since it is one of the 27 basic data types according to the spec. The `ServersOnNetworks` needs to move to the `services` package IMO. I've also added `XmlElement` and `StatusCode` as type definitions.

I've also added my own implementation of `Variant` which handles all of the data types. Encoding is still missing but decoding already works. 

I've kept all the tests including the ones for the arrays which required some small modifications to the test runners. 

I've also added a `Reader` and `Writer` class which is based on the code in #115 and from comments of @sbinet which now allows to write codec code as follows:

```go
type ReadValueID struct {
	NodeID       *NodeID
	AttributeID  uint32
	IndexRange   string
	DataEncoding *QualifiedName
}

func (v *ReadValueID) DecodeOPCUA(r io.Reader) error {
	gr := gopcua.NewReader(r)
	v.NodeID = new(NodeID)
	gr.ReadStruct(v.NodeID)
	v.AttributeID = gr.ReadUint32()
	v.IndexRange = gr.ReadString()
	v.DataEncoding = new(QualifiedName)
	gr.ReadStruct(v.DataEncoding)
	return gr.Error()
}

func (v *ReadValueID) EncodeOPCUA(w io.Writer) error {
	gw := gopcua.NewWriter(w)
	gw.WriteStruct(v.NodeID)
	gw.WriteUint32(v.AttributeID)
	gw.WriteString(v.IndexRange)
	gw.WriteStruct(v.DataEncoding)
	return gw.Error()
}
```

I'll patch `Variant` encoding and the rest of the service classes in the next days and when all tests pass we should be good :)

Edit: the reason the CI fails is because the code in `services` is referring to types which have been removed. The tests in datatypes pass except for the ones with `Variant`.

magiconair:
@wmnsk I have a question regarding the encoding of arrays. If I read the spec correctly, then the encoder should try to make the distinction between `nil` and `empty` and encode them with length `-1` and `0` respectively. The decoder should try to preserve that distinction if possible. 

> Part 6. 5.2.5 Arrays
> One dimensional Arrays are encoded as a sequence of elements preceded by the number of elements encoded as an Int32 value. If an Array is null, then its length is encoded as −1. An Array of zero length is different from an Array that is null so encoders and decoders shall preserve this distinction.

Currently, all empty and `nil` arrays are encoded with length `0`. Also, `nil` array values in the `NewXXX` functions are encoded as empty arrays. Here is one example but there are many others.

https://github.com/wmnsk/gopcua/blob/master/services/activate-session-request.go#L39

Also, all the tests are using length `0` for `nil` array values. 

The reflection codec uses `-1` for `nil` arrays and the length incl. zero for all other cases. I think this is the correct behavior according to the spec and also regarding common sense (no surprise for the user). What I am not sure of is whether this actually matters in terms of interoperability?

If it doesn't matter then I'll update the tests to use length `-1` for `nil` arrays and `0` for empty ones.

magiconair:
The other thing is that once we go down that road then the `NewXXX` functions also become somewhat obsolete since you just create a regular Go struct with Go types. We would have named arguments and zero values for fields could be omitted. Since users of the library should use a high level client anyway I don't think this is an issue.

magiconair:
~~One more thing I've just discovered is that varadic arguments like `...string` always seem to generate an empty array and never `nil` which makes encoding arrays dependent on how the object is created.~~

~~I've stumbled over this in here and the special casing for the empty first value should probably be handled in a higher-level client and not the base encoding layer.~~

https://github.com/wmnsk/gopcua/blob/master/services/find-servers-on-network-request.go#L43-L59

Edit: this is not the case (see https://play.golang.org/p/xI0et9k0mC2) I think this is just the way the test and the code was written.

wmnsk:
Thank you very much for the great work. Let me check precisely next week, as I'm on holidays this week, but so far the changes seem nice.

For the question about the array, I agree on the idea to handle nil and empty in the different way as described in the spec. So please proceed your work. Thanks.

wmnsk:
Hi @magiconair, sorry not responding.
I've started working on looking into your code locally.

Let me make sure, do you still have any works remaining in this PR?

magiconair:
Hi @wmnsk, no worries. I was sick for a while and then got swamped by work. This isn't done yet but there should not be much more work left. I'll try to pick this up soon.

wmnsk:
@magiconair Got it, thanks. Take good care of yourself 😃

magiconair:
Haven't forgotten. Just super busy but I want to finish this off. Please be patient :)

wmnsk:
No problem. Good luck with what you’re working on!

magiconair:
@wmnsk yes, lets pick this up again. Since the reflection codec handles the allocations the entire buffer and message handling and decoding needs work. I'll try to clean up as much as possible tonight and then you can have a look, OK?

wmnsk:
Sounds good, thank you.
Looking forward to the updates 😃

magiconair:
It has been a while since I've worked on that code. 

The tests compile now but some `uacp` tests fail and others time out. I suggest `go test -timeout 10s ./...` for now.

I also think that the new encoder/decoder can be cleaned up a bit more but let us focus on getting this functional first. 

The main issue is that server and client no longer deal with `[]byte` but messages and that encoding/decoding should happen transparently in the code. There are a number of tests which send arbitrary byte sequences like `0xdeadbeef` around and these tests have to be rewritten. Also, some examples look like they could use some love. 

I'll have a look at the failing `uacp` tests. @wmnsk could you try to get the server functional again?

magiconair:
Found the reason for the failing `uacp` tests. Encoding a `Hello` message now encodes only the message itself but not the message header with `HELF` and the message size. I'll push a change soon.

datatypes/variant.go
```diff
@@ -5,92 +5,299 @@
...
-// NewVariant creates a new Variant.
-func NewVariant(data Data) *Variant {
-	v := &Variant{
-		EncodingMask: uint8(data.DataType()),
-		Value:        data,
-	}
-	return v
+func NewVariant(v interface{}) *Variant {
+	va := &Variant{}
+	va.Set(v)
```
golangcibot: Error return value of `va.Set` is not checked (from `errcheck`)

datatypes/diagnostic-info.go
```diff
@@ -0,0 +1,221 @@
...
+		str = append(str, fmt.Sprintf("%d", d.NamespaceURI))
+	}
+	if d.HasLocale() {
+		str = append(str, fmt.Sprintf("%d", d.Locale))
+	}
+	if d.HasLocalizedText() {
+		str = append(str, fmt.Sprintf("%d", d.LocalizedText))
+	}
+	if d.HasAdditionalInfo() {
+		str = append(str, fmt.Sprintf("%s", d.AdditionalInfo))
```
golangcibot: the argument is already a string, there's no need to use fmt.Sprintf (from `gosimple`)

datatypes/diagnostic-info.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+	if d.HasInnerStatusCode() {
+		str = append(str, fmt.Sprintf("%d", d.InnerStatusCode))
+	}
+	if d.HasInnerDiagnosticInfo() {
+		str = append(str, fmt.Sprintf("%s", d.InnerDiagnosticInfo))
```
golangcibot: should use String() instead of fmt.Sprintf (from `gosimple`)

uasc/session.go
```diff
@@ -333,8 +316,8 @@ func (s *Session) handleActivateSessionRequest(as *services.ActivateSessionReque
 		}
 		*/
 
-		for _, str := range as.LocaleIDs.Strings {
-			s.cfg.LocaleIDs = append(s.cfg.LocaleIDs, str.Get())
+		for _, str := range as.LocaleIDs {
```
golangcibot: should replace loop with `s.cfg.LocaleIDs = append(s.cfg.LocaleIDs, as.LocaleIDs...)` (from `gosimple`)

magiconair:
I've added some debug logging to see where the handshake hangs. I am currently testing this with `go test -v -timeout 5s -run TestClientWrite ./uasc` and I've tracked it down to the `Decode` of the `OpenSecureChannelRequest` method but don't know why it hangs/crashes there.

See `services/service.go` line 114.

magiconair:
```
$ go test -v -timeout 10s -run TestClientWrite ./uasc
=== RUN   TestClientWrite
2019/02/11 23:31:49 uacp: connect
2019/02/11 23:31:49 uacp: send *uacp.Hello
2019/02/11 23:31:49 uacp: hello 1
2019/02/11 23:31:49 uacp: recv 64 bytes
2019/02/11 23:31:49 uacp: recv *uacp.Hello
2019/02/11 23:31:49 uacp: recv hello
2019/02/11 23:31:49 uacp: send *uacp.Acknowledge
2019/02/11 23:31:49 uacp: recv 28 bytes
2019/02/11 23:31:49 uacp: recv *uacp.Acknowledge
2019/02/11 23:31:49 uacp: recv ack
2019/02/11 23:31:49 uacp: established
2019/02/11 23:31:49 uasc: send OPNF 164 bytes 4f504e46a4000000000000002f000000687474703a2f2f6f7063666f756e646174696f6e2e6f72672f55412f5365637572697479506f6c696379234e6f6e65ffffffffffffffff01000000050d00000100be01000014082b9459c2d4010100000000000000ffffffffffff000000000000000000000000000100000020000000b94c1f212325c2edaaf32e6398d86fd600884806a66a694fe15a876c5560608080ee3600
2019/02/11 23:31:49 uacp: recv 164 bytes
2019/02/11 23:31:49 uasc: recv 164 bytes 4f504e46a4000000000000002f000000687474703a2f2f6f7063666f756e646174696f6e2e6f72672f55412f5365637572697479506f6c696379234e6f6e65ffffffffffffffff01000000050d00000100be01000014082b9459c2d4010100000000000000ffffffffffff000000000000000000000000000100000020000000b94c1f212325c2edaaf32e6398d86fd600884806a66a694fe15a876c5560608080ee3600
2019/02/11 23:31:49 m.Header: MessageType: OPN, ChunkType: 70, MessageSize: 164, SecureChannelID: 0
2019/02/11 23:31:49 m.AsymmetricSecurityHeader: SecurityPolicyURI: http://opcfoundation.org/UA/SecurityPolicy#None, SenderCertificate: [], ReceiverCertificateThumbprint: []
2019/02/11 23:31:49 m.SequenceHeader: SequenceNumber: 1, RequestID: 3333
2019/02/11 23:31:49 gr.Error: <nil>
2019/02/11 23:31:49 Decode: id:446
2019/02/11 23:31:49 uasc: decode failed: invalid node id type 12
2019/02/11 23:31:54 uasc: send OPNF 164 bytes 4f504e46a4000000000000002f000000687474703a2f2f6f7063666f756e646174696f6e2e6f72672f55412f5365637572697479506f6c696379234e6f6e65ffffffffffffffff02000000050d00000100be01000012c0269759c2d4010200000000000000ffffffffffff00000000000000000000000000010000002000000096447067716a61701c6ea8bcb1fdbd86c4f290e23881756b8492e982063c7cc580ee3600
2019/02/11 23:31:54 uacp: recv 164 bytes
2019/02/11 23:31:55 uasc: recv 164 bytes 4f504e46a4000000000000002f000000687474703a2f2f6f7063666f756e646174696f6e2e6f72672f55412f5365637572697479506f6c696379234e6f6e65ffffffffffffffff02000000050d00000100be01000012c0269759c2d4010200000000000000ffffffffffff00000000000000000000000000010000002000000096447067716a61701c6ea8bcb1fdbd86c4f290e23881756b8492e982063c7cc580ee3600
2019/02/11 23:31:55 m.Header: MessageType: OPN, ChunkType: 70, MessageSize: 164, SecureChannelID: 0
2019/02/11 23:31:55 m.AsymmetricSecurityHeader: SecurityPolicyURI: http://opcfoundation.org/UA/SecurityPolicy#None, SenderCertificate: [], ReceiverCertificateThumbprint: []
2019/02/11 23:31:55 m.SequenceHeader: SequenceNumber: 2, RequestID: 3333
2019/02/11 23:31:55 gr.Error: <nil>
2019/02/11 23:31:55 Decode: id:446
2019/02/11 23:31:55 uasc: decode failed: unexpected EOF
panic: test timed out after 10s
```

The fact that decoding is failing with different errors is suspicious.

wmnsk:
I’ve started looking into the issue but found no clues yet... continued to work on tomorrow 🙃

magiconair:
I've started debugging as well and realized that the `DecodeOPCUA(r io.Reader)` signature might not be so useful. I think the `io.Reader(b []byte) (int64, error)` might be better since we have all the bytes when decoding a message. 

I can refactor the `gopcua.Reader` to a `gopcua.Buffer` which does more or less the same thing but on a byte array instead of a reader since this allows us to look at the content of the buffer during decoding. And this would also only be necessary for the messages where we we cannot use reflection.

I'll try that a bit to see how far I get.

buffer.go
```diff
@@ -0,0 +1,291 @@
...
+	"io"
+	"math"
+	"time"
+
+	"github.com/wmnsk/gopcua/utils"
+)
+
+const (
+	null    = 0xffffffff
+	f32qnan = 0xffc00000
```
golangcibot: `f32qnan` is unused (from `varcheck`)

buffer.go
```diff
@@ -0,0 +1,291 @@
...
+	f64qnan = 0xfff8000000000000
```
golangcibot: `f64qnan` is unused (from `varcheck`)

buffer.go
```diff
@@ -0,0 +1,291 @@
...
+	if b.err != nil {
+		return
+	}
+	n, err := r.Decode(b.buf[b.pos:])
+	if err != nil {
+		b.err = err
+		return
+	}
+	b.pos += n
+	return
```
golangcibot: S1023: redundant `return` statement (from `gosimple`)

magiconair:
@wmnsk I've cleaned up the Encoder/Decoder and found some bugs in the general Message decoder. So now more tests pass for example `ClientWrite` and `ServerWrite` but at least one still hangs.

buffer.go
magiconair: yes, need to re-add this behavior

magiconair:
`TestClientSessionWrite` and `TestServerSessionWrite` both hang and `TestMessage` and `TestDecodeMessage` fail with some small issues. The message size isn't set in the header and one 0 vs -1 difference. Almost there.

magiconair:
`TestClientSessionWrite` and `TestServerSessionWrite` are the only failing tests left.

magiconair:
Looks like they at least partially hang because of data races. These are also on `master`. Try running `go test ./... -race` I think the `lenChan` and sharing of the `rcvBuf` has something to do with this. Still digging.

conn/secure_channel.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/wmnsk/gopcua/datatypes"
+	"github.com/wmnsk/gopcua/services"
+	"github.com/wmnsk/gopcua/uasc"
+)
+
+type SecureChannel struct {
+	c           *Conn
+	cfg         *uasc.Config
+	reqHeader   *services.RequestHeader
+	resHeader   *services.ResponseHeader
```
golangcibot: `resHeader` is unused (from `structcheck`)

magiconair:
I haven't been able to find the cause for the hanging test yet. I suspect a data race but could not nail it down.

One thing that looks weird is the approach with the `lenChan` and the shared receive buffer. We should probably just move `[]byte` and objects around.

Then both the session and the secure channel modify the sequence numbers and request handles before sending a message and this should also probably happen just in one place. 

To understand how this should work I have spend half a day or so to re-write the uasc layer in a separate package so that it fits with the new codec. I've copied the relevant code from the current implementation and only worked on how the layers are setup and messages are decoded and passed on. 

This is in the `conn` package and currently only provides the client but the structure looks a bit cleaner since I've moved all the counter handling into a single method and also inject `TypeID` and `RequestHeader` via reflection. For example:

```go
func (s *SecureChannel) sendOpenSecureChannelRequest() error {
	nonce := make([]byte, 32)
	if _, err := rand.Read(nonce); err != nil {
		return err
	}

	svc := &services.OpenSecureChannelRequest{
		ClientProtocolVersion:    0,
		SecurityTokenRequestType: services.ReqTypeIssue,
		MessageSecurityMode:      s.cfg.SecurityMode,
		ClientNonce:              nonce,
		RequestedLifetime:        s.cfg.Lifetime,
	}
	return s.send(svc)
}

func (s *SecureChannel) handleOpenSecureChannelResponse() error {
	svc, err := s.recv()
	if err != nil {
		return err
	}
	resp, ok := svc.(*services.OpenSecureChannelResponse)
	if !ok {
		return fmt.Errorf("got %T, want OpenSecureChannelResponse", svc)
	}
	s.cfg.SecurityTokenID = resp.SecurityToken.TokenID
	return nil
}
```

The `send` method then does all the heavy lifting with the counters, type ids and the request header:

```go
func (s *SecureChannel) send(svc services.Service) (err error) {
	// use reflection to set typeid and request header
	// the send method will update the request header fields
	// and reset them if the call failed.
	//
	// this is a workaround to avoid changing the existing datastructures
	// and break existing code. We should send type id and request header
	// separately here and remove these fields from the structs.:w
	typeID := datatypes.NewFourByteExpandedNodeID(0, svc.ServiceType())
	val := reflect.ValueOf(svc)
	val.Elem().Field(0).Set(reflect.ValueOf(typeID))
	val.Elem().Field(1).Set(reflect.ValueOf(s.reqHeader))

	// update counters and reset them on error
	s.cfg.SequenceNumber++
	s.reqHeader.RequestHandle++
	s.reqHeader.Timestamp = time.Now()
	defer func() {
		if err != nil {
			s.cfg.SequenceNumber--
			s.reqHeader.RequestHandle--
		}
	}()

	// todo(fs): should we drop the headers from Message and generate them here?
	// encode the message
	m := uasc.NewMessage(svc, s.cfg)
	b, err := m.Encode()
	if err != nil {
		return err
	}

	// send the message
	_, err = s.c.Write(b)
	return err
}
```

With this approach I managed to get something like this working:

```go
func TestClientRead(t *testing.T) {
	c := NewClient("opc.tcp://localhost:4840", nil)
	if err := c.Connect(); err != nil {
		t.Fatal(err)
	}
	defer c.Close()

	v, err := c.Read("ns=0;i=2258")
	if err != nil {
		t.Fatal(err)
	}
	t.Logf("time: %v", v)
}

// --- PASS: TestClientRead (0.01s)
//     conn_test.go:41: time: 2019-02-24 20:48:54.994226 +0000 UTC
```

I really want to get this change over the line and then work on a generic client and server with a node store. Right now, I need to use a cgo solution which isn't fun :) The way the code in the `uasc` package deals with the messages seems related so I probably have to touch that one, too.

magiconair:
Note that I've squashed the changes except for the new uasc code.

wmnsk:
Thank you so much for the work! 
Here's a qiuck review:

1. I want to keep any kind of OPC-UA Connection compliant with `net.Conn` interface.
So, could you rename the `Read()` method as `ReadOPCUA()` and leave the form of `Read()` as `Read([]byte) (n int, err error)`?

2. the name of package `conn` doesn't look good because it is conventionally used to represent a connection created by `Dial()` and `Accept()`.
How about using `network` or anything different?

```go
conn, err := Dial(...) // this won't work when "conn" package is imported
```

magiconair:
Sure and yes `conn` is a bad choice. :) Was just trying to get something to work. 

Next thing is to get the async multiplexing to work on the secure channel. I think that once we have this in place then we can move forward with more high-level clients. I really would also like to have a fully functioning server with a node store and the default node tree in place at some point as well. 

Do you think you can help with the server implementation once the new client code is working?

wmnsk:
Sure, I'd be happy to work on server side, as the networking function is exactly what I tried to improve next.
I think it may also be good for me to work on it because it makes me dive much deeper into your codes, which would be better review of this PR.

uascnew/secure_channel.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/wmnsk/gopcua/datatypes"
+	"github.com/wmnsk/gopcua/services"
+	"github.com/wmnsk/gopcua/uasc"
+)
+
+type SecureChannel struct {
+	c           *Conn
+	cfg         *uasc.Config
+	reqHeader   *services.RequestHeader
+	resHeader   *services.ResponseHeader
```
golangcibot: `resHeader` is unused (from `structcheck`)

uascnew/secure_channel.go
```diff
@@ -0,0 +1,243 @@
...
+	if s.cfg.SecureChannelID == 0 {
+		s.cfg.SecureChannelID = h.SecureChannelID
+		log.Printf("conn %d: set secure channel id to %04x", -1, s.cfg.SecureChannelID)
+	}
+
+	return m.Service, nil
+}
+
+func (s *SecureChannel) monitor() {
+	go s.run(s.ctx)
```
golangcibot: Error return value of `s.run` is not checked (from `errcheck`)

magiconair:
I have moved the old network code from `uasc` to `uasc/old` and commented it out including the tests until we can remove it. Refactoring it around the new networking and encoding code makes less sense since there are some substantial differences in approach.

I have also moved the encoder to the `ua` package so that we can have a high-level `Client` and `Server` in the main package without having circular references. I think at least the `datatypes` should be in the `ua` package as well and maybe even the `services` package.

I've also thought a bit about the request/response pattern and realized that we are dealing with Futures in the secure channel. I have refactored `Send` to `SendAsync` which now returns a channel with the value and the error. The `Send` method is the sync variant and takes a callback.

```go
func (s *SecureChannel) Send(svc services.Service, h func(interface{}) error) error {
	ch, err := s.SendAsync(svc)
	if err != nil {
		return err
	}

	if h == nil {
		return nil
	}

	// todo(fs): handle timeout
	resp := <-ch
	if resp.Err != nil {
		return resp.Err
	}
	return h(resp.V)
}

type Response struct {
	V   interface{}
	Err error
}

func (s *SecureChannel) SendAsync(svc services.Service) (resp chan Response, err error) {
    ...
}
```

Then the `Read` call in the `Client` can now be written like this:

```go
func (c *Client) Read(id *uad.NodeID) (*uad.Variant, error) {
	req := &uas.ReadRequest{
		MaxAge:             0,
		TimestampsToReturn: 0,
		NodesToRead: []*uad.ReadValueID{
			&uad.ReadValueID{
				NodeID:       id,
				AttributeID:  uad.IntegerIDValue,
				DataEncoding: &uad.QualifiedName{},
			},
		},
	}

	var res *uad.Variant
	err := c.sechan.Send(req, func(v interface{}) error {
		r, ok := v.(*uas.ReadResponse)
		if !ok {
			return fmt.Errorf("invalid response: %T", v)
		}
		if len(r.Results) > 0 {
			res = r.Results[0].Value
		}
		return nil
	})
	return res, err
}
```

The `examples/datetime` now uses the high-level client from the main package. I'll try to build a functional server to make sure that the basic handshaking works but then I think we should merge this to master.

magiconair:
@wmnsk I am struggling a bit with your `uacp.Conn` implementation and maybe I don't fully understand how uacp works. 

The way I understand the UACP protocol is that it performs a basic HEL/ACK handshake to negotiate the protocol version and buffer sizes and then gets out of the way to pass the raw connection interface to the UASC layer. That means that after the handshake it no longer parses any messages since all parsing - including the header (MsgType, ChunkType, MsgSize) - is done in UASC.

Your code however also checks for `HEL` messages on an existing server connection which only makes sense if a client would multiplex multiple secure channels over the same TCP connection or would reuse an existing connection for a new secure channel after the current one has been closed. Is that how OPC/UA clients behave? I thought they create a new TCP connection per secure channel.

Maybe you can share some wisdom here.

wmnsk:
I read the spec again, and your understanding seems right, as "7.1.3  Establishing a connection , Part6" says:
> Connections  may  be  initiated  by  the  Client  or  by  the  Server  when  they  create  a TransportConnection and establish a communication with their peer.

Now the only thing I'm not really sure is error handling. Some of the `ERR` message might be sent after a connection is established, so if `Conn` stops parsing everything, errors cannot be handled gracefully.

To be more precise, in the below table, 

* Red ones should be handled during the UACP handshake
* Green ones should be handled in the upper(UASC) layer
* For the rest I'm not sure... These errors should at least be notified to the user before closing the conn, I think.

image

magiconair:
Handling the errors in `uacp` after the handshake is indeed a problem since a generic `net.Conn` implementation cannot guarantee to see all messages without parsing them itself. The upper layer could make partial reads, for example. 

We could model `uacp.Read` so that it only returns full message blocks which means parsing the first 8 bytes, checking for message types and then reading `MessageSize` bytes. However, then the caller cannot make random calls to `Read` anymore and must provide a buffer which is large enough to hold the full message. 

This all feels a bit arbitrary. I think keeping `uacp.Conn` simple, handle the handshake (incl. errors during the handshake), negotiate the buffer sizes and then get out of the way provides the cleanest separation. Then we handle the rest in the secure channel. 

This is what this could look like (ignore that it is in the wrong package). Might also make sense to move all net related code to something like `uanet`.

https://github.com/northvolt/gopcua/blob/reflect-codec/uasc/conn.go

wmnsk:
Clean v.s. Practical always bothers me when implementing networking protocols, and every time I decide not to follow the spec at some moment 😇
I think one of the biggest concern to take the practicality over cleanliness is the changes in the spec in the future. However the spec is not likely to change shortly, and in that case we'll need a lot of work for other parts anyway.

So, as my conclusion, I agree with you about putting networking-related codes in a package together.

magiconair:
I'll focus on getting something working and then you can have a look.

magiconair:
I've tried adding another service object to handle subscriptions and took the opportunity to refactor the handling of the the service type id. I've dropped the `Service` interface and created a map of service ids to types. This then enables to move the TypeID into the `Message` object which means we can drop the `ServiceType` method on the service objects. This means that soon the service objects are just structs without any helper methods or constructors.

I've also started to simplify the service tests by having a static null Request and response header instead of having a different one for every test. 

I would actually like to move the Request/Response Header  into the Message object as well but I think they are actually part of the object even though they are filled by the client/server.

Since the server is a lot more work I will try to handle read, write, subscribe and maybe method call for the high-level client since this is what I need for project right now.

magiconair:
Hi, @wmnsk I have a question on how we want to proceed from here. 

I have now reached a state where this library is becoming useful and will be working on it in the coming months. My immediate focus will be on the client getting read, write, subscribe and browse working but I will need a full server eventually. Adding security is high on the list right after error handling. My focus is on the TCP protocol with binary encoding. I don't plan to support any other protocols or encodings at this point.

My guess is that I will be writing most/all of the code and this means that I need full access rights to the project for refactoring, adding examples, etc. since I need this for work. I think the only thing missing from current `master` is the server implementation which was functional but bare-bones. Maybe we can move that to `legacy` until we have an implementation again.

I don't want to hi-jack your project but I need to make a choice on governance.

Given what I will have to build in the coming weeks I can:

1. Fork
2. Ask you to move this project to a neutral Github org like `gopcua` where we share the ownership and some of my colleagues can also contribute.

Having one high-quality Go client for OPC/UA and a single community has lots of benefits so option 2 is my preference. This also depends a bit on what your plans are in general. Maybe you have lost interest or have other commitments. In that case we could also transfer the ownership of the repo to us.

I have been a full-time Go developer since 2011 and have two somewhat successful open source projects (https://github.com/magiconair/properties and https://github.com/fabiolb/fabio) and know the amount of work that can come with maintaining them.

Again, I don't want to hi-jack your repo but I have to figure out how to move forward.

Please let me know what you think.

Frank

uacp/conn.go
```diff
@@ -1,440 +1,326 @@
...
+	a.sndBufSize = ack.SendBufSize
+	log.Printf("conn %d: recv ACK:%v", a.id, ack)
 	return nil
 }
 
-type state uint8
+func (a *Conn) srvhandshake(endpoint string) error {
+	b, err := a.recv()
+	if err != nil {
+		a.sendError(BadTCPInternalError)
```
golangcibot, line 205: Error return value of `a.sendError` is not checked (from `errcheck`)

uacp/conn.go
```diff
@@ -1,440 +1,326 @@
...
-	srvStateEstablished
-)
+	// HEL or RHE?
+	msgtyp := string(b[:4])
+	msg := b[hdrlen:]
+	switch msgtyp {
+	case "HELF":
+		hel := new(Hello)
+		if _, err := ua.Decode(msg, hel); err != nil {
+			a.sendError(BadTCPInternalError)
```
golangcibot, line 216: Error return value of `a.sendError` is not checked (from `errcheck`)

uacp/conn.go
```diff
@@ -1,440 +1,326 @@
...
+			return err
+		}
+		if hel.EndPointURL != endpoint {
+			a.sendError(BadTCPEndpointURLInvalid)
```
golangcibot, line 220: Error return value of `a.sendError` is not checked (from `errcheck`)

wmnsk:
Hi, thank you for the continuous work and sorry I was taking these days off for a trip.

Actually giving you the collaborator privilege is what I was thinking to do after completed merging this changes, but if your colleagues will also be able to contribute, creating organization seems fine.

So I’d be happy to take option 2.

magiconair:
Thank you. I have created https://github.com/gopcua and invited you as a member. You can move the repo there and I'll update your privileges once you have accepted the invitation. I might also get a vanity domain for the library. ~~We could use `go.opcua.net` or `go.opcua.org`~~

Update: They're both taken.

wmnsk:
For the organization, how about using some other keyword more generic like `ics`, `scada`, `industry` etc, instead of `opcua` so that we can collaborate with other factors in the industry in the future?

Also I created #gopcua channel in Gophers Slack, which would be helpful for general discussions not related to existing Issues/PRs on GitHub.

magiconair:
We can do that, however I would focus on getting something working first until we branch out.

magiconair:
I've made you an owner. Could you enable 2FA authentication for your account since I would like to make this mandatory for this org?

magiconair:
Got both `gopcua.dev` and `gopcua.org` :)

magiconair:
need to add support for chunking b/c Siemens...

wmnsk:
> I have now reached a state where this library is becoming useful and will be working on it in the coming months. My immediate focus will be on the client getting read, write, subscribe and browse working but I will need a full server eventually. Adding security is high on the list right after error handling. My focus is on the TCP protocol with binary encoding. I don't plan to support any other protocols or encodings at this point.

Just to clarify, next steps for me would be;
* implementing fuzzing for the robustness enhancement, not to panic easily.
* maybe implementing server? but if you can do it'd be faster, as I won't be able to work on this full-time.

And, after releasing the first version, I'm wondering to create a side project for security point of view, like a testing tool to audit/fuzz other endpoints.

> Got both gopcua.dev and gopcua.org :)

Cool. Thanks!
The purpose is to publish a static website for the project, right? In that case let's use [Hugo](https://gohugo.io/) and GitHub pages so we can keep working only with golang.

magiconair:
Fuzzing is important but we are still lacking some basic functionality like support for chunking and error handling. I will focus on getting this to work first, then add more tests and features. I don't have experience in setting up fuzzing so please go ahead. However, we also need a server with an address space and some higher level functions.

The vanity domains would be also for the import path, e.g.

```go
import "gopcua.dev/opcua"
// or
import "gopcua.org/opcua"
```

instead of 

```go
import "github.com/gopcua/opcua"
```

magiconair:
hugo is a good choice :)

wmnsk:
So, maybe it's time to merge this PR to move to the organization?

> The vanity domains would be also for the import path, e.g.

Yeah it's nice to have that less redundant import path 👍

magiconair:
Lets do it

wmnsk:
Merged. Then I'll create `gopcua/opcua` repo.

wmnsk:
Working on fixing import path etc.

wmnsk:
@magiconair
Now it's done. Pls check.

magiconair:
Hmm, I thought you would just move the repo. People would then be forwarded automatically. You can do this with Transfer ownership as long as you are a member of that other organization. Keeps all the stars, forks, issues, and followers unless you want to start from scratch.

magiconair:
We can still do that by deleting the new repo and then transferring this repo to the org.

wmnsk:
Ah I haven’t had that idea. Let me give it a try.

kung-foo:
Regarding vanity package paths, does this require someone to promise to maintain the domain? if someone is only involved in the project for a short period of time, I wouldn't want to see the paths break or be forced to change.

magiconair:
The domain currently sits with Northvolt and I will be working on this for a while. Also, changing the paths is a simple search and replace if that ever should happen.

PR #126, opened 2018-10-25
Do not run unit tests in parallel

magiconair:
There is little benefit unless the tests are long-running.

obsolete by #128


magiconair:
obsolete by #128

PR #127, opened 2018-10-25
type all service constants

magiconair:
Setting the type of a constant only on the first value of a list of
constants does not have the effect to set the type for all subsequent
values. The current code would only set the first value to an `uint16` and
leave all other values to be of type `int`.

This is not an issue in a direct comparison since the Go compiler will
adjust the type of the constant but it fails in the following pattern:

```go
id := uint16(5)
if got, want := id, ServiceXXX; got != want {
     t.Fatalf("got %d want %d", got, want)
}
```
In this case got is of type `uint16` and want of type `int` which produces a
compiler error. Setting the type explictly avoids this.

### DIFF
--- a/services/service.go
+++ b/services/service.go
@@ -12,28 +12,28 @@ import (
 // ServiceType definitions.
 const (
 	ServiceTypeFindServersRequest           uint16 = 422
-	ServiceTypeFindServersResponse                 = 425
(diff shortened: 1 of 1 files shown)


wmnsk:
LGTM. Thanks!

PR #128, opened 2018-10-25
Simplify codec tests

magiconair:
A lot of the tests follow the same pattern which can be expressed in a helper package.

I'll add one commit per object so that we can test the changes.

~~So please don't merge this yet.~~

This can be merged now.


magiconair:
I've refactored all tests which do binary encoding testing to use a common test case and test runner in `util/codectest`. I'm open for a better name.

Each case is in its own commit and when I had to make a small adjustment to the code it is a separate commit as well. 

The only test which is still in the old form is the `uasc/message_test.go` which fails in the payload comparison. I've kept both the old and the new test and marked the new test to be skipped. Maybe you know what is going on.

I have copied all test cases verbatim and ran both the old and the new test before deleting the old one.

Some decoders had a small inconsistency in handling the payload on encoding and decoding (`nil` vs `[]byte{}`) which I've fixed.

~~Could you please merge #125 , #126 and #127 before this one since they are somewhat in order?~~

Actually, just merge #127 before #128 since the other two are obsolete with this patch. Merging them all results in merge conflicts. 

Thank you
Frank

wmnsk:
Looks like a nice hack, thanks!
Let me see the changes a bit more carefully as it contains so many changes.

magiconair:
Thank you and yes, this is a big change. OTOH, if we need to touch the test execution we only need to touch one place.

wmnsk:
The changes looks OK. Thank you for much work!

* Could you rebase(squash) commits to into a few number of commits? Then I'll make merge-commit.
I think aggregating "refactor XXX" commits does not harm on maintenance in the future.

* For the name "codectest", which in my understanding means "encoding/decoding test", I don't have any idea there's better name or not. So it's OK for me unless someone better in English has good idea. (cc: @zemirco, @dwhutchison)

magiconair:
You can do that in github if you enable this for the repo in the settings. It allows both squashing and rebasing. 

However, for this change I would try to keep the individual commits so that it is visible what has happened for every single test. Especially the commits where I made small changes to the behavior of the codec code. So rebase yes, squash I would advise against.

wmnsk:
What I just wanted to do is making the commits fewer without changing the author(Squqshing on GitHub by myself changes the author isn't it?)

IMHO the granularity of commit can be at least per-package(directory) even we want to look back the changes in the future. Could you tell me the reason to leave it as per-file basis?

_To be honest it is acceptable for me to leave it, as I have not cared so much about it, but for the future I just want to discuss how it should be now_

magiconair:
That's true. It changes the author. Forgot about that.

There is no hard and fast rule. I tend to make more complex changes more granular and regular to make them easier to review in pieces. I guess it is just habit.

wmnsk:
OK.
I don't care much about it before first release, so I'll merge this PR as it is.
When releasing, I'll create CONTRIBUTING.md to make it consistent. So please give me your feedback on it then, and let's discuss if needed 😃

(I don't even have the criteria of "first release" for now, though...)

magiconair:
Let me spend 10 min on rebasing. I'll try to group the changes. If that's simple then we're done.

magiconair:
```
pick 698b51f add go-cmp to circle-ci config
pick 0bab248 add package for testing encoding and decoding
pick 9bcfe51 sync encode/decode behavior for payload
pick 4b37200 refactor codec tests
pick cc49919 attempt to refactor TestMessage but fails in payload
```

magiconair:
Better?

magiconair:
I've rebased it on master as well.

wmnsk:
Looks better. Thank you!

PR #131, opened 2018-10-31
Make NodeID a union type

magiconair:
NodeID is currently a polymorphic type with different implementations behind a rather large interface. This makes marshalling a bit different for `NodeID` since the decoder has to determine first the type and then create the right object. By making NodeID a regular struct this problem goes away and NodeID is created like any other type. This will make a generic codec (generated or reflected) simpler since it does not have to deal with this special case. The reflection codec for example will not have to deal with interfaces just to handle `NodeID`.

Since NodeID stores the state in unexported fields the codec tests break since `go-cmp` does not handle them by default and there is no simple option to turn this behavior on. There is an open issue in https://github.com/google/go-cmp/issues/40 with a somewhat clumsy workaround, IMO.

My suggested solution to this is to replace `go-cmp` with `github.com/pascaldekloe/goe/verify` (at least for the codectest) which we developed within my team at eBay for running tests. (so I have some skin in the game) `go-cmp` also feels a bit feature-rich which goes against the keep-it-simple attitude of Go, IMHO.

The heavy lifting was done in #128 which makes this a simple change either way. 

I'm open for suggestions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,6 +14,7 @@ jobs:
       - checkout
       - run: go get -u github.com/pkg/errors
       - run: go get -u github.com/google/go-cmp/cmp
+      - run: go get -u github.com/pascaldekloe/goe/verify
       - run: go install ./...
       - run: go get -v -t -d ./...
       - run: go get -u golang.org/x/lint/golint
--- a/datatypes/expanded-node-id.go
+++ b/datatypes/expanded-node-id.go
@@ -14,13 +14,13 @@ import (
 //
 // Specification: Part 6, 5.2.2.10
 type ExpandedNodeID struct {
-	NodeID
+	NodeID       *NodeID
 	NamespaceURI *String
 	ServerIndex  uint32
 }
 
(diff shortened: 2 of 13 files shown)


magiconair:
I'll update the comments but can you give an example on when you'd want to update the `NodeID` fields? What would the behavior be if you update a string value on a numeric id? noop, error, panic?

wmnsk:
I don't want to leave it for long time without setters, so implementing in this PR is the best for me.
For second question, I'm expecting to return error for attempts to set wrong type value.

If you're not so much interested in implementing such feature or you have another thing you want to do earlier I'll try that instead after merging this PR.

wmnsk:
Thanks.
Yes, I understand letting it immutable is good for normal use. I expect the situation like manipulating the packets received to "recycle" it, and that's why I think the setter can be useful.

magiconair:
I've added setters for the identifier and the namespace as well as a getter for the namespace and the tests. I did not add a setter for the encoding mask since that would allow you to change the type of the node id. There are setters for the flags.

wmnsk:
Cool. Thank you!

PR #134, opened 2018-11-12
Correct the symmetric keygeneration algorithm

dwhutchison:
Been busy and haven't had much time to work on this, but I found a few bugs in the symmetric security algorithm logic as I was implementing a security-enabled client.  I've got a client that can establish a secured UA Session but not without some ugly hacks to the uacp/uasc packages.  I'll put together some issues and/or some small PRs for this.

===
Corrected a misinterpretation of the symmetric encryption key
generation algorithm.
Also added method SignatureLength() to assist the caller in
determining how large the signature suffix of an encrypted
message is

### DIFF
--- a/securitypolicy/cryptoRsaOaep.go
+++ b/securitypolicy/cryptoRsaOaep.go
@@ -14,7 +14,7 @@ import (
 	_ "crypto/sha256"
 )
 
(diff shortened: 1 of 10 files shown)


wmnsk:
Looks nice. Thank you @dwhutchison!

PR #139, opened 2019-02-26
test with go1.12 on circleci

magiconair:
circle-ci does not have a go1.12 container yet.


wmnsk:
I’d like to keep this package follow the policy that golang itself follows officially. 
https://golang.org/doc/devel/release.html#policy

So, at the time CCI become compatible with Go 1.12, the task for 1.10 should be removed.

magiconair:
Dropped support for go1.10

wmnsk:
Thanks!

For GO111MODULES, `on` should be quoted not to be interpreted as true by yaml.

```yaml
        environment:
          GO111MODULE: "ON"
```

https://circleci.com/blog/go-v1.11-modules-and-circleci/

magiconair:
Was just looking at that. :)

wmnsk:
Reran CCI.
Now the container for 1.12 is available but it seems we’re facing another problem...

magiconair:
I've dropped the linter since you are using the GolangCI bot. Also dropped the working directory since checking out to the home dir will force the new module behavior.

wmnsk:
Looking good, thank you!

PR #149, opened 2019-03-17
service struct code generation

magiconair:
This patch contains the code generator which I've written to generate the service structs. The current code reads the type definitions from the schema and generates the enums. I'll add the service structs tomorrow.


cmd/service/main.go
```diff
@@ -0,0 +1,113 @@
...
+
+	path := filepath.Join(out, "enums_gen.go")
+	if err := ioutil.WriteFile(path, src, 0644); err != nil {
+		log.Fatalf("Failed to write %s: %v", path, err)
+	}
+	log.Printf("Wrote enums to %s", path)
+
+}
+
+func snakeCase(s string) string {
```
golangcibot: `snakeCase` is unused (from `deadcode`)

cmd/service/main.go
```diff
@@ -0,0 +1,112 @@
...
+
+	path := filepath.Join(out, "enums_gen.go")
+	if err := ioutil.WriteFile(path, src, 0644); err != nil {
+		log.Fatalf("Failed to write %s: %v", path, err)
+	}
+	log.Printf("Wrote enums to %s", path)
+
+}
+
+// func snakeCase(s string) string {
```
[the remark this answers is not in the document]
magiconair: Yeah, if I really need multiple files. If the structs fit into a single file we might not need it.

cmd/service/main.go
```diff
@@ -0,0 +1,271 @@
...
+
+		objects = append(objects, o)
+	}
+	return objects
+}
+
+type Kind int
+
+const (
+	KindInvalid Kind = iota
```
golangcibot: `KindInvalid` is unused (from `deadcode`)

magiconair:
This patch generates all available OPC/UA extension objects and enumerations. All generated files now have a `_gen.go` suffix, are generated with `go generated` and also removed before generation.

I have grouped all enums into a single file and the extension objects into separate files. However, the extension objects only consist of the struct itself. I would prefer to keep them all in a single `extobjs_gen.go` file and have separate tests.

I would also like to keep the existing documentation but that would need to be moved to a separate file, e.g. a Markdown file with separate headers so that the code generator can find the corresponding section by name.

Translating the code to the generated structs revealed some bugs in the current code:

1. `CloseSessionRequest` had a session id variable which is not in the type definition. Since this is the last message sent it did not matter.
2. The `RequestedTimeout` and `RevisedTimeout` values in the `OpenSecureChannelRequest/Response` are `float64` and not `uint64`.
3. `ServersOnNetwork` should be `ServerOnNetwork`
4. Some enum definition names were a bit off. 
5. `AdditionalHeader` does not exist as a separate type but is just an `ExtensionObject`.

Also, I've updated the generated names to Go standards like `ID`, `URL`, `URI`, `XML`, ...

All tests pass and I have only commented out the existing struct definitions. I will remove them and the constructors before merging however, that requires touching all the tests again.

magiconair:
Not done with the constructor replacement. I'm about half way through.

magiconair:
I'm going to clean up a bit more. Will remove all the commented code now.

Struggling a bit on how to handle structs with Encoding masks but I think the approach I have now in `DataValue` seems to work. Define constants which can be or`ed, provide a single `Has()` method to check a set of flags and an `UpdateMask()` function to update the mask in one fell swoop. It is still the responsibility of the caller to set the mask accordingly.

The other thing is that in Go the zero value should be useful and having to set the `AdditionalHeader` field to an empty `ExtensionObject` feels off. We can add that to the `Encode()` method to add it but then we are mixing binary encoding with the OPC/UA protocol. It is the protocol that requires setting the field and not the encoder unless I'm missing something.

magiconair:
This is done now and completes the cleanup and codec refactor. 

If there are no objections then I would like to merge this.

wmnsk:
Good job!

PR #150, opened 2019-03-20
StatusCode implements Error

magiconair:
This patch generates the status code values to implement the error interface.


wmnsk:
Looks elegant.

PR #155, opened 2019-03-27
Make AttributeID a type and add String() method for all enums

magiconair:
This patch improves the enums as follows:

* Rename `Type` to `TypeID` for the built-in types and make it a typed enum type.
* Goify `TypeID` names `TypeIDGUID`, `TypeIDXMLElement`, `TypeIDNodeID` and `TypeIDExtendedNodeID`
* Make `AttributeID` a type and have the code generator use it for all `AttributeID` fields instead of `uint32`
* Add `String()` method for all enums with golang.org/x/tools/cmd/stringer


ua/enums.go
```diff
@@ -10,44 +10,40 @@ const (
...
+	AttributeIDExecutable              AttributeID = 21
+	AttributeIDUserExecutable          AttributeID = 22
+	AttributeIDDataTypeDefinition      AttributeID = 23
+	AttributeIDRolePermissions         AttributeID = 24
+	AttributeIDUserRolePermissions     AttributeID = 25
+	AttributeIDAccessRestrictions      AttributeID = 26
+	AttributeIDAccessLevelEx           AttributeID = 27
 )
 
 // datatypes
```
wmnsk: Maybe it's safer to add explicit type(uint8) here?

magiconair: Since all other enums are typed and generated from the schema definition attribute ids and types are the only outliers.

magiconair: Also, the `AttributeID` field in all service objects is an `uint32`

PR #156, opened 2019-03-28
WIP: uasc securitypolicies - testing version of security policy client code

dwhutchison:
** DO NOT MERGE YET **

This is a first draft of the security implementation.  I'm submitting this PR to open it up
for testing and comments.  This is **not** a complete implementation yet.

Sample code on connecting to a server is in examples/crypto

I've tested Basic128Rsa15 and Basic256 policies against Kepware and Ignition (Basic128Rsa15 only).
Kepware was very particular about the format of the client certificate so any testing with other servers
with different policy support will help greatly.

A few things to note,
- Server logic is not written yet and old functions haven't been tested; it may have broken completely
- Testing has only been done with None, Basic128Rsa15, and Basic256 policies
- Changes to the UASC headers broke testing; fixing the testing will
  follow implementing the server code in case further changes are necessary
- There is a lot of code cleanup to occur; commented out struct fields, etc, will disappear eventually

Please review and test and let me know of any comments.

### DIFF
--- a/client.go
+++ b/client.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/gopcua/opcua/keyring"
 	"github.com/gopcua/opcua/ua"
 	"github.com/gopcua/opcua/uacp"
 	"github.com/gopcua/opcua/uasc"
@@ -32,7 +33,12 @@ func (c *Client) Open() error {
 	if err != nil {
 		return err
 	}
-	sechan := uasc.NewSecureChannel(conn, nil)
+
+	if c.config == nil {
+		c.config = uasc.NewClientConfigSecurityNone(3600000)
+	}
+
+	sechan := uasc.NewSecureChannel(conn, c.config)
 	if err := sechan.Open(); err != nil {
 		conn.Close()
 		return err
@@ -40,10 +46,32 @@ func (c *Client) Open() error {
 	sechan.EndpointURL = c.Addr
 	c.sechan = sechan
 
+	// todo(dh): Should probably be able to use a different cert than the secchan cert
+	var appURI string
+	if c.config.LocalThumbprint != nil {
+		cert, err := keyring.Certificate(c.config.LocalThumbprint)
+		if err == nil && len(cert.URIs) > 0 {
+			appURI = cert.URIs[0].String()
+		}
+	}
+
+	var tokenID *ua.AnonymousIdentityToken
+	for _, t := range c.config.ServerEndpoint.UserIdentityTokens {
+		// todo(dh): Allow more than anonymous authentication eventually
+		if t.TokenType == ua.UserTokenTypeAnonymous {
+			tokenID = &ua.AnonymousIdentityToken{PolicyID: t.PolicyID}
+			break
+		}
+	}
+	if tokenID == nil {
+		tokenID = &ua.AnonymousIdentityToken{PolicyID: "Anonymous"}
+	}
+
 	// todo(fs): this should probably be configurable.
 	sessionCfg := uasc.NewClientSessionConfig(
+		appURI,
 		[]string{"en-US"},
-		&ua.AnonymousIdentityToken{PolicyID: "open62541-anonymous-policy"},
+		tokenID,
 	)
 
 	session := uasc.NewSession(sechan, sessionCfg)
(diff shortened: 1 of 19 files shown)


magiconair:
I've installed WIP to the org so that adding `WIP` to the title of a PR prevents it from being merged.

wmnsk:
Using draft PR feature introduced recently can be an option for contributors.
https://github.blog/2019-02-14-introducing-draft-pull-requests/

magiconair:
I haven’t found a way to retroactively make a PR a draft PR. If that’s possible then we don’t need this. 

—
Frank Schröder

dwhutchison:
Interesting, I didn't realize draft PRs were a thing.  Let me know if you'd prefer I close this one and reopen it as a draft PR; that seems to be more appropriate and I have no problems with that.

wmnsk:
@dwhutchison
No need for that. It’s just for your information, hoping it helps you next time 😃

client.go
```diff
@@ -134,6 +162,49 @@ func (c *Client) Browse(req *ua.BrowseRequest) (*ua.BrowseResponse, error) {
 	return res, err
 }
 
+// GetEndpoints reads all available endpoints from the server at endpoint
+func GetEndpoints(endpoint string) ([]*ua.EndpointDescription, error) {
```
magiconair, line 166: I'll refactor the client to split the `Open()` method to just connect with a secure channel and then to create a session separately. Then we can also add `GetEndpoints()` as a method to client and make this function a convenience helper.

client.go
```diff
@@ -134,6 +162,49 @@ func (c *Client) Browse(req *ua.BrowseRequest) (*ua.BrowseResponse, error) {
...
+	err = c.sechan.Send(req, func(v interface{}) error {
+		r, ok := v.(*ua.GetEndpointsResponse)
+		if !ok {
+			return fmt.Errorf("invalid response: %T", v)
+		}
+		res = r
+		return nil
+	})
+
+	c.Close()
```
magiconair, line 204: Usually, you call `defer Close()`with after `Open` like below. Wait until I've pushed the client refactor.

```go
c, err := Open(...)
if err != nil {
    return err
}
defer c.Close()
```

examples/crypto/crypto.go
```diff
@@ -0,0 +1,104 @@
...
+	v, err := c.Node(ua.NewNumericNodeID(0, 2258)).Value()
+	if err != nil {
+		log.Fatal(err)
+	}
+	log.Printf("time: %v", v.Value)
+}
+
+func getCert() (*x509.Certificate, *rsa.PrivateKey, error) {
+
+	certPEM, err := ioutil.ReadFile("cert.pem")
```
magiconair, line 71: Can you use https://golang.org/pkg/crypto/tls/#LoadX509KeyPair for this?

keyring/keyring.go
```diff
@@ -0,0 +1,97 @@
...
+	"sync"
+)
+
+type privateKeys struct {
+	thumbprint []byte
+	cert       *x509.Certificate
+	key        *rsa.PrivateKey
+}
+
+var localKeys sync.Map
```
magiconair, line 25: I suggest to make this a `struct Keyring` and the function as methods. Global variables make testing harder.

securitypolicy/securitypolicy.go
```diff
@@ -23,15 +23,16 @@ import (
...
-	blockSize       int
-	minPadding      int
-	encrypt         func(cleartext []byte) (ciphertext []byte, err error)
-	decrypt         func(ciphertext []byte) (cleartext []byte, err error)
-	signature       func(message []byte) (signature []byte, err error)
-	verifySignature func(message, signature []byte) error
-	signatureLength int
-	encryptionURI   string
-	signatureURI    string
+	blockSize           int
```
magiconair, line 26: Can you refactor this as a struct with methods instead of function variables?

uasc/secure_channel.go
```diff
@@ -262,6 +415,9 @@ func (s *SecureChannel) recv() {
 			if err == io.EOF {
 				return
 			}
+			if err != nil {
+				log.Fatal(err)
```
magiconair, line 419: `log.Fatal` might be a problem here.

uasc/secure_channel.go
```diff
@@ -51,6 +56,12 @@ type SecureChannel struct {
 	// handle which is part of the Request and Response headers.
 	mu      sync.Mutex
 	handler map[uint32]chan Response
+
+	// Encryption Algorithms for sending and receiving messages
+	// During the OpenSecureChannel phase, this contains the Asymmetric Algorithms
+	// After both secure nonces are received from the OpenSecureChannel phase,
+	// this is replaced by the Symmetric algorithm methods for all other communication.
+	encryptionAlgorithm *securitypolicy.EncryptionAlgorithm
```
magiconair, line 64: Maybe create a package `crypto` and move the `Keyring` and `EncryptionAlgorithm` in there. Also, I'd like to find a shorter name for `encryptionAlgorithm`. How about `enc`?

magiconair (commented):
Added some style comments. Will review more thoroughly when it is ready.

dwhutchison:
Thanks for the comments; I'll work through them.
I found that the Ignition 8 Beta that's out implements the remaining security policies so I was able to test against them all.  I've pushed a small fix to clean up issues I found.
I've successfully connected to the following policies:
http://opcfoundation.org/UA/SecurityPolicy#None
http://opcfoundation.org/UA/SecurityPolicy#Basic128Rsa15
http://opcfoundation.org/UA/SecurityPolicy#Basic256
http://opcfoundation.org/UA/SecurityPolicy#Basic256Sha256
http://opcfoundation.org/UA/SecurityPolicy#Aes256_Sha256_RsaPss
http://opcfoundation.org/UA/SecurityPolicy#Aes128_Sha256_RsaOaep

magiconair:
Cool. I'll push the client refactor tomorrow morning. Too late today.

keyring/keyring.go
dwhutchison: Yea, that bothered me from the start.  I'm debating if this is actually required at all.  I'll put this in the securitypolicy library and try to find a good spot for this struct.

securitypolicy/securitypolicy.go
dwhutchison: I'll try refactor this into interfaces and see how it looks.  I originally liked the closure concept because it allowed the encryption to be initialized without keeping the keys stored in config objects, but I admit it's not as clear as it could be.

uasc/secure_channel.go
dwhutchison: Yep; remnant of debugging that I forgot to remove.  This will disappear.

uasc/secure_channel.go
dwhutchison: package crypto is used in the standard library so I'd like to avoid that name (which is used extensively in the securitypolicy package.  Open to alternative ideas.
enc works for the name variable.  I'll change that.

dwhutchison:
Closing this; cleaned up version in: 
https://github.com/gopcua/opcua/tree/feature/crypto-client

PR #160, opened 2019-04-04
Cleanup to prep for crypto

dwhutchison:
A few small updates I wanted incorporated as I work on the server side of the crypto code.  Currently restructuring a lot of secure channel logic to allow for incoming server connections as well as outgoing client connections. These are some of the simpler changes from #156 that won't be affected by future secure channel changes.

### DIFF
--- a/uasc/asymmetric_security_header.go
+++ b/uasc/asymmetric_security_header.go
@@ -51,3 +51,14 @@ func (a *AsymmetricSecurityHeader) String() string {
 		a.ReceiverCertificateThumbprint,
 	)
 }
+
+// Len returns the Header Length in bytes.
+func (h *AsymmetricSecurityHeader) Len() int {
+	var l int
+	l += 12
+	l += len(h.SecurityPolicyURI)
+	l += len(h.SenderCertificate)
+	l += len(h.ReceiverCertificateThumbprint)
+
+	return l
+}
--- a/uasc/symmetric_security_header.go
+++ b/uasc/symmetric_security_header.go
@@ -35,9 +35,14 @@ func (h *SymmetricSecurityHeader) Encode() ([]byte, error) {
 }
 
 // String returns Header in string.
-func (s *SymmetricSecurityHeader) String() string {
+func (h *SymmetricSecurityHeader) String() string {
 	return fmt.Sprintf(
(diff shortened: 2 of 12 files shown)


uasc/symmetric_security_header.go
```diff
@@ -35,9 +35,14 @@ func (h *SymmetricSecurityHeader) Encode() ([]byte, error) {
...
+func (h *SymmetricSecurityHeader) String() string {
 	return fmt.Sprintf(
 		"TokenID: %d",
-		s.TokenID,
+		h.TokenID,
 	)
 }
+
+// Len returns the Header Length in bytes.
+func (h *SymmetricSecurityHeader) Len() int {
```
magiconair, line 46: Do we still need the `Len()` methods? I've removed them from all other structs.

uasc/asymmetric_security_header.go
```diff
@@ -51,3 +51,14 @@ func (a *AsymmetricSecurityHeader) String() string {
 		a.ReceiverCertificateThumbprint,
 	)
 }
+
+// Len returns the Header Length in bytes.
+func (h *AsymmetricSecurityHeader) Len() int {
```
magiconair, line 56: Do we still need the `Len()` methods? I've removed them from all other structs.

uasc/symmetric_security_header.go
dwhutchison: Yes; I’ll need these two for the crypto to figure out which bytes to encrypt and which to leave plaintext. 
The message header and the (A)symmetricSecurityHeaders stay plaintext and the rest is encrypted so I need a way to find that barrier in the encoded byte stream. 
Obviously the Asymmertic one is more important as it’s variable length but having both makes the code easier to understand.

magiconair:
@dwhutchison I've invited you to the org so that we can work on a branch.

magiconair:
Also, let me know if you need help on the secure channel logic for the server. Curious what kind of changes are required.

dwhutchison:
Awesome, thanks. 
I’ll put an issue together with some of the things I noticed and an outline of what I’m thinking then push a WIP branch

dwhutchison:
Closing; superseded by #166

Issue #161, opened 2019-04-05
Discussion on crypto support


dwhutchison:
Creating this issue for a place for discussion outside of a PR.

While trying to implement the crypto functions for the server path, I noticed a few functions which didn't work for receiving messages.  Primarily, the `recv()` and `SendAsync()` functions within the uasc package were tailored for sending client messages and not capable of receiving messages.  The majority of the code is fine and the problems weren't major, but they do need some adjustment.  For example, the SendAsync method would panic while trying to copy the `reqHdr` field into a ResponseHeader and the recv method could only handle messages with existing handlers (which are tied to client-generated requestIDs).

I'm preparing a PR for three things:
1. Generalize the above uasc functions to allow for processing both 'Request' and 'Response' types.
2. Create an initial server application layer to receive and process unsolicited messages.
3. Incorporate the crypto functions into the transport layer
I'm working on all three simultaneously but if I can extract parts of each, I'll try publish them individually.

Within the server logic I'm working on a concept of a 'Channel Broker' to manage multiple secure channels on one listener.  With this, I'm hoping to have a method of accepting connections, creating secure channels, and handling requests from multiple connections as well as slightly decoupling the secure channel from the `uacp.Conn`.  This is because clients are allowed to reconnect to a server provided the secure channel hasn't expired so this should allow for a means to 'reattach' a new `Conn` to an existing secure channel.  Having a system monitoring all active secure channels will also allow for a means to prune expired connections.

I anticipate client logic remaining the same as it is now; i.e. the application controlling the creation and use of SecureChannels.  One thing I'm aware of, but unsure exactly how to implement it at this time, is that clients will also need a way to reconnect to a server utilizing an existing SecureChannel if the uacp.Conn is broken.  I don't anticipate this being too difficult (try to send, fail, close the uacp.Conn, open a new one, replace the old conn reference with the new one, continue on).  Also need to check the spec to see if the full HEL/ACK sequence needs to be repeated in this case.

#156  was a proof of concept that the crypo functions themselves are functional, but will need more work to allow for both sending and receiving messages, decent secret management, etc.  Now that I've looked closer at what it'll take to implement the server component of the logic, I should be able to clean up that sequencing.

Comments are welcome; when I have things cleaned up a bit I'll push a WIP branch for review.

magiconair:
Based on experience with previous complex PRs I suggest to try hard to separate them. The sequence you have outlined sounds like three logical independent PRs. Keep the changes in separate branches which are based on another. 

[...]

PR #162, opened 2019-04-06
refactor config to use options

magiconair:
This patch refactors the client configuration to include both the secure channel and session configuration and to configure the client with options.

See #161

### DIFF
--- a/client.go
+++ b/client.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"crypto/rand"
 	"fmt"
-	mrand "math/rand"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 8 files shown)


magiconair:
This configures some options of the configuration but is not complete. Should be good enough for now though and it is easy to expand. See `config.go` in the main package for options.

magiconair:
`datetime` does not work so something is different.

magiconair:
Found it. The `UserDataSignature` in the session config was nil. I think the penny finally dropped on how to solve this in a generic way. If a field is a struct and is `nil` the encoder must encode the empty value. Then we can finally drop these empty fields.

PR #171, opened 2019-04-19
Honor session timeout

### LINKED ISSUE #170: RequestedSessionTimeout not set in func (c *Client) CreateSession
I was tring  to connect to an Siemens S7, using the datetime example I was unable to create a session.. 

the error returned was `StatusBadSessionIDInvalid`. 
so logging the tcp packets with wireshark and matching the results with a working client I noticed that the message `CreateSessionRequest `has a
`opcua.RequestedSessionTimeout = 0 `
so the S7 replied with 
`aopcua.RevisedSessionTimeout=1`

when at next step we call `ActivateSession `the session is already expired.

in the source code `client.go` I noticed that   in 
`func (c *Client) CreateSession` the `RequestedSessionTimeout `of the struct `ua.CreateSessionRequest` is not set  so it has 0 value.

for me worked adding `RequestedSessionTimeout: cfg.SessionTimeout`
 ```
req := &ua.CreateSessionRequest{

		ClientDescription:       cfg.ClientDescription,

		EndpointURL:             c.endpointURL,

		SessionName:             fmt.Sprintf("gopcua-%d", time.Now().UnixNano()),

		ClientNonce:             nonce,

		ClientCertificate:       c.cfg.Certificate,

		RequestedSessionTimeout: cfg.SessionTimeout,
	}
```

magiconair:
This patch honors the session timeout when creating a
session. It also changes the type of the SessionTimeout
from `float64` to a `time.Duration` and sets the default to 
60 seconds.

Fixes #170

### DIFF
--- a/client.go
+++ b/client.go
@@ -164,11 +164,12 @@ func (c *Client) CreateSession(cfg *uasc.SessionConfig) (*Session, error) {
 	}
 
 	req := &ua.CreateSessionRequest{
-		ClientDescription: cfg.ClientDescription,
-		EndpointURL:       c.endpointURL,
-		SessionName:       fmt.Sprintf("gopcua-%d", time.Now().UnixNano()),
-		ClientNonce:       nonce,
-		ClientCertificate: c.cfg.Certificate,
+		ClientDescription:       cfg.ClientDescription,
+		EndpointURL:             c.endpointURL,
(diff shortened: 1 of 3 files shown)


magiconair:
After the discussion in #170 we're setting the default session timeout to 60 seconds aka one minute :)

PR #175, opened 2019-04-25
handle MSGA

magiconair:
patch for handling MSGA chunks. We really need a mechanism to test this ...


magiconair:
Standard examples still work.

PR #179, opened 2019-04-29
Publish

ghost:
Publish request. Tested on our server.
CreateSubscriptionRequest and MonitoredItemCreateRequest will follow.

### DIFF
--- a/client.go
+++ b/client.go
@@ -7,7 +7,6 @@ package opcua
 import (
 	"context"
 	"crypto/rand"
-	"errors"
 	"fmt"
 	"log"
 	"sync"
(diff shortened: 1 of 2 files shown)


magiconair:
Thank you. What exactly is your server and can you also provide a working example in the `examples` folder?

ghost:
This is a proprietary server, for our own needs.
Added an example.

examples/publish/publish.go
```diff
@@ -0,0 +1,51 @@
...
+	endpoint := flag.String("endpoint", "opc.tcp://localhost:4840", "OPC UA Endpoint URL")
+	flag.BoolVar(&debug.Enable, "debug", false, "enable debug logging")
+	flag.Parse()
+	log.SetFlags(0)
+
+	c := opcua.NewClient(*endpoint)
+	if err := c.Connect(); err != nil {
+		log.Fatal(err)
+	}
+	defer func() {
```
magiconair: `defer c.Close()` is good enough.

examples/publish/publish.go
```diff
@@ -0,0 +1,51 @@
...
+		_ = c.Close()
+	}()
+
+	var ch = make(chan opcua.PublishNotificationData)
```
magiconair: `ch := make(chan opcua.PublishNotificationData)`

examples/publish/publish.go
```diff
@@ -0,0 +1,51 @@
...
+	go c.Publish(ch)
+
+	for {
+		var response = <-ch
+
+		var t = time.Now().Format(time.RFC3339)
+		if response.Error != nil {
+			log.Printf("%s - %s \n", t, response.Error.Error())
```
magiconair: I think `log.Printf("%s - %v", t, response.Error)` should also work.

examples/publish/publish.go
```diff
@@ -33,7 +33,7 @@ func main() {
 
 		var t = time.Now().Format(time.RFC3339)
```
magiconair: `t := time.Now().Format(time.RFC3339)`. Also drop the `\n` from the `log.Printf`

magiconair: might also want to rename `response` to `resp`

ghost:
Anything else? Do i need to remove `var ...` from `client.go`?

magiconair:
In parts this looks a bit like C code ported to Go. I'll do a quick pass and cleanup. Could you re-test after that?

ghost:
Yes, no problem.

magiconair:
Actually, I can't since I can't push to your branch :( I miss gerrit ...

As for style: lets keep it uniform to simplify code review and keep it in line with other Go code.

client.go
```diff
@@ -426,3 +426,99 @@ func (c *Client) Subscribe(intv time.Duration) (*Subscription, error) {
...
+		}
+
+		// Prepare SubscriptionAcknowledgement for next PublishRequest
+		var acks []*ua.SubscriptionAcknowledgement
+		for _, i := range res.AvailableSequenceNumbers {
+			ack := &ua.SubscriptionAcknowledgement{
+				SubscriptionID: res.SubscriptionID,
+				SequenceNumber: i,
+			}
+			acks = append(acks, ack)
```
golangcibot, line 483: SA4010: this result of append is never used, except maybe in other appends (from `staticcheck`)

magiconair:
@wsnotify I've fixed the code a bit and pushed the type switch to the client since it does not make any difference whether you do this in the `Publish` method or later in the client. But this way the code in `client.go` is more generic.

magiconair:
Also, I'm having some issues checking out your `Publish` branch. I'm getting into detached head mode. I can make modifications but I'm not sure what happens here. Maybe I'm doing something wrong.

client.go
```diff
@@ -429,113 +428,95 @@ func (c *Client) Subscribe(intv time.Duration) (*Subscription, error) {
...
 }
 
-func (c *Client) Publish(dataChan chan<- PublishNotificationData) {
-	var (
-		subAck        = make([]*ua.SubscriptionAcknowledgement, 0)
-		pubRequest    = &ua.PublishRequest{}
-		pubResponse   *ua.PublishResponse
-		responseError ua.StatusCode
-		err           error
-	)
```
ghost: We can reuse this vars in for loop for better performance.

client.go
```diff
@@ -429,113 +428,95 @@ func (c *Client) Subscribe(intv time.Duration) (*Subscription, error) {
...
 		}
 
-		if pubResponse.NotificationMessage == nil {
-			dataChan <- PublishNotificationData{
-				Error:          errors.New("empty NotificationMessage"),
-				SubscriptionID: pubResponse.SubscriptionID,
+		if res.NotificationMessage == nil {
+			notif <- PublishNotificationData{
+				SubscriptionID: res.SubscriptionID,
+				Error:          fmt.Errorf("empty NotificationMessage"),
```
ghost, line 489: `errors.New()`?

client.go
magiconair: Did you test that this is actually the case or do you just assume that? Generally, you don't declare variables upfront in Go but declare them closer where they are being used. We also don't know whether the compiler is already optimizing this therefore you should profile this if you are concerned.

client.go
magiconair: Simplicity here as well. The code uses `fmt.Errorf` in some places already and there is no win in using a different approach for some cases.

ghost:
Tested, works for me.

magiconair:
Thank you!

PR #182, opened 2019-05-06
Fix Invalid PolicyID on Anonymous connections

dwhutchison:
After sending a `CreateSessionRequest`, check if the UserIdentityToken was left unset by the user and set it to anonymous auth with a valid policy ID from the server.

### DIFF
--- a/client.go
+++ b/client.go
@@ -180,12 +180,12 @@ func (c *Client) CreateSession(cfg *uasc.SessionConfig) (*Session, error) {
 			return nil
 		}
 
-	// Ensure we have a valid identity token that the server will accept before trying to activate a session
+		// Ensure we have a valid identity token that the server will accept before trying to activate a session
 		if c.sessionCfg.UserIdentityToken == nil {
 			opt := AuthAnonymous()
 			opt(c.cfg, c.sessionCfg)
 
-			p := anonPolicyID(resp.ServerEndpoints)
+			p := anonymousPolicyID(resp.ServerEndpoints)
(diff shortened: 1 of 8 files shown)


client.go
```diff
@@ -187,6 +180,16 @@ func (c *Client) CreateSession(cfg *uasc.SessionConfig) (*Session, error) {
 			return nil
 		}
 
+	// Ensure we have a valid identity token that the server will accept before trying to activate a session
```
magiconair: Indent seems off

client.go
```diff
@@ -199,6 +202,22 @@ func (c *Client) CreateSession(cfg *uasc.SessionConfig) (*Session, error) {
 	return s, err
 }
 
+const defaultAnonymousPolicyID = "Anonymous"
+
+func anonPolicyID(endpoints []*ua.EndpointDescription) string {
+	for _, e := range endpoints {
+		if e.SecurityMode == ua.MessageSecurityModeNone && e.SecurityPolicyURI == ua.SecurityPolicyURINone {
```
magiconair: Two nitpicks:

1. Either `anon` or `Anonymous` but not both
1. How about flipping the `if` to avoid one level of nesting?

```go
for _, e := range endpoints {
    if e.SecurityMode != ua.MessageSecurityModeNone || 
        e.SecurityPolicyURI != ua.SecurityPolicyURINone {
        continue
    }
    for _, t := range e.UserIdentityToken { ... }
}
```

magiconair:
LGTM

PR #188, opened 2019-05-08
fix: return error status from Node.Attribute()

alexbrdn:
return error status from Node.Attribute() to prevent nil dereference
on cast in BrowseName() and DisplayName() when e.g. node not found

### DIFF
--- a/node.go
+++ b/node.go
@@ -65,9 +65,15 @@ func (n *Node) Attribute(attrID ua.AttributeID) (*ua.Variant, error) {
 		return nil, err
 	}
 	if len(res.Results) == 0 {
-		return nil, nil
+		// #188: we return StatusBadUnexpectedError because it is unclear, under what
+		// circumstances the server would return no error and no results in the response
+		return nil, ua.StatusBadUnexpectedError
 	}
-	return res.Results[0].Value, nil
+	value := res.Results[0].Value
+	if res.Results[0].Status != ua.StatusOK {
+		return value, res.Results[0].Status
+	}
+	return value, nil
 }
 
 // References retrns all references for the node.


node.go
```diff
@@ -65,9 +66,13 @@ func (n *Node) Attribute(attrID ua.AttributeID) (*ua.Variant, error) {
 		return nil, err
 	}
 	if len(res.Results) == 0 {
-		return nil, nil
+		return nil, fmt.Errorf("no values for %v", rv)
```
magiconair: Do we have a status code for this?

magiconair: Maybe `StatusBadNotFound` ?

magiconair: or `StatusBadNodeAttributesInvalid` ?

magiconair: or `StatusBadAttributeIDInvalid` ?

alexbrdn: Frankly, I am not sure under what circumstances this can happen at all. This line is executed when the `Read()` returned no error, but the `Results` are empty. I don't know if we can assume that attribute id is invalid or the attribute couldn't be found. One would presume that in those cases the server would reply with an appropriate status.

If anything, I would probably go with `StatusBadNotFound`. Are you sure it makes sense to assign a status code in this case?

magiconair: Apparently, this happens when the node is not found. If we don't use status codes then the error message becomes part of the public API and users will have to check for that. Lets stick to status codes unless there is a good reason not to.

Also, if this happens when the node is not found then `StatusBadNotFound` probably makes the most sense. I would add a comment explaining the reasoning in case we need to revisit this, e.g.

```go
// #188: we return StatusBadNotFound since the assumption is that this condition only 
// occurrs when the node is not found. Otherwise, StatusBadAttributeIDInvalid might
// also be an option.
```

Maybe you can have a quick look what other SDKs are returning in that case, e.g. github.com/open62541 and github.com/FreeOPCUA

alexbrdn: Actually, the server I am working with does return a result with `StatusBadNotFound` when the node is not found. That was the primary reason I have added the check `res.Results[0].Status != ua.StatusOK`. But I see your point about sticking to status codes.

[open62541](https://github.com/open62541/open62541/blob/master/src/client/ua_client_highlevel.c#L373) returns `StatusBadUnexpectedError`. [freeopcua](https://github.com/FreeOpcUa/freeopcua/blob/master/src/clientapp/opcua_main.cpp#L629) appears to ignore the matter altogether (assuming I have found the correct place in code during my quick search)

What do you think about going with `StatusBadUnexpectedError`?

magiconair: works for me.

alexbrdn: done

magiconair:
👍

PR #196, opened 2019-05-16
Issue #185: ensure uniqueness of request ids to prevent incorrect response to handler assignment

### LINKED ISSUE #185: I get ua.BrowseResponse on ua.PublishRequest
Goroutine 1: Sending PublishRequest and waiting for response
Goroutine 2: Sending BrowseRequest and waiting for response
Goroutine 1: Received ua.BrowseResponse
Goroutine 2: waiting

Seems like related to SequenceNumber in #184

alexbrdn:
- increase `SecureChannel.cfg.RequestId` on each request, otherwise all requests have the same request id and the responses to parallel requests may get assigned incorrectly
- drop decrementing the counters:
  * multiple parallel requests may get the same ids
  * SequenceId is supposed to be monotonically rising anyways
- split and reduce the block protected by mutex (see #186). There's no need to encode the message and write to secure channel in protected block
- add a fail-safe check to prevent overriding response handler if there is already a handler waiting for response with the same id

closes gopcua/opcua#185

### DIFF
--- a/uasc/secure_channel.go
+++ b/uasc/secure_channel.go
@@ -230,34 +230,31 @@ func (s *SecureChannel) SendAsync(svc interface{}, authToken *ua.NodeID) (resp c
 	}
 
 	s.mu.Lock()
-	defer s.mu.Unlock()
-
 	// the request header is always the first field
 	val := reflect.ValueOf(svc)
 	val.Elem().Field(0).Set(reflect.ValueOf(s.reqhdr))
-	// update counters and reset them on error
+	// update counters
 	s.cfg.SequenceNumber++
+	s.cfg.RequestID++
 	s.reqhdr.AuthenticationToken = authToken
 	s.reqhdr.RequestHandle++
 	s.reqhdr.Timestamp = time.Now()
-	defer func() {
-		if err != nil {
-			s.cfg.SequenceNumber--
-			s.reqhdr.RequestHandle--
-		}
-	}()
 
 	// encode the message
 	m := NewMessage(svc, typeID, s.cfg)
+	reqid := m.SequenceHeader.RequestID
+	s.mu.Unlock()
 	b, err := m.Encode()
 	if err != nil {
 		return nil, err
 	}
-	reqid := m.SequenceHeader.RequestID
 
 	// encrypt the message prior to sending it
 	// if SecurityMode == None, this returns the byte stream untouched
 	b, err = s.signAndEncrypt(m, b)
+	if err != nil {
+		return nil, err
+	}
 
 	// send the message
 	if _, err := s.c.Write(b); err != nil {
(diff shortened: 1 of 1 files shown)


alexbrdn:
@magiconair, as I am very new to the matter, I appreciate your feedback. My understanding of the existing code and the problem at hand might be way off, so don't hesitate to tear it apart.

uasc/secure_channel.go
```diff
@@ -267,9 +261,12 @@ func (s *SecureChannel) SendAsync(svc interface{}, authToken *ua.NodeID) (resp c
 
 	// register the handler
 	resp = make(chan Response)
-	// s.mu.Lock()
+	s.mu.Lock()
+	if _, ok := s.handler[reqid]; ok {
```
magiconair: you can simplify this to 

```go
if s.handler[reqid] != nil {
    return nil, fmt.Errorf("error: duplicate handler registration for request id %d", reqid)
}

**Review by magiconair — COMMENTED** [review]
Small improvement. Otherwise, LGTM

**Comment by magiconair** [comment]
Also, that approach makes sense. Thank you!

**Comment by magiconair** [comment]
But that would mean that sending some of your requests had to generate an error.

File: uasc/secure_channel.go (hunk 1 above)
Comment (alexbrdn) [reply uasc/secure_channel.go]: good point. changed it

**Comment by alexbrdn** [comment]
>But that would mean that sending some of your requests had to generate an error.

No, none generated error. It's just that there were two requests sent in parallel and the handler got overwritten. So there were two goroutines waiting for response and only one handler. If the incorrect response arrived first, it produced an error. If the correct arrived first, then the second goroutine was left waiting forever for a response.

**Comment by magiconair** [comment]
I don't see how that would be possible since `SendAsync` holds the lock for the duration of the call including updating the counters and the map. How can two go routines get the same request id unless the first one fails and resets the counter. 

I've also just noticed that we're not handling the the error for 

```go
b, err = s.signAndEncrypt(m, b)
```

Could you please add that as well?

magiconair:
Without the lock from #186 this was possible but with the lock ...

magiconair:
I assume that with your fix you don't see the error anymore.

alexbrdn:
the `RequestId` is assigned from Config and it was never updated before this change. So _all_ requests had the same id. Only `SequenceNumber` and `RequestHandle` were incremented/decremented.

I'll add an error check.

alexbrdn:
That's true, I don't get the error anymore. I think I'll be able to release the code that produced the error as part of `examples/subscribe/subscribe.go` next week. I'm working on subscription functionality and this was a stumbling block.

magiconair:
Subscription would be awesome! Thanks for digging into this.

magiconair:
Aaaah, now I see it. duh. So the examples were too simple. :)

Issue #191, opened 2019-05-09
How to browse folder's in a NodeClass?


cngobd:
@magiconair I want to browse all available folder's  belong to  a  NodeClass,
my NodeClass index in Prosys Opc UA simulation server :
NodeClass :
       NamespaceIndex:       0
       NamespaceURI:          http://opcfoundation.org/UA/
       IdentifierType :            Numeric
       Identifier:                    85
How to config the browse.go in examples.  Please?

magiconair:
The example in `browse.go` is a starting point but not a generic example for browsing since the OPC/UA data model is too complex. You can use the recursive `browseChildren` function as a starting point but how you filter and what you output depends on your specific use case.

cngobd:
@magiconair Thank you for your help, now I want to browse all folders organized by root when I connect to the Prosys OpcUA server , is there any functions can implement it?
please

magiconair:
The code in `browse.go` should be a starting point but you have to find our yourself on how to do this. I don't have an example for this.

cngobd:
@magiconair which functions should I use to get the "user access level" of a node?
Please

magiconair:
@769139671 I've created #199 with the methods for `AccessLevel` and `UserAccessLevel`. I'll wait for some comments from the other developers but then we can merge this.

cngobd:
@magiconair Thank you!

magiconair:
I've just merged #199

PR #199, opened 2019-05-16
Add methods for AccessLevel and UserAccessLevel

magiconair:
This patch adds methods to get the access level and user access level of a node. Since the values can be masks I've also added `HasAccessLevel` and `HasUserAccessLevel` helpers. Not sure if that is a good idea since we then have to do this for all mask types. But maybe not. 

What do others think?

Addresses the question in https://github.com/gopcua/opcua/issues/191#issuecomment-491645673


magiconair:
I'll merge this and start to add tests for them in a separate PR.

PR #200, opened 2019-05-16
More browsing

magiconair:
This is a patch to improve browsing with a somewhat more complex example which works for my current use-case. I think it is time to get this merged. I'll clean this up and then press the button.

### DIFF
--- a/examples/browse/browse.go
+++ b/examples/browse/browse.go
@@ -9,12 +9,30 @@ import (
 	"flag"
 	"fmt"
 	"log"
+	"os"
+	"text/tabwriter"
 
 	"github.com/gopcua/opcua"
 	"github.com/gopcua/opcua/debug"
 	"github.com/gopcua/opcua/id"
 	"github.com/gopcua/opcua/ua"
+	"github.com/pkg/errors"
 )
+
+type NodeDef struct {
+	NodeID      *ua.NodeID
+	NodeClass   ua.NodeClass
+	BrowseName  string
+	Description string
+	AccessLevel ua.AccessLevelType
+	Path        string
+	DataType    string
+	Writable    bool
+}
+
+func (n NodeDef) String() string {
+	return fmt.Sprintf("%s\t%s\t%s\t%v\t%s", n.NodeID, n.Path, n.DataType, n.Writable, n.Description)
+}
 
 func join(a, b string) string {
 	if a == "" {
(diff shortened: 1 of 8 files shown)


node.go
```diff
@@ -116,16 +126,58 @@ func (n *Node) Attribute(attrID ua.AttributeID) (*ua.Variant, error) {
...
+	if refs == 0 {
+		refs = id.References
+	}
+	var nodes []*Node
+	res, err := n.References(refs, dir, mask, includeSubtypes)
+	if err != nil {
+		return nil, err
+	}
+	for _, r := range res {
+		nodes = append(nodes, n.c.Node(r.NodeID.NodeID))
```
alexbrdn, line 160: I wrote a browse example along the same lines to explore a little and have identified a problem with using this pattern: `client.Node(reference.NodeID.NodeID)`.
The `ExpandedNodeId` in `reference.NodeID` may have been actually expanded and contains a namespace uri and 0 as namespace index (as per the spec). Now, assigning `reference.NodeID.NodeID` to a `Node`'s id copies the mask which contains (in my case) NamespaceUri flag and ServerIndex flag. Encoding this NodeID in the BrowseRequest writes 0xC0 as DataEncoding byte into the stream. And the Server responds with StatusBadDecodingError, because the stream doesn't contain the promised namespace URI and ServerIndex and actually the ExpandedNodeId is not allowed in BrowseRequest anyway.

You can test this by browsing the Prosys OPC UA Simulation Server in default configuration. Funny thing is, the namespace that gets expanded is the OPC UA core namespace and AFAICT it's only this namespace that gets expanded. Nevertheless, I see this as a problem with the API as the Reference.NodeID.NodeID allows one to get hold of a NodeID, which is has a mask a NodeID is not supposed to have.

A possible solution would be to make `ExpandedNodeID.nodeID` private and add a function `ExpandedNodeID.NodeID()` to safely convert to NodeID or return an error. One problem I see is one would need a lookup to resolve a namespace URI to an index.

Maybe this needs to be extracted into an issue of its own.

kung-foo: ran into this exact issue yesterday

kung-foo: a super quick'n'dirty work around I used:

```diff
diff --git a/node.go b/node.go
index 5adf925..00fc9fd 100644
--- a/node.go
+++ b/node.go
@@ -156,6 +156,9 @@ func (n *Node) ReferencedNodes(refs uint32, dir ua.BrowseDirection, mask ua.Node
                return nil, err
        }
        for _, r := range res {
+               r.NodeID.NodeID.ClearIndexFlag()
+               r.NodeID.NodeID.ClearURIFlag()
+
                nodes = append(nodes, n.c.Node(r.NodeID.NodeID))
        }
        return nodes, nil
diff --git a/ua/node_id.go b/ua/node_id.go
index 8533e8c..d178f47 100644
--- a/ua/node_id.go
+++ b/ua/node_id.go
@@ -179,6 +179,10 @@ func (n *NodeID) SetURIFlag() {
        n.mask |= 0x80
 }
 
+func (n *NodeID) ClearURIFlag() {
+       n.mask &^= 0x80
+}
+
 // IndexFlag returns whether the Index flag is set in EncodingMask.
 func (n *NodeID) IndexFlag() bool {
        return n.mask&0x40 == 0x40
@@ -189,6 +193,10 @@ func (n *NodeID) SetIndexFlag() {
        n.mask |= 0x40
 }
 
+func (n *NodeID) ClearIndexFlag() {
+       n.mask &^= 0x40
+}
+
 // Namespace returns the namespace id. For two byte node ids
 // this will always be zero.
 func (n *NodeID) Namespace() uint16 {
```

magiconair:
Thanks @kung-foo and @alexbrdn. I'll try to wrap up the other open issues so that we can move forward.

magiconair:
rebased to master

AndreySerebrovC2M:
Hello! Will you merge it some day?

magiconair:
Hey @AndreySerebrovC2M, I'll clean this up and merge it. Thanks for reminding me.

examples/browse/browse.go
```diff
@@ -6,49 +6,164 @@ package main
...
-	"fmt"
 	"log"
+	"os"
+	"strconv"
 
 	"github.com/gopcua/opcua"
 	"github.com/gopcua/opcua/debug"
 	"github.com/gopcua/opcua/id"
 	"github.com/gopcua/opcua/ua"
+	"github.com/pkg/errors"
```
kung-foo: didn't want to use the 1.13 errors?

kung-foo: nm. commented on old commit ;)

magiconair: What would this look like with  go1.13 errors? You still don't get stack traces, do you? But thanks for noticing. I've dropped that for now.

magiconair:
This has been sitting here for too long. There is some overlap with #266 but I'll sort that out. Merging this now.

Issue #190, opened 2019-05-09
How to connect to the server with username & password?


cngobd:
I want to connect to my Prosys OpcUA server with User Authentication Methods: Username & Passwords, and with the Security Modes: None

```go
c:=opcua.NewClient(*endpoint,opcua.AuthUsername("yc","abc"),opcua.SecurityFromEndpoint())
	if err := c.Connect(); err != nil {
		fmt.Println("err mark 1")
		log.Fatal(err)
	} else {
		fmt.Println("connect success")
	}
	defer c.Close()
```

How to config `opcua.SecurityFromEndpoint()`? it needs two arguments
please

dwhutchison:
Have a look at the crypto example.  While it’s a bit verbose, it contains the steps necessary to connect to your server with SecPolicy None and AuthMode UserName. 

The steps involved would be: 
- load a client certificate (this shouldn’t be necessary with sec mode None; it’s on my radar to allow UserName auth without providing a cert). 
- get a list of the server endpoints from GetEndpoints()
- loop through the endpoints and find the endpoint you wish to use based on your sec policy/mode (None/None)
- provide that Endpoint to SecurityFromEndpoint with the second parameter as ua.UserTokenTypeUserName
- Provide the AuthUserName() option as you are

That should configure the client as you’re expecting. 

Note you can cache the results of GetEndpoints as they rarely change. This could save the request on each connect.

cngobd:
@dwhutchison  The first parameter of  opcua.SecurityFromEndpoint() in opcua.NewClient() is *ua.EndpointDescription,  I configured it like this:
/*
var ep ua.EndpointDescription
	ep.EndpointURL = "opc.tcp://localhost:4840"
	ep.SecurityMode = ua.MessageSecurityModeNone
	ep.SecurityPolicyURI = "http://opcfoundation.org/UA/SecurityPolicy#None"

c := opcua.NewClient(*endpoint, opcua.AuthUsername("yc", "abc123"),
opcua.SecurityFromEndpoint(&ep, ua.UserTokenTypeUserName))
/
is that right? I ran it but it ruturns : The user identity token is valid but the server has rejected it. StatusBadIdentityTokenRejected (0x80210000)

if I opened my Prosys OpcUA server and config the User Authentication Methods ,enable the method of Anonymous, it will automatically connect in anonymous condition without of username and password
please

cngobd:
[...]

PR #204, opened 2019-05-18
allow username auth without providing a client certificate

dwhutchison:
Remove the restriction that all non-None asymmetric security policies
need to have both RSA keys present.  This allows for using the
uapolicy package to encrypt usernames for authorization, but
since decryption is never required, a private key is not necessary.

See #190 for discussion


magiconair:
LGTM

PR #205, opened 2019-05-21
feat: add config option for request timeout and handle the timeout in secure channel

alexbrdn:


uasc/secure_channel.go
```diff
@@ -93,7 +93,7 @@ func NewSecureChannel(endpoint string, c *uacp.Conn, cfg *Config) (*SecureChanne
 		reqhdr: &ua.RequestHeader{
 			AuthenticationToken: ua.NewTwoByteNodeID(0),
 			Timestamp:           time.Now(),
-			TimeoutHint:         0xffff,
+			TimeoutHint:         uint32(cfg.RequestTimeout.Round(time.Millisecond).Nanoseconds() / 1e6),
```
magiconair: I think this should be just
```go
uint32(cfg.RequestTimeout/time.Millisecond)
```

magiconair (commented):
Small improvement. Could you also add an `Option` like this one, please?

https://github.com/gopcua/opcua/blob/master/config.go#L46-L53

uasc/secure_channel.go
alexbrdn: Absolutely. That was stupid of me :)

alexbrdn:
added `Option` and cleaned up as suggested

uasc/secure_channel.go
magiconair: nah :)

magiconair:
LGTM

PR #206, opened 2019-05-22
fix(secure_channel): continue running receive loop on error

alexbrdn:
This fixes two errors:
- The `recv()` loop should only be stopped when the secure channel or the connection is closed. The existing code aborted the loop when an error was returned by the Server. Consequently, any subsequent requests never received their responses and waited until timeout.
- #205 was incomplete. I forgot to delete the handler on request timeout. (I'm not sure about smartness of returning three values from `SendAsync()`, but it seems not worthy of creating a return value struct.)


magiconair:
This looks fine to me. We can always improve it later if necessary.

PR #207, opened 2019-05-22
A Minimal Implementation of Subscription Services

alexbrdn:
Provide initial minimal implementation of Subscription services and necessary MonitoredItems services.
- high-level API functions `Subscribe()`, `Unsubscribe()`, and `PublishLoop()`
- low-level functions `Publish()`, `CreateSubscription()`, `DeleteSubscriptions()`, and `DeleteMonitoredItems()` 
- a working example of the whole subscribe/unsubscribe cycle
- fix a bug that prevented serialization of `nil` `ExtensionObject` field values (didn't allow passing `nil` as `MonitoredItemCreateRequest.RequestedParameters.Filter`)

Comments are appreciated, especially on the high-level API design.


magiconair:
Started looking. will continue tomorrow.

magiconair (commented):
Apart from some minor nitpicks like import ordering and zero values that should be omitted I have these comments/observations:

1. `Subscription` feels big enough to move into its own class like `Node`. Then the `Monitor` and `Delete` operations would happen on the `Subscription` object and not via the client. Is there a reason why operations should happen on the subscription id outside of the `Subscription` object? (similar to attach/detach session).
1. The caller should be able to provide a custom channel instead of just the channel buffer size. Create a default one if none is provided.
1. Cancellation of the run loop should happen via a `context` unless there is a good reason not to. I think this would also help to terminate/unblock all the go routines that are used for notifications and error handling.
1. #210 was a motivated by this PR to keep the error messages consistent and to reduce the `c.Send` boilerplate.

The example generally feels too complex and we should try to hide/reduce more of the boilerplate. But that could also stem from the current cancellation logic. How about this?

```go
	sub, err := c.Subscribe(ua.SubscriptionParameters{
		Notifs : make(chan *ua.PublishNotif, 10), // custom channel
	})
	if err != nil {
		...
	}
	defer sub.Close() // Unsubscribe

	if err := sub.Monitor(items...); err != nil {
		...
	}

	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	go sub.Run(ctx)

	for {
		select {
		case <-ctx.Done():
			return
		case item := <-sub.C: // similar to time.C
			...
		}
	}

	sub.Unmonitor(items...)
```

magiconair:
@alexbrdn I've invited you as a member to this org so that you can push to branches in this repo. This makes collaboration on a PR a bit simpler.

alexbrdn:
Thanks for the invite. I'm going to review/incorporate your comments soon.

magiconair:
@alexbrdn Could you not force-push your changes? This makes it harder to review what you have changed. I think understanding what you have changed between the attempts is interesting during this stage of the change. 

If you want to squash before we merge then that is fine with me as long as the commits don't lump random things together.

magiconair:
Ah, my bad. This was just the rebase to master. Sorry about that.

alexbrdn:
@magiconair, I've moved the Subscription into an own file and changed the API along the lines of what you suggested. Let me know what you think of it.

client.go
```diff
@@ -472,101 +586,141 @@ type PublishNotificationData struct {
 	Value          interface{}
 }
 
-func (c *Client) Publish(notif chan<- PublishNotificationData) {
-	// Empty SubscriptionAcknowledgements for first PublishRequest
-	var acks = make([]*ua.SubscriptionAcknowledgement, 0)
+// Publish() sends a single Publish request with given acknowledgements
+func (c *Client) Publish(acks []*ua.SubscriptionAcknowledgement) (*ua.PublishResponse, error) {
+	req := &ua.PublishRequest{
+		SubscriptionAcknowledgements: acks,
```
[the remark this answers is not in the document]
alexbrdn: fixed

alexbrdn:
Decided to drop `Publish()` from `Client`'s API. (CreateSubscription(), DeleteSubscriptions(), and DeleteMonitoredItems() are already gone.) I think it's better to try and keep Client's API high-level. for such simple low-level calls, anyone can do a `Send()`

examples/subscribe/subscribe.go
```diff
@@ -27,67 +28,47 @@ func main() {
...
-		if result.StatusCode != ua.StatusOK {
-			log.Fatal(result.StatusCode)
-		}
-	}
-	miId := res.Results[0].MonitoredItemID
-	log.Printf("created MonitoredItem with id %v", miId)
 
-	quit := time.After(30 * time.Second)
-LOOP:
+	ctx, _ := context.WithTimeout(context.Background(), 30*time.Second)
```
magiconair: don't omit `cancel`. Just call `defer cancel()` out of habit.

magiconair (commented):
Next round. I'll have another look later.

client.go
```diff
@@ -479,84 +451,37 @@ func (c *Client) Subscribe(params SubscriptionParameters) (*Subscription, error)
...
-
-// Unsubscribe() deletes the given Subscription from server and stops the Publish loop that was
-// started by Subscribe()
-func (c *Client) Unsubscribe(sub *Subscription) error {
-	if registeredSub, ok := c.subscriptions[sub.SubscriptionID]; ok {
-		close(registeredSub.stopPublishLoop)
-		delete(c.subscriptions, sub.SubscriptionID)
+func setDefaultSubscriptionParameters(params *SubscriptionParameters) {
+	if params.MaxNotificationsPerPublish == 0 {
+		params.MaxNotificationsPerPublish = 10000
```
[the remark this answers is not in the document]
alexbrdn: absolutely

client.go
```diff
@@ -479,84 +451,37 @@ func (c *Client) Subscribe(params SubscriptionParameters) (*Subscription, error)
...
-		DataEncoding: &ua.QualifiedName{},
-	}
-	params := ua.MonitoringParameters{
-		ClientHandle:     clientHandle,
-		DiscardOldest:    true,
-		Filter:           nil,
-		QueueSize:        10,
-		SamplingInterval: 0.0,
+	if params.Notifs == nil {
+		params.Notifs = make(chan PublishNotificationData)
```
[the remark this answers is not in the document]
alexbrdn: I don't have a preference either way. What's your reasoning?

client.go
```diff
@@ -634,15 +524,15 @@ func (c *Client) notifySubscription(response *ua.PublishResponse) {
 	}
 
 	if status != ua.StatusOK {
-		sub.Channel <- PublishNotificationData{
+		sub.Notifs <- PublishNotificationData{
```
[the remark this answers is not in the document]
alexbrdn: Good point

examples/subscribe/subscribe.go
```diff
@@ -27,67 +28,47 @@ func main() {
 	}
 	defer c.Close()
 
-	subParams := opcua.NewDefaultSubscriptionParameters()
-	sub, err := c.Subscribe(*subParams)
+	sub, err := c.Subscribe(&opcua.SubscriptionParameters{
+		Interval: 500 * time.Millisecond,
+		Notifs:   make(chan opcua.PublishNotificationData, 37),
```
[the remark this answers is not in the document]
alexbrdn: That's just a silly random number. I think I'll simply take the line out. The example doesn't need to override the default.

examples/subscribe/subscribe.go
alexbrdn: Wilco

client.go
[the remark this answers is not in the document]
magiconair: Generally, I always have to think about side effects (or lack thereof) when using value instead of pointer semantics. For one, (I think) all calls in the API use struct pointers and this one is different. So there should be a reason for this. Otherwise, it just trips up users of the API. 

Two, the struct is going to be copied when passed around but the embedded value is an `interface{}`. So you never create a deep copy. 

I tend to use value semantics for structs when I don't want users to change the values or trip them up if they do, like for a `Config` struct. For the rest I tend to go with pointers unless there is a good reason not to.

If you don't have a preference then I would aim for consistency with the rest of the API, hence pointer.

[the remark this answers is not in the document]
alexbrdn: Points taken. I'll change to `*PublishNotificationData`

alexbrdn:
I appreciate your feedback. My Go Fu grows stronger by the hour!

Next iteration ready.

magiconair:
This looks quite clean. Let me try it.

magiconair:
@alexbrdn do you see this as well? Why doesn't circle-ci catch this?

```
~/src/github.com/gopcua/opcua (alexbrdn-feat/subscription-initial)
$ go test ./...
# github.com/gopcua/opcua/examples/publish
examples/publish/publish.go:28:9: c.PublishLoop undefined (type *opcua.Client has no field or method PublishLoop)
?   	github.com/gopcua/opcua	[no test files]
?   	github.com/gopcua/opcua/cmd/id	[no test files]
?   	github.com/gopcua/opcua/cmd/service	[no test files]
?   	github.com/gopcua/opcua/cmd/service/goname	[no test files]
?   	github.com/gopcua/opcua/cmd/status	[no test files]
?   	github.com/gopcua/opcua/debug	[no test files]
?   	github.com/gopcua/opcua/examples/accesslevel	[no test files]
?   	github.com/gopcua/opcua/examples/browse	[no test files]
?   	github.com/gopcua/opcua/examples/crypto	[no test files]
?   	github.com/gopcua/opcua/examples/datetime	[no test files]
?   	github.com/gopcua/opcua/examples/endpoints	[no test files]
?   	github.com/gopcua/opcua/examples/history-read	[no test files]
ok  	github.com/gopcua/opcua/ua	0.014s
ok  	github.com/gopcua/opcua/uacp	0.041s
ok  	github.com/gopcua/opcua/uapolicy	1.250s
ok  	github.com/gopcua/opcua/uasc	0.028s
ok  	github.com/gopcua/opcua/utils	(cached)
```

magiconair:
something is off

```shell
~/src/github.com/gopcua/opcua (alexbrdn-feat/subscription-initial *)
$ go run examples/subscribe/subscribe.go -endpoint opc.tcp://1.2.3.4:4840 -node 'ns=4;i=2180'
created Subscription with id 3170885363
MonitoredItem with client handle 42 = 23.9
MonitoredItem with client handle 42 = 25
The sequence number is unknown to the server. StatusBadSequenceNumberUnknown (0x807A0000)
```

Used the Prosys client to connect to our Siemens S7 1500 PLC and changed a value manually after subscribing to it. First change worked but then it blew up.

magiconair:
Also, small improvement for the example:

```patch
diff --git a/examples/subscribe/subscribe.go b/examples/subscribe/subscribe.go
index fde2427..d9ddc32 100644
--- a/examples/subscribe/subscribe.go
+++ b/examples/subscribe/subscribe.go
@@ -7,17 +7,17 @@ package main
 import (
 	"context"
 	"flag"
+	"log"
+	"time"
+
 	"github.com/gopcua/opcua"
 	"github.com/gopcua/opcua/debug"
 	"github.com/gopcua/opcua/ua"
-	"log"
-	"time"
 )
 
 func main() {
 	endpoint := flag.String("endpoint", "opc.tcp://localhost:4840", "OPC UA Endpoint URL")
-	namespace := flag.Uint("namespace", 0, "Namespace id of the node to subscribe to")
-	stringId := flag.String("id", "", "String id of the node to subscribe to")
+	nodeID := flag.String("node", "", "NodeID to subscribe to")
 	flag.BoolVar(&debug.Enable, "debug", false, "enable debug logging")
 	flag.Parse()
 	log.SetFlags(0)
@@ -37,7 +37,12 @@ func main() {
 	defer sub.Cancel()
 	log.Printf("created Subscription with id %v", sub.SubscriptionID)
 
-	miCreateRequest := opcua.NewMonitoredItemCreateRequestWithDefaults(ua.NewStringNodeID(uint16(*namespace), *stringId), ua.AttributeIDValue, 42)
+	id, err := ua.ParseNodeID(*nodeID)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	miCreateRequest := opcua.NewMonitoredItemCreateRequestWithDefaults(id, ua.AttributeIDValue, 42)
 	res, err := sub.Monitor(ua.TimestampsToReturnBoth, miCreateRequest)
 	if err != nil || res.Results[0].StatusCode != ua.StatusOK {
 		log.Fatal(err)
```

magiconair:
I think this is related to one of your previous changes:

```shell
$ go run examples/subscribe/subscribe.go -endpoint opc.tcp://1.2.3.4:4840 -node 'ns=4;i=2180' -debug
debug: Connect to opc.tcp://1.2.3.4:4840
debug: conn 1: start HEL/ACK handshake
debug: conn 1: sent HELF with 61 bytes
debug: conn 1: recv ACKF with 28 bytes
debug: conn 1: recv &uacp.Acknowledge{Version:0x0, ReceiveBufSize:0x2000, SendBufSize:0x2000, MaxMessageSize:0x200000, MaxChunkCount:0x100}
debug: conn 1/1: send *ua.OpenSecureChannelRequest with 132 bytes
debug: conn 1: recv OPNF with 136 bytes
debug: conn 1/1: set secure channel id to 3191374479
debug: conn 1/1: recv OPNF with 136 bytes
debug: conn 1/1: res: OK (0x0)
debug: conn 1/1: recv *ua.OpenSecureChannelResponse
debug: conn 1/2: send *ua.CreateSessionRequest with 228 bytes
debug: conn 1: recv MSGC with 8192 bytes
debug: conn 1/2: recv MSGC with 8192 bytes
debug: conn 1: recv MSGF with 3254 bytes
debug: conn 1/2: recv MSGF with 3254 bytes
debug: conn 1/2: res: OK (0x0)
debug: conn 1/2: recv *ua.CreateSessionResponse
debug: conn 1/3: send *ua.ActivateSessionRequest with 117 bytes
debug: conn 1: recv MSGF with 96 bytes
debug: conn 1/3: recv MSGF with 96 bytes
debug: conn 1/3: res: OK (0x0)
debug: conn 1/3: recv *ua.ActivateSessionResponse
debug: conn 1/4: send *ua.CreateSubscriptionRequest with 84 bytes
debug: conn 1: recv MSGF with 72 bytes
debug: conn 1/4: recv MSGF with 72 bytes
debug: conn 1/4: res: OK (0x0)
debug: conn 1/4: recv *ua.CreateSubscriptionResponse
created Subscription with id 3170885364
debug: conn 1/5: send *ua.CreateMonitoredItemsRequest with 116 bytes
debug: conn 1: recv MSGF with 83 bytes
debug: conn 1/5: recv MSGF with 83 bytes
debug: conn 1/5: res: OK (0x0)
debug: conn 1/5: recv *ua.CreateMonitoredItemsResponse
debug: conn 1/6: send *ua.PublishRequest with 66 bytes
debug: conn 1: recv MSGF with 132 bytes
debug: conn 1/6: recv MSGF with 132 bytes
debug: conn 1/6: res: OK (0x0)
debug: conn 1/6: recv *ua.PublishResponse
MonitoredItem with client handle 42 = 3.9874936e+07
debug: conn 1/7: send *ua.PublishRequest with 74 bytes
debug: conn 1/8: send *ua.PublishRequest with 74 bytes
debug: conn 1: recv MSGF with 52 bytes
debug: conn 1/7: recv MSGF with 52 bytes
debug: conn 1/7: res:The operation timed out. StatusBadTimeout (0x800A0000)
debug: conn 1/7: The operation timed out. StatusBadTimeout (0x800A0000)
debug: conn 1/7: no handler for *ua.ServiceFault
debug: conn 1: recv MSGF with 136 bytes
debug: conn 1/8: recv MSGF with 136 bytes
debug: conn 1/8: res: OK (0x0)
debug: conn 1/8: recv *ua.PublishResponse
The sequence number is unknown to the server. StatusBadSequenceNumberUnknown (0x807A0000)
debug: conn 1/9: send *ua.PublishRequest with 74 bytes
debug: conn 1/10: send *ua.PublishRequest with 74 bytes
debug: conn 1/11: send *ua.DeleteSubscriptionsRequest with 70 bytes
debug: conn 1: recv MSGF with 52 bytes
debug: conn 1/9: recv MSGF with 52 bytes
debug: conn 1/9: res:There is no subscription available for this session. StatusBadNoSubscription (0x80790000)
debug: conn 1/9: There is no subscription available for this session. StatusBadNoSubscription (0x80790000)
debug: conn 1/9: no handler for *ua.ServiceFault
debug: conn 1: recv MSGF with 52 bytes
debug: conn 1/10: recv MSGF with 52 bytes
debug: conn 1/10: res:There is no subscription available for this session. StatusBadNoSubscription (0x80790000)
debug: conn 1/10: There is no subscription available for this session. StatusBadNoSubscription (0x80790000)
debug: conn 1: recv MSGF with 64 bytes
debug: conn 1/11: recv MSGF with 64 bytes
debug: conn 1/11: res: OK (0x0)
debug: conn 1/11: recv *ua.DeleteSubscriptionsResponse
debug: conn 1/12: send *ua.CloseSessionRequest with 63 bytes
debug: conn 1: recv MSGF with 52 bytes
debug: conn 1/12: recv MSGF with 52 bytes
debug: conn 1/12: res: OK (0x0)
debug: conn 1/12: recv *ua.CloseSessionResponse
debug: conn 1/13: send *ua.CloseSecureChannelRequest with 57 bytes
debug: conn 1: close
```

alexbrdn:
That's interesting. Apart from the publish example. That one is plain embarrassing. I did adapt the example the first time I changed API, but missed it the second time. Speaking of which, there's no more `Client.Publish()` as it's role is taken by `Subscription.Run()`. Should I simply delete the publish example?

I haven't been able to reproduce the `StatusBadSequenceNumberUnknown` error. It also puzzles me, frankly. AFAIKT, the sequence number can only come from a `PublishResponse`, so the error can only be valid if we process the same response twice.

I will also look at this interesting piece:
```
debug: conn 1/7: send *ua.PublishRequest with 74 bytes
debug: conn 1/8: send *ua.PublishRequest with 74 bytes
```
I want to undertand why it sends two PublishRequests right from the start. There's a chance that `StatusBadSequenceNumberUnknown` is related to this.

I'll let you know when I have some results.

alexbrdn:
I have an explanation for `StatusBadSequenceNumberUnknown`.
When the timeout hits, `acks` don't get reset and are sent again to the server with the next Publish. However, the Server has already received the previous acks and has deleted the sequence numbers from retransmission queue. Hence `StatusBadSequenceNumberUnknown`.

dwhutchison:
Finally got a chance to take a look at this.  It's looking good.
@alexbrdn I was able to reproduce that error and I came to the same conclusion you did.  The error is coming from the fact that we're sending an acknowledge for the same sequence twice.  The reason we're sending it twice is because the `PublishRequest` containing the first acknowledge times out and we try it again.  However, the timeout is erroneous because the server is intentionally holding back on sending a response until it has changed data to report.

I was able to get the error to go away if I modified the RequestHeader's timeoutHint with a value larger than the `RevisedPublishingInterval * RevisedMaxKeepAliveCount`.  When that time expires server-side, the server will send a keepalive response back, even if there's no data.  This means `PublishRequest` packets would have huge timeouts (several minutes or longer) while normal requests have the default 10s timeout.  The following patch adds the ability to set a custom timeout in the stack.  You can then use `c.sendWithTimeout(...)` in `publishOnce` to set a custom timeout for publish requests.  Note the timeout should be one of: `c.cfg.RequestTimeout`, `RevisedPublishingInterval * RevisedMaxKeepAliveCount` (plus an offset to allow the server time to process the keepalive), and `UINT32MAX = ^uint32(0)`.  We should take the larger of the two timeouts while still making sure we're not overflowing the size of the timeout field.

Note this would only prevent the timeout on the first `PublishRequest` in a queue.  If we preemptively sent out a few PublishRequests for the server to hold onto while it's waiting for changed data, the first one will return with a keepalive but the rest will still time out (not saying this is bad, but we'll need to still expect timeouts).  So, we should still probably deal with the ack queue.  Not sure if it's OK to just assume the sequences were ack'd and clear the queue after we send, if we just ignore the `StatusBadSequenceNumberUnknown`, or some more clever 3rd option.

```diff
diff --git client.go client.go
index f2840fb..edceac9 100644
--- client.go
+++ client.go
@@ -347,13 +347,20 @@ func (c *Client) DetachSession() (*Session, error) {
 
 // Send sends the request via the secure channel and registers a handler for
 // the response. If the client has an active session it injects the
-// authenticaton token.
+// authentication token.
 func (c *Client) Send(req interface{}, h func(interface{}) error) error {
+	return c.sendWithTimeout(req, c.cfg.RequestTimeout, h)
+}
+
+// sendWithTimeout sends the request via the secure channel with a custom timeout and registers a handler for
+// the response. If the client has an active session it injects the
+// authenticaton token.
+func (c *Client) sendWithTimeout(req interface{}, timeout time.Duration, h func(interface{}) error) error {
 	var authToken *ua.NodeID
 	if s := c.Session(); s != nil {
 		authToken = s.resp.AuthenticationToken
 	}
-	return c.sechan.Send(req, authToken, h)
+	return c.sechan.SendWithTimeout(req, authToken, timeout, h)
 }
 
 // Node returns a node object which accesses its attributes
diff --git uasc/secure_channel.go uasc/secure_channel.go
index 96bc6be..cd6338c 100644
--- uasc/secure_channel.go
+++ uasc/secure_channel.go
@@ -201,7 +201,12 @@ func (s *SecureChannel) closeSecureChannel() error {
 
 // Send sends the service request and calls h with the response.
 func (s *SecureChannel) Send(svc interface{}, authToken *ua.NodeID, h func(interface{}) error) error {
-	ch, reqid, err := s.SendAsync(svc, authToken)
+	return s.SendWithTimeout(svc, authToken, s.cfg.RequestTimeout, h)
+}
+
+// SendWithTimeout sends the service request and calls h with the response with a specific timeout.
+func (s *SecureChannel) SendWithTimeout(svc interface{}, authToken *ua.NodeID, timeout time.Duration, h func(interface{}) error) error {
+	ch, reqid, err := s.sendAsyncWithTimeout(svc, authToken, timeout)
 	if err != nil {
 		return err
 	}
@@ -216,7 +221,7 @@ func (s *SecureChannel) Send(svc interface{}, authToken *ua.NodeID, h func(inter
 			return resp.Err
 		}
 		return h(resp.V)
-	case <-time.After(s.cfg.RequestTimeout):
+	case <-time.After(timeout):
 		s.mu.Lock()
 		s.popHandlerLock(reqid)
 		s.mu.Unlock()
@@ -227,6 +232,12 @@ func (s *SecureChannel) Send(svc interface{}, authToken *ua.NodeID, h func(inter
 // SendAsync sends the service request and returns a channel which will receive the
 // response when it arrives.
 func (s *SecureChannel) SendAsync(svc interface{}, authToken *ua.NodeID) (resp chan Response, reqID uint32, err error) {
+	return s.sendAsyncWithTimeout(svc, authToken, s.cfg.RequestTimeout)
+}
+
+// sendAsyncWithTimeout sends the service request with a specific timeout and returns a channel which will receive the
+// response when it arrives.
+func (s *SecureChannel) sendAsyncWithTimeout(svc interface{}, authToken *ua.NodeID, timeout time.Duration) (resp chan Response, reqID uint32, err error) {
 	typeID := ua.ServiceTypeID(svc)
 	if typeID == 0 {
 		return nil, 0, fmt.Errorf("unknown service %T. Did you call register?", svc)
@@ -245,6 +256,7 @@ func (s *SecureChannel) SendAsync(svc interface{}, authToken *ua.NodeID) (resp c
 	s.reqhdr.AuthenticationToken = authToken
 	s.reqhdr.RequestHandle++
 	s.reqhdr.Timestamp = time.Now()
+	s.reqhdr.TimeoutHint = uint32(timeout / time.Millisecond)
 
 	// encode the message
 	m := NewMessage(svc, typeID, s.cfg)
```

alexbrdn:
@dwhutchison, I like your approach of providing different timeouts for Publish requests.

Here's what I came up with in the meantime:
```diff
diff --git a/subscription.go b/subscription.go
index 859d66d..3ff8dfc 100644
--- a/subscription.go
+++ b/subscription.go
@@ -127,8 +127,7 @@ func publishOnce(c *Client, acks []*ua.SubscriptionAcknowledgement) (*ua.Publish
 // Run() starts an infinite loop that sends PublishRequests and delivers received
 // notifications to registered Subscriptions.
 func (s *Subscription) Run(ctx context.Context) {
-       // Empty SubscriptionAcknowledgements for first PublishRequest
-       var acks = make([]*ua.SubscriptionAcknowledgement, 0)
+       var acks []*ua.SubscriptionAcknowledgement

        for {
                select {
@@ -137,7 +136,14 @@ func (s *Subscription) Run(ctx context.Context) {
                default:
                        res, err := publishOnce(s.c, acks)
                        if err != nil {
-                               if err == ua.StatusBadTimeout {
+                               if err == ua.StatusBadSequenceNumberUnknown || err == ua.StatusBadTimeout {
+                                       // StatusBadSequenceNumberUnknown means that previous Publish of this loop
+                                       // resulted in some error, but the Server did receive the request and processed
+                                       // acknowledgements. Hence need to clear acks to get unstuck.
+                                       // StatusBadTimeout is a known example of an error that may cause
+                                       // StatusBadSequenceNumberUnknown so it makes sense to preempt that by clearing
+                                       // the acks right after timeout
+                                       acks = make([]*ua.SubscriptionAcknowledgement, 0)
                                        continue
                                } else if err == ua.StatusBadNoSubscription {
                                        // ignore it as probably the cause is that all subscriptions are already deleted,
```
This will prevent the loop getting stuck sending the same no longer valid acks again and again, but will result in some unnecessary back and forth between client and server.

I think if we should probably implement both (your approach and mine.)

dwhutchison:
Hmm, I just noticed that responses with `StatusBadSequenceNumberUnknown` can have valid notification data.  That means we both need to report the data back to the user and also add that sequence to the next ack.

The PublishResponse Status field seems to only be related to the acks in the request and doesn't describe the status of the notification data.   Maybe we should just log `StatusBadSequenceNumberUnknown` and then treat it like err == nil as this will report the data, clear old acks, and set up new ones if required.

As for clearing acks on timeouts, we could leave the functionality as-is (i.e. repeat the ack on the next request).  We'd expect a  `StatusBadSequenceNumberUnknown` if the server had processed the first one, but that's OK since we're dealing with that error better.  If the server hadn't received the first ack due to a legitimate comms loss, then it should be re-acked.

Does that work?  Am I missing something?

alexbrdn:
That's news to me. I guess, then, that all responses we get from server (as opposed to communication errors) may contain notification data, irrespective of the error status code and need to be dealt with accordingly. I would assume in this case, that the server has processed all the acks and we can clear them whenever we get a response we can parse. Won't need special handling for `StatusBadSequenceNumberUnknown` then.

jkissinger:
@alexbrdn @dwhutchison any idea when this will be merged?  It looks like it's ready, but I may be missing something.  Thanks.

magiconair:
Let me try this again.

subscription.go
```diff
@@ -0,0 +1,219 @@
...
+	DefaultSubscriptionMaxNotificationsPerPublish = 10000
+	DefaultSubscriptionLifetimeCount              = 10000
+	DefaultSubscriptionMaxKeepAliveCount          = 3000
+	DefaultSubscriptionInterval                   = 100 * time.Millisecond
+	DefaultSubscriptionPriority                   = 0
+)
+
+type Subscription struct {
+	SubscriptionID            uint32
+	RevisedPublishingInterval float64
```
magiconair: Can we make this a `time.Duration` just like `Interval` below?

subscription.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+}
+
+type PublishNotificationData struct {
+	SubscriptionID uint32
+	Error          error
+	Value          interface{}
+}
+
+// Cancel() deletes the Subscription from Server and makes the Client forget it so that publishing
```
magiconair, line 64: Can we rename this to `Close()` to be more in line with the stdlib?

subscription.go
```diff
@@ -0,0 +1,219 @@
...
+// loops cannot deliver notifications to it anymore
+func (s *Subscription) Cancel() error {
+	delete(s.c.subscriptions, s.SubscriptionID)
```
magiconair: This does not look safe for concurrent use and the `Subscription` object should not modify the internal state of the `Client` directly. Use a method.

subscription.go
```diff
@@ -0,0 +1,219 @@
...
+		SubscriptionID:   s.SubscriptionID,
+	}
+	var res *ua.DeleteMonitoredItemsResponse
+	err := s.c.Send(req, func(v interface{}) error {
+		return safeAssign(v, &res)
+	})
+	return res, err
+}
+
+func publishOnceForSubscription(s *Subscription, acks []*ua.SubscriptionAcknowledgement) (*ua.PublishResponse, error) {
```
magiconair: Can this be a method of `Subscription`, e.g. 

```go
func (s *Subscription) publish(acks []*ua.SubscriptionAcknowledgement) (*ua.PublishResponse, error) {...}
```

subscription.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	var res *ua.PublishResponse
+	err := s.c.sendWithTimeout(req, s.getPublishTimeout(), func(v interface{}) error {
+		return safeAssign(v, &res)
+	})
+	return res, err
+}
+
+func (s *Subscription) getPublishTimeout() time.Duration {
```
magiconair: Can we rename this to just `publishTimeout()`?

subscription.go
```diff
@@ -0,0 +1,219 @@
...
+func (s *Subscription) Run(ctx context.Context) {
+	var acks []*ua.SubscriptionAcknowledgement
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		default:
+			res, err := publishOnceForSubscription(s, acks)
+			if err != nil {
```
magiconair: I've tried to refactor this with different approaches and this is the best I could come up with.

```go
// send the next publish request
// note that res contains data even if an error was returned
res, err := s.publish(acks)

switch {
case err == ua.StatusBadSequenceNumberUnknown:
	// At least one ack has been submitted repeatedly.
	// Ignore the error. Acks will be cleared below
case err == ua.StatusBadTimeout:
	// ignore and continue the loop
case err == ua.StatusBadNoSubscription:
	// Subscription already deleted and the publishing loop is still running
	// but will be stopped shortly.
	// todo(fs): how?

case err != nil:
	s.c.notifySubscriptions(ctx, err)
}

if res != nil {
...
```

subscription.go
```diff
@@ -0,0 +1,219 @@
...
+				// StatusBadSequenceNumberUnknown means at least one ack has been submitted repeatedly
+				// Ignore the error. Acks will be cleared below
+				if err == ua.StatusBadSequenceNumberUnknown ||
+					// Ignore StatusBadTimeout. No need to do anything except continue the loop
+					err == ua.StatusBadTimeout ||
+					// Ignore StatusBadNoSubscription as probably the cause is that all subscriptions
+					// are already deleted, but the publishing loop is still running and will be stopped shortly
+					err == ua.StatusBadNoSubscription {
+				} else {
+					s.c.notifySubscriptions(ctx, err)
```
magiconair: Why is this subscription notifying all other subscriptions?

subscription.go
```diff
@@ -0,0 +1,219 @@
...
+							SubscriptionID: res.SubscriptionID,
+							SequenceNumber: i,
+						}
+						acks = append(acks, ack)
+					}
+				}
+			}
+
+			if err == nil {
+				s.c.notifySubscription(ctx, res)
```
magiconair, line 184: Why is `notifySubscription` not a method of `Subscription`? 

Either we only notify our subscription, then it should be a method. Or we notify other subscriptions in which case there should be only one run loop in the client.

Or am I missing something?

client.go
```diff
@@ -54,13 +55,16 @@ type Client struct {
 
 	// once initializes session
 	once sync.Once
+
+	// map of active subscriptions managed by this client
+	subscriptions map[uint32]Subscription
```
magiconair: I would guess that you need to guard this with a mutex and prevent direct access to it. This should also be a `map[uint32]*Subscription`. Same reasoning as before.

For formatting, I would also sort it above the `once`.

client.go
```diff
@@ -418,140 +430,120 @@ func (c *Client) Browse(req *ua.BrowseRequest) (*ua.BrowseResponse, error) {
...
 	})
-	return &Subscription{res}, err
+	if err != nil {
+		return nil, err
+	}
+	if res.ResponseHeader.ServiceResult != ua.StatusOK {
+		return nil, res.ResponseHeader.ServiceResult
+	}
+
+	sub := Subscription{
```
magiconair: Create a pointer here.

client.go
```diff
@@ -418,140 +430,120 @@ func (c *Client) Browse(req *ua.BrowseRequest) (*ua.BrowseResponse, error) {
...
+		go func(s Subscription) {
+			s.sendNotification(ctx, &PublishNotificationData{Error: err})
+		}(sub)
+	}
 }
 
-func (c *Client) Publish(notif chan<- PublishNotificationData) {
-	// Empty SubscriptionAcknowledgements for first PublishRequest
-	var acks = make([]*ua.SubscriptionAcknowledgement, 0)
+func (c *Client) notifySubscription(ctx context.Context, response *ua.PublishResponse) {
```
magiconair, line 500: This method is fishy since it is in the `Client` but is only called from within the `Subscription` and then only calls subscription methods. Either this should be a method of `Subscription` or there should only be one run loop. See other comment on the same topic.

examples/subscribe/subscribe.go
```diff
@@ -25,9 +28,52 @@ func main() {
...
+		log.Fatal(err)
+	}
+
+	miCreateRequest := opcua.NewMonitoredItemCreateRequestWithDefaults(id, ua.AttributeIDValue, 42)
+	res, err := sub.Monitor(ua.TimestampsToReturnBoth, miCreateRequest)
+	if err != nil || res.Results[0].StatusCode != ua.StatusOK {
+		log.Fatal(err)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
magiconair:
```go
// add an arbitrary timeout to demonstrate how to stop a subscription
// with a context.
d := 30 * time.Second
ctx, cancel := context.WithTimeout(context.Background(), d)
...
	for {
 		select {
 		case <-ctx.Done():
 			log.Print("Subscription stopped after %s", d)
 			return
```

examples/subscribe/subscribe.go
```diff
@@ -25,9 +28,52 @@ func main() {
...
-	sub, err := c.Subscribe(time.Second)
+	sub, err := c.Subscribe(&opcua.SubscriptionParameters{
+		Interval: 500 * time.Millisecond,
+	})
 	if err != nil {
 		log.Fatal(err)
 	}
-	log.Printf("sub: %v", sub)
+	defer sub.Cancel()
+	log.Printf("created Subscription with id %v", sub.SubscriptionID)
```
magiconair: `Created subscription ...`

magiconair (changes requested):
This seems to work better but there are a couple of things which need clarification before we can merge this.

subscription.go
dwhutchison: There needs to be another exit condition on this loop as well.  On a critical error (I encountered a SessionIDInvalid), the loop never exits because it's only waiting for a context close.  I'm thinking `s.c.notifySubscriptions(ctx, err)` should be followed by a `return`

uasc/secure_channel.go
```diff
@@ -213,10 +219,13 @@ func (s *SecureChannel) Send(svc interface{}, authToken *ua.NodeID, h func(inter
...
 	case resp := <-ch:
 		if resp.Err != nil {
+			if resp.V != nil {
+				_ = h(resp.V) // ignore result because resp.Err takes precedence
+			}
 			return resp.Err
 		}
 		return h(resp.V)
-	case <-time.After(s.cfg.RequestTimeout):
+	case <-time.After(timeout + timeoutLeniency): // `+ timeoutLeniency` to give the server a chance to respond to TimeoutHint
```
dwhutchison, line 230: Correct me if I'm wrong but the server's response to a timeout should be to drop the request so we'd never see a response.
If you're trying to allow for a buffer between the PublishRequest timeout and the keepalive interval, would it be more beneficial to calculate the keepalive interval and add `timeoutLeniency` to the `timeoutHint` sent to the server?  
eg. if the keepalive interval is at 60000ms, the timeoutHint is 60250ms? (Or better yet, have the leniency be a function of the Subscription Interval to guarantee full overlap?)

uasc/secure_channel.go
```diff
@@ -245,6 +260,7 @@ func (s *SecureChannel) SendAsync(svc interface{}, authToken *ua.NodeID) (resp c
 	s.reqhdr.AuthenticationToken = authToken
 	s.reqhdr.RequestHandle++
 	s.reqhdr.Timestamp = time.Now()
+	s.reqhdr.TimeoutHint = uint32(timeout / time.Millisecond)
```
dwhutchison, line 268: We should probably check for a minimum timeout here to make sure the user didn't put in something too small.  Maybe make minimum `cfg.RequestTimeout`?

subscription.go
```diff
@@ -0,0 +1,219 @@
...
+	"time"
+
+	"github.com/gopcua/opcua/ua"
+)
+
+const (
+	DefaultSubscriptionMaxNotificationsPerPublish = 10000
+	DefaultSubscriptionLifetimeCount              = 10000
+	DefaultSubscriptionMaxKeepAliveCount          = 3000
+	DefaultSubscriptionInterval                   = 100 * time.Millisecond
```
dwhutchison, line 15: These defaults seem OK, but this turns into a 5-minute keepalive interval.  Our default SessionTimeout is 1 minute so our session ends up timing out before the server even gets a chance to send the keepalive.  I recommend increasing the default SessiontTimeout.  There's no firm recommendation in the spec or test cases that I could find, but 20 minutes seems to be the most common with other clients (shortest I saw was 5 minutes, longest was 1 hour).

subscription.go
alexbrdn: Yes, we can. `float64` is the type defined for response by the spec. But I see no problem converting it  for ease of use

subscription.go
alexbrdn: I have no strong opinion on this, but it really seems to me that the natural thing to do to stop a subscription is to "cancel" it. Just like you don't "repeat" a subscription, but rather "renew" it.

subscription.go
alexbrdn: good point

subscription.go
alexbrdn: yes, it can

subscription.go
alexbrdn: yep

client.go
alexbrdn: yep. will do

uasc/secure_channel.go
alexbrdn: I have seen the server send a response with `StatusBadTimeout` after the `TimeoutHint` has elapsed, so I do think it makes sense to wait a little longer than `TimeoutHint`.
From Part 4, Table 170:
>For a Server this timeout is only a hint and can be used to cancel long running operations to free resources. If the Server detects a timeout, he can cancel the operation by sending the Service result Bad_Timeout. The Server should wait at minimum the timeout after he received the request before cancelling the operation. The Server shall check the timeoutHint parameter of a Publish request before processing a Publish response. If the request timed out, a Bad_Timeout Service result is sent and another Publish request is used.

subscription.go
alexbrdn: OK. Changed to 20 minutes.

uasc/secure_channel.go
alexbrdn: OK. Sounds reasonable

alexbrdn:
@jkissinger, we're getting there. I hope we'll manage to do it this week, 'cause I'll be away for the next two

client.go
alexbrdn: It's in the `Client`, because only the `Client` is supposed to know the subscription to which the notifications in the response belong to. Remember that a `Publish` sent from one subscription's loop can return with notifications for another subscription.

No, there shouldn't be only one loop. We should maintain at least one loop per subscription so that the server always has a Publish request to send notifications for each subscription. In the future, we should extend the API to allow for queuing of Publish requests to try and maintain a certain number of Publish requests for any given subscription (See Part 4, Chapter 5.13.5.1)

subscription.go
alexbrdn: For some mysterious reason, there's no reply box under [this comment](https://github.com/gopcua/opcua/pull/207#discussion_r290119821), so I'll reply here.

Do you think we should assume that any error we didn't explicitly ignore is irrecoverable? I tend to agree, but I'm not very sure.

subscription.go
alexbrdn: Do you think we should assume that any error we didn't explicitly ignore is irrecoverable? I tend to agree, but I'm not very sure.

subscription.go
alexbrdn: Because Publish requests are not tied to a particular subscription and you can't know, in case of an error, which subscription(s) need to know about the error. At least if you don't have a SubscriptionID in the response. Which gives me an idea that we could check for SubscriptionID and only notify all subscriptions if we really can't pinpoint one.

subscription.go
alexbrdn: I think I've covered this [here](https://github.com/gopcua/opcua/pull/207#discussion_r290176234)

subscription.go
alexbrdn: but then it can just as well be a method of `Client`, e.g.
```go
func (c *Client) publish(acks []*ua.SubscriptionAcknowledgement, timeout time.Duration) (*ua.PublishResponse, error) {...}
```

client.go
alexbrdn: Although, we don't start the loops. The user does. Nevertheless, he should be able to start multiple loops

alexbrdn:
@magiconair, @dwhutchison, I think I've implemented and/or covered all of the points from reviews, except renaming `Subscription.Cancel()` to `Subscription.Close()`. I've marked as resolved all points where I see no need for further discussion. Let me know what you think.

subscription.go
alexbrdn: How about introducing `Close()` as a synonym to `Cancel()`? E.g.
```go
func (s *Subscription) Close() error {
    return s.Cancel()
}
```

magiconair:
Thank you. I’ll have a look tonight. Busy day

subscription.go
magiconair: Either Close or Cancel. I’ll do some digging but this isn’t that important.

magiconair:
I can't test this right now b/c of #214. Need to ask colleagues to re-enable unencrypted access again.

magiconair (approved):

magiconair:
In order to test this on our PLC I need to merge #216 first. I'll try to do that tomorrow and then update the subscription example.

kung-foo:
I tested this against the Prosys test server and have some feedback about the interface (though before I jump in let me say just thanks the PR! working code is awesome)

I first wanted to simplify the interface even more. I took inspiration from the NATS pub/sub api which I have some experience with (https://github.com/nats-io/nats.go#basic-usage).

goals:
1) make the channel that I (as the client) interact with as simple possible. This means only one update per message and also only return information I care about.
2) hide all the things I don't care about
3) add support for mapping the `NodeID` to a `clientHandle` automatically (therefore _returning_ the `NodeID` in the data change message).
4) handle slow consumers
5) minimize or eliminate (hide) "big structs of properties" 

The inline code below uses this struct (defined in the gist) which wraps the existing PR's code:
monitor.go: https://gist.github.com/kung-foo/b184a432de2e3063f891f96d42b25795

```golang
msgCh := make(chan *DataChangeMessage, 16) // TODO(jca): sz?
ctx, cancel := context.WithCancel(context.TODO())
defer cancel()

monitor, err := NewNodeMonitor(client, msgCh)
if err != nil {
	log.Fatal(err)
}

defer monitor.Unsubscribe()

// this gets called asynchronously when something goes wrong that can't be propagated through the channel itself
monitor.SetErrorHandler(func(_ *opcua.Client, _ *opcua.Subscription, err error) {
	log.Printf(err)
	cancel()
}

monitor.Add(
	"ns=5;s=Sinusoid1",
	"ns=5;s=Counter1",
)

monitor.AddNodes(ua.NewStringNodeID(5, "Square1"))

go monitor.Subscribe(ctx)

for {
	select {
	case msg := <-msgCh:
		log.Printf("buffer=%d node=%s value=%v", len(msgCh), msg.NodeID, msg.Value.Value)

		// just for fun remove one after only one message
		if msg.NodeID.String() == "ns=5;s=Sinusoid1" {
			monitor.RemoveNodes(msg.NodeID)
		}
	}
}
```

Other thoughts:
* There's some simplification that could happen by merging this code with @alexbrdn's implementation. Unnecessary goroutines, etc
* I still don't like having to explicitly call `go monitor.Subscribe(ctx)`
* There seem to be enough options on a subscription that it warrants the `options ...Option ` style approach
* I embedded the `*ua.DataValue` directly in the `DataChangeMessage` because my fingers would not let me type `msg.Value.Value.Value` one more time. I kinda want to change the `Variant.Value` field to something else like `.Raw` 
* None of this has to be part of the core `opcua` library. I can see either having a separate high-level API package that is not inspired by the spec, or in another project altogether.
* I'm still a little annoyed at having a `context.cancel` _and_ a `monitor.Unsubscribe()`. Should they be same or maybe trigger each other?

magiconair:
@kung-foo All good feedback. Lets get this over the line first and then take the next step. We've been sitting on this for too long. I managed to get this working today as well.

magiconair:
@kung-foo Please open another issue with your suggestions and/or a PR. I know the `Value.Value.Value` sucks but maybe we can solve that separately. :)

PR #210, opened 2019-05-26
implement a type-safe assign

magiconair:
This patch implements a generic type-safe assign helper to reduce the type checking boilerplate code for `c.Send()` handlers. Instead of 

```go
var res *X
err := c.Send(req, func(v interface{}) error {
    if r, ok := v.(*X); !ok { return fmt.Errorf(...) }
    res = r
    return nil
})
return res, err
```

we can now write

```go
var res *X
err := c.Send(req, func(v interface{}) error {
    return safeAssign(v, &res)
})
return res, err
```

This implementation uses reflection to achieve this but I'm open for other suggestions.

### DIFF
--- a/client.go
+++ b/client.go
@@ -9,6 +9,7 @@ import (
 	"crypto/rand"
 	"fmt"
 	"log"
+	"reflect"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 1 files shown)


alexbrdn:
I guess this is the case where you wish Go had generics :)

magiconair:
Not really. This is actually a good example how you can solve this with a couple of lines of code. I'm sure there are other use-cases where generics are way more desirable. In the 7-8 years I've been using Go I've neither missed generics or a debugger. :)

alexbrdn:
I see what you're saying. Maybe it's just me reflexively flinching at use of reflection.

Issue #214, opened 2019-06-04
Basic128Rsa15+Sign+Anonymous requires private key


magiconair:
Our Siemens SIMATIC S7-1500 got reconfigured and anonymous access disabled. Minimum requirement is `Basic128Rsa15+Sign`. Running the `datetime` example unmodified results in a panic:

```sh
$ go run examples/datetime/datetime.go  -endpoint opc.tcp://192.168.118.30:4840
failed to send close secure channel request
panic: close of closed channel

goroutine 34 [running]:
github.com/gopcua/opcua/uasc.(*SecureChannel).Close(0xc0000f2050, 0x1226940, 0xc0000a0140)
	/Users/frank/src/github.com/gopcua/opcua/uasc/secure_channel.go:114 +0x49
github.com/gopcua/opcua/uasc.(*SecureChannel).recv(0xc0000f2050)
	/Users/frank/src/github.com/gopcua/opcua/uasc/secure_channel.go:351 +0xfe0
created by github.com/gopcua/opcua/uasc.(*SecureChannel).Open
	/Users/frank/src/github.com/gopcua/opcua/uasc/secure_channel.go:106 +0x3f
exit status 2
```

After updating the example with

```patch
$ git diff
diff --git a/examples/datetime/datetime.go b/examples/datetime/datetime.go
index 97f90ec..de5fa37 100644
--- a/examples/datetime/datetime.go
+++ b/examples/datetime/datetime.go
@@ -19,7 +19,12 @@ func main() {
        flag.Parse()
        log.SetFlags(0)

-       c := opcua.NewClient(*endpoint)
+       opts := []opcua.Option{
+               opcua.SecurityPolicy(ua.SecurityPolicyURIBasic128Rsa15),
+               opcua.SecurityMode(ua.MessageSecurityModeSign),
+               opcua.AuthAnonymous(),
+       }
+       c := opcua.NewClient(*endpoint, opts...)
        if err := c.Connect(); err != nil {
                log.Fatal(err)
        }
```

I have to provide a private key

```sh
$ go run examples/datetime/datetime.go  -endpoint opc.tcp://192.168.118.30:4840
invalid channel config: Security policy 'http://opcfoundation.org/UA/SecurityPolicy#Basic128Rsa15' requires a private key
exit status 1
```

This is not required when using the Prosys client with the same server.

@dwhutchison any idea?

dwhutchison:
[...]

PR #216, opened 2019-06-05
Add configuration helpers for security settings

magiconair:
Add config helpers to 

* set security policy from a string
* set security mode from a string
* load private key from a file
* load cert and private key from a pair of files

### DIFF
--- a/config.go
+++ b/config.go
@@ -222,7 +222,7 @@ func loadCertificate(filename string) ([]byte, error) {
 		return nil, fmt.Errorf("Failed to load certificate: %s", err)
 	}
 
-	if strings.HasSuffix(filename, ".pem") {
+	if !strings.HasSuffix(filename, ".pem") {
 		return b, nil
 	}
 
--- a/config.go
+++ b/config.go
@@ -121,13 +121,7 @@ func SecurityModeString(s string) Option {
 // SecurityPolicy sets the security policy uri for the secure channel.
 func SecurityPolicy(s string) Option {
 	return func(c *uasc.Config, sc *uasc.SessionConfig) {
-		switch s {
-		case "None", "Basic128Rsa15", "Basic256", "Basic256Sha256",
-			"Aes128_Sha256_RsaOaep", "Aes256_Sha256_RsaPss":
-			c.SecurityPolicyURI = ua.SecurityPolicyURIPrefix + s
-		default:
-			c.SecurityPolicyURI = s
-		}
+		c.SecurityPolicyURI = ua.FormatSecurityPolicyURI(s)
 	}
 }
 
--- a/config.go
+++ b/config.go
@@ -311,6 +311,9 @@ func setPolicyID(t interface{}, policy string) {
 // AuthPolicyID sets the policy ID of the user identity token
 // Note: This should only be called if you know the exact policy ID the server is expecting.
 // Most callers should use SecurityFromEndpoint as it automatically finds the policyID
+// todo(fs): Should we make 'policy' an option to the other
+// todo(fs): AuthXXX methods since this approach requires context
+// todo(fs): and ordering?
 func AuthPolicyID(policy string) Option {
 	return func(c *uasc.Config, sc *uasc.SessionConfig) {
 		if sc.UserIdentityToken == nil {
@@ -332,6 +335,7 @@ func AuthAnonymous() Option {
 
 		_, ok := sc.UserIdentityToken.(*ua.AnonymousIdentityToken)
 		if !ok {
+			// todo(fs): should we Fatal here?
 			log.Printf("non-anonymous authentication already configured, ignoring")
 			return
 		}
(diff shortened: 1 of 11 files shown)


magiconair:
The goal is to make adding the security flags simpler. Still need to update the rest of the examples.

For now, checking if this is the right direction. 

Motivation is from #214

magiconair:
@dwhutchison I've added a `X509KeyPair` function to load cert/key.pem. Could you double check whether I got this right?

config.go
```diff
@@ -116,20 +143,78 @@ func PrivateKey(key *rsa.PrivateKey) Option {
...
 
+// PrivateKeyFile sets the RSA private key in the secure channel configuration
+// from a PEM or DER encoded file.
+func PrivateKeyFile(keyFile string) Option {
+	return func(c *uasc.Config, sc *uasc.SessionConfig) {
+		if keyFile == "" {
+			return
+		}
+
+		b, err := ioutil.ReadFile(keyFile)
```
magiconair: I should move this to a function and write a test for that.

magiconair: Started to write tests for all config options. Will add one for this as well.

magiconair:
tests are not complete yet

config.go
```diff
@@ -116,20 +154,78 @@ func PrivateKey(key *rsa.PrivateKey) Option {
 	}
 }
 
+// PrivateKeyFile sets the RSA private key in the secure channel configuration
+// from a PEM or DER encoded file.
+func PrivateKeyFile(keyFile string) Option {
```
dwhutchison: I like the idea of supporting multiple filetypes and this is probably the best way to handle these helper functions.  That means, however, we should have one for `Certficate`, too.

Users need to provide both a cert and a key in every case so the two config options should feel the same.

config.go
```diff
@@ -116,20 +154,78 @@ func PrivateKey(key *rsa.PrivateKey) Option {
...
+	if err == nil && len(x509cert.URIs) > 0 {
+		appURI = x509cert.URIs[0].String()
+	}
+	sc.ClientDescription.ApplicationURI = appURI
+}
+
+// X509KeyPair sets the client X509 certificate and private key in the secure
+// channel configuration from a pair of files and also detects and sets the
+// ApplicationURI from the URI within the certificate.
+func X509KeyPair(certFile, keyFile string) Option {
```
dwhutchison: If we add a `CertificateFile` (or similar) method, this now feels redundant and too constrained (these files needs to be PEM encoded, whereas the other methods would accept DER as well).  
I think it would be beneficial to have a generic `CertificateFile` method that accepts multiple file types, and if we wanted a convenience function that accepts both the cert and the key file simultaneously, it should just be a wrapper for the other two.  Both files are required so it makes sense to have a single method to load them.

magiconair: makes sense. Will do.

magiconair: Done.

config.go
```diff
@@ -116,21 +154,111 @@ func PrivateKey(key *rsa.PrivateKey) Option {
...
+		pk, err := x509.ParsePKCS1PrivateKey(derBytes)
+		if err != nil {
+			log.Fatalf("Failed to parse private key: %s", err)
+		}
+
+		c.LocalKey = pk
+	}
+}
+
+func loadPrivateKey(filename string) (*rsa.PrivateKey, error) {
```
golangcibot, line 176: `loadPrivateKey` is unused (from `deadcode`)

config.go
magiconair: Makes way more sense this way.

config.go
magiconair: Duh.

magiconair:
Do we need to support ECDSA or ECC keys?

dwhutchison:
> Do we need to support ECDSA or ECC keys?

I don't think so unless it's required as part of PubSub / MQTT / etc., which I haven't really dug into.  Regardless, it shouldn't be required for us now.

config.go
magiconair: done.

magiconair:
When writing tests for the `AuthXXX` options I wasn't sure what to do with the `AuthPolicyID` since it requires context and ordering. My table driven test does not work for this case and I take this as a signal on whether `AuthPolicyID` should be refactored. Maybe we should make `policy` an arg to the other `AuthXXX` methods since they are special already.

I'm also wondering how we should handle errors since the options cannot return errors but options like `RemoteCertificateFile` can fail. For that option, I'm calling `log.Fatal` but the `AuthXXX` options only `log.Print` on conflict. 

Either this is a conflict which we can ignore, or we should `Fatal` as well or we handle errors in a different way. We could, for example, set an error in the `Config` object and return it on `Dial()`.

Anybody have an opinion on this? 

@dwhutchison I'm not sure if the `AuthXXX` failures are fatal or not.

Note: I'm probably going to merge this as-is and leave the comments in there since I want to merge #207 and fix this in a subsequent PR. Depends on how much time I have today to test subscriptions.

PR #227, opened 2019-06-27
fix(secure_channel): fix data race in sendAsyncWithTimeout()

### LINKED ISSUE #225: Data Race in SecureChannel.sendAsyncWithTimeout
Started a Publish loop for a Subscription and let the main goroutine exit therefore triggering deferred `Client.Close()`. Got basically the same data races as in #184.
Turns out this problem is obviously introduced by me (#196) when I excluded `m.Encode()` from the synchronized region. In this case, `Client.Close()` writes the `s.reqhdr` fields while a Publish request being sent reads from them while encoding.

On one hand just moving `s.mu.Unlock()` after the line with `m.Encode()` fixes this problem. On the other hand, I'm not sure this is a robust solution. It more or less forces to serialize requests being sent through SecureChannel. I imagine storing a prototype of a RequestHeader in SecureChannel and making a copy of it for each request (for safe encoding) would be more robust. Seems there's no need to perform a deep copy, so I don't see problems with that.

alexbrdn:
closes #225

### DIFF
--- a/uasc/secure_channel.go
+++ b/uasc/secure_channel.go
@@ -47,8 +47,8 @@ type SecureChannel struct {
 	// cfg is the configuration for the secure channel.
 	cfg *Config
 
-	// reqhdr is the header for the next request.
-	reqhdr *ua.RequestHeader
+	// reqhdrproto is the prototype for the header for the next request.
+	reqhdrproto *ua.RequestHeader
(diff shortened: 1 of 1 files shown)


alexbrdn:
superseded by #234

PR #230, opened 2019-06-28
fix: make Read example actually print the read value

alexbrdn:
tiny little fix in the Read example to actually see the value that has been read


magiconair (approved):
LGTM

PR #235, opened 2019-07-07
improved cancellation of monitorChannel() goroutine

dwhutchison:
Force cancellation of the monitorChannel() goroutine when the session is closed.  monitorChannel() currently relies on the Receive returning an error to exit but if if that doesn't occur for any reason, monitorChannel() may remain executing.

### DIFF
--- a/client.go
+++ b/client.go
@@ -102,6 +102,9 @@ type Client struct {
 	subscriptions map[uint32]*Subscription
 	subMux        sync.RWMutex
 
+	//cancelMonitor cancels the monitorChannel goroutine
+	cancelMonitor context.CancelFunc
+
 	// once initializes session
 	once sync.Once
 }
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -102,6 +102,9 @@ type Client struct {
 	subscriptions map[uint32]*Subscription
 	subMux        sync.RWMutex
 
+	//cancelMonitor cancels the monitorChannel goroutine
+	cancelMonitor context.CancelFunc
```
kung-foo, line 106: it is normal to store a context related item in a struct?

dwhutchison: I hesitated with this as well but the alternative would likely be a done channel.  Using a done channel, the select statement grows by one case statement and has the same body as the ctx.Done() case.  This felt redundant so using the context seemed more natural as we were already using it. 

That said, I'm not opposed to changing it if there are strong arguments against it.

magiconair: I think this is OK. If the context would need to cancel more than just the monitor I'd suggest to make it more generic but this is good enough.

magiconair:
I'll merge and tag a new version.

PR #237, opened 2019-07-11
add method:translate relative path of browse name to node id

### LINKED ISSUE #236: How to access a node by StringNodeId ?
Is there any way to access a node by path of DisplayName?

for example:

ua.NewStringNodeId(2,"demo_led.temperature")

image

sdghchj:
Closes #236

### DIFF
--- a/examples/translate/translate.go
+++ b/examples/translate/translate.go
@@ -19,6 +19,8 @@ import (
 func main() {
 	endpoint := flag.String("endpoint", "opc.tcp://localhost:4840", "OPC UA Endpoint URL")
 	flag.BoolVar(&debug.Enable, "debug", false, "enable debug logging")
+	ns := flag.Int("namespace", 0, "namespace of node")
+	nodePath := flag.String("path", "", "path of a node's browse name")
 	flag.Parse()
 	log.SetFlags(0)
 
@@ -31,7 +33,7 @@ func main() {
 	defer c.Close()
 
 	root := c.Node(ua.NewTwoByteNodeID(id.ObjectsFolder))
-	nodeId, err := root.TranslateBrowsePathInSameNamespaceToNodeId(2, "demo_led.temperature")
+	nodeId, err := root.TranslateBrowsePathInSameNamespaceToNodeId(uint8(*ns), *nodePath)
 	if err != nil {
 		log.Fatal(err)
 		return
--- a/examples/translate/translate.go
+++ b/examples/translate/translate.go
@@ -18,11 +18,10 @@ import (
 
 func main() {
 	endpoint := flag.String("endpoint", "opc.tcp://localhost:4840", "OPC UA Endpoint URL")
-	flag.BoolVar(&debug.Enable, "debug", false, "enable debug logging")
 	ns := flag.Int("namespace", 0, "namespace of node")
+	nodePath := flag.String("path", "device_led.temperature", "path of a node's browse name")
+	flag.BoolVar(&debug.Enable, "debug", false, "enable debug logging")
 
-	// example: "device_led.temperature"
-	nodePath := flag.String("path", "", "path of a node's browse name")
 	flag.Parse()
 	log.SetFlags(0)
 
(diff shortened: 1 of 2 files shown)


examples/translate/translate.go
```diff
@@ -0,0 +1,40 @@
...
+	ctx := context.Background()
+
+	c := opcua.NewClient(*endpoint)
+	if err := c.Connect(ctx); err != nil {
+		log.Fatal(err)
+	}
+	defer c.Close()
+
+	root := c.Node(ua.NewTwoByteNodeID(id.ObjectsFolder))
+	nodeId, err := root.TranslateBrowsePathInSameNamespaceToNodeId(2, "demo_led.temperature")
```
alexbrdn: I'd say, for the example to be portable, the arguments of `TranslateBrowsePathInSameNamespaceToNodeId()` should be provided as command line flags, just like the endpoint. Maybe add a comment with your values as an example.

node.go
```diff
@@ -141,3 +143,49 @@ func (n *Node) References(refs *ua.NodeID) (*ua.BrowseResponse, error) {
...
+					if len(resp.Results[0].Targets) > 0 {
+						nodeId = resp.Results[0].Targets[0].TargetID.NodeID
+						return nil
+					}
+				} else {
+					return errors.New(resp.Results[0].StatusCode.Error())
+				}
+			}
+		}
+		return errors.New("unknown error")
```
alexbrdn: I recall having [this discussion](https://github.com/gopcua/opcua/pull/188#discussion_r282367121) about error codes.
> If we don't use status codes then the error message becomes part of the public API and users will have to check for that. Lets stick to status codes unless there is a good reason not to.

Maybe you should stick to status codes  here as well.
The same applies to line 173.

sdghchj:
@alexbrdn check again

alexbrdn:
Code looks OK to me.
Test OK.

alexbrdn:
I think it's worth mentioning that this closes #236

examples/translate/translate.go
```diff
@@ -0,0 +1,44 @@
...
+	"log"
+
+	"github.com/gopcua/opcua"
+	"github.com/gopcua/opcua/debug"
+	"github.com/gopcua/opcua/id"
+	"github.com/gopcua/opcua/ua"
+)
+
+func main() {
+	endpoint := flag.String("endpoint", "opc.tcp://localhost:4840", "OPC UA Endpoint URL")
```
magiconair, line 20: Can you order the flags like in the other examples, please?

```go
endpoint := ...
ns := ...
nodePath := ...
flag.BoolVar(...)
```

Also please add the example `device_led.temperature` to the flag description or remove it.

examples/translate/translate.go
```diff
@@ -0,0 +1,44 @@
...
+	if err := c.Connect(ctx); err != nil {
+		log.Fatal(err)
+	}
+	defer c.Close()
+
+	root := c.Node(ua.NewTwoByteNodeID(id.ObjectsFolder))
+	nodeId, err := root.TranslateBrowsePathInSameNamespaceToNodeId(uint8(*ns), *nodePath)
+	if err != nil {
+		log.Fatal(err)
+		return
```
magiconair: no need to `return` after `log.Fatal`

node.go
```diff
@@ -141,3 +142,49 @@ func (n *Node) References(refs *ua.NodeID) (*ua.BrowseResponse, error) {
...
+					return resp.Results[0].StatusCode
+				}
+			}
+		}
+		return ua.StatusBadUnexpectedError
+	})
+	return
+}
+
+//translate a browseName relative path to NodeID
```
magiconair: Is `browseNamePath` something that is used in OPC/UA or is this just the value of the `BrowseName` field? If yes, then just call it `browseName`. 

* NodeId -> NodeID
* don't name return values in the func declaration, e.g. `func (n *Node) Translate(...) (*ua.NodeID, error) { ... }`
* Also, try to keep the variable names short if the function is short, e.g.

```go
func ... {
    var names []*ua.QualifiedName
    segments := strings.Split(browseName, ".")
    for _, s := range segments {
        qn := &ua.QualifiedName{NamespaceIndex: uint16(ns), Name: s}
        names = append(names, qn)
    }
    return ...
}
```

magiconair (commented):
Some Go style comments. Once fixed we can merge it.

magiconair:
This is good enough and thanks for the fast response. I may make some minor tweaks after it is merged.

Thank you for your contribution!

PR #238, opened 2019-07-11
Support user-defined types (UDT)

### LINKED ISSUE #213: Add support for UDTs
It could be as simple as registering the type with the decoder but I think there is a bit more work involved. We need this for our own setup but I think this will be useful for others as well. Just adding this as a reminder to pick it up at some point.

magiconair:
This patch adds support for user-defined types which are encoded as
extension objects.

Fixes #213


ua/extension_object.go
```diff
@@ -10,6 +10,38 @@ import (
 	"github.com/gopcua/opcua/id"
 )
 
+// eotypes contains all known extension objects.
+var eotypes = NewTypeRegistry()
+
+// init registers known built-in extension objects.
+func init() {
+	// RegisterExtensionObject(id.AnonymousIdentityToken_Encoding_DefaultBinary, new(AnonymousIdentityToken))
```
alexbrdn: seems to me, there's no need for this commented out code

magiconair: Indeed ...

magiconair:
I'll merge this to move forward. We can iterate on the code if there are issues.

PR #242, opened 2019-07-19
Add support for arrays in variants

### LINKED ISSUE #241: Unable to write arrays due to StatusBadEndOfStream
Below is the code I'm using:

```
func arrayWriteTest(client *opcua.Client) error {
  variant, err := ua.NewVariant(int32(0))
  if err != nil {
    return err
  }
  variant.ArrayLength = 8
  variant.Value = []int32{0, 1, 2, 3, 4, 5, 6, 7}
  variant.EncodingMask = byte(variant.EncodingMask + ua.VariantArrayValues)
  wv := &ua.WriteValue{
    NodeID:      ua.NewStringNodeID(4, "my.node.id"),
    AttributeID: ua.AttributeIDValue,
    Value: &ua.DataValue{
      EncodingMask: ua.DataValueValue,
      Value:        variant,
    },
  }
  req := &ua.WriteRequest{
    NodesToWrite: []*ua.WriteValue{wv},
  }
  _, err = client.Write(req)
  return err
}
```

Here's the logs:
>debug: uasc 1/4: send *ua.WriteRequest with 128 bytes
>debug: conn 1: recv ERRF with 16 bytes
>debug: uasc 1/4: Cannot move beyond end of the stream. StatusBadEndOfStream (0x80B00000)
>debug: Received error: Cannot move beyond end of the stream. StatusBadEndOfStream (0x80B00000)
>2019/07/17 16:30:07.972023 failed array write test: Cannot move beyond end of the stream. StatusBadEndOfStream (0x80B00000)

I'll try to get server logs tomorrow, unless someone is able to reproduce/debug before then.

magiconair:
This patch adds support for one and multi-dimensional arrays in Variant
types.

Fixes #241

### DIFF
--- a/ua/variant.go
+++ b/ua/variant.go
@@ -23,13 +23,13 @@ const (
 //
 // Specification: Part 6, 5.2.2.16
 type Variant struct {
-	// EncodingMask contains the type and the array flags
+	// mask contains the type and the array flags
 	// bits 0:5: built-in type id 1-25
 	// bit 6: array dimensions
 	// bit 7: array values
-	EncodingMask byte
+	mask byte
 
-	// ArrayLength is the number of elements in the array.
+	// arrayLength is the number of elements in the array.
 	// This field is only present if the 'array values'
 	// flag is set.
 	//
@@ -37,24 +37,24 @@ type Variant struct {
 	// field specifies the total number of elements. The original array can be
 	// reconstructed from the dimensions that are encoded after the value
 	// field.
-	ArrayLength int32
+	arrayLength int32
 
-	// ArrayDimensionsLength is the numer of dimensions.
+	// arrayDimensionsLength is the numer of dimensions.
 	// This field is only present if the 'array dimensions' flag
 	// is set.
-	ArrayDimensionsLength int32
+	arrayDimensionsLength int32
 
-	// ArrayDimensions is the size for each dimension.
+	// arrayDimensions is the size for each dimension.
 	// This field is only present if the 'array dimensions' flag
 	// is set.
-	ArrayDimensions []int32
+	arrayDimensions []int32
 
-	Value interface{}
+	value interface{}
 }
 
 func NewVariant(v interface{}) (*Variant, error) {
 	va := &Variant{}
-	if err := va.Set(v); err != nil {
+	if err := va.set(v); err != nil {
 		return nil, err
 	}
 	return va, nil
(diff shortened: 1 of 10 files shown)


kung-foo:
Is there any value in pulling out the array related code into a separate struct? It seems like fields like `ArrayLength` should only be seen _if_ `VariantArrayValues` is set. Additionally, I'm not sure why it is a field and not a function (getter). Since we should be able to discern the value in both decode and encode functions. Otherwise it's just one more thing the developer needs to know (i.e. I didn't know that I _didn't_ need to set `ArrayDimensionsLength` because `Set(...)` does it for me.)

Otherwise, LGTM :)

magiconair:
Now that `Set` does the work I think we can unexport them. `Variant` is somewhat of a beast but it's the only one so I'd like to keep it in one place unless there is a good reason for pulling some code out.

kung-foo:
I was thinking of two structs like `Variant` and `VariantArray` (or slice). This would allow the `Length()` function to only exist on `VariantArray`. So not so much about splitting the code apart, rather ensuring the struct's interface is as minimal as possible. i.e. a `String` `Variant` shouldn't have an `ArrayLength` method at all.

coding out loud here:

```golang
v, _ := NewVariant([]uint32{0, 1, 2}) // v is *VariantArray cast to *Variant?
va, _ := v.AsArray() // returns *VariantArray
va.Length()
va.ElementAt(2) // returns *Variant? (note, I would not create an API like this. This is just to show how array specific function can exist here)
```

ua/variant.go
```diff
@@ -347,6 +457,7 @@ func (m *Variant) Float() float64 {
 	}
 }
 
+// Int returns the int value if the type is one of the int types.
```
kung-foo, line 481: minor nit on the comment/api: this returns an `int` _no matter what_. I'd rather see it return an error if it is not possible to "cast" it to the requested type. I know this makes the API even more verbose, but that's the pain of dynamic typing. It hasn't bit me yet, but I am 100% sure that in the future I will mistakenly ask for the wrong type and it won't be caught in any tests because it returns a real value.
I suppose we could go crazy and add `MustInt() int` and panic on error ;)

magiconair: While going through the code to unexport the variant fields I've noticed that in a lot of places we're just calling `Value()` instead of the specific method. I had the opposite reaction than you. 

I think that in almost all cases you know which type you're getting and the convenience functions should return that typed value without error checking. If you need to check you can always use `Value()` and do a proper type check. 

We should add methods for all supported types like `LocalizedText` and `XmlElement`, ... to complete the set. We could also add `Int8()` and the likes but that might be overkill.

magiconair:
@kung-foo I've unexported all Variant fields and replaced them with functions. Need to think about about the `VariantArray` idea. Maybe we can do that in another iteration if this is a bit more concrete.

ua/variant.go
kung-foo: Noob OPC question: do (server) types change much? Meaning if I write a client assuming that the temperature is coming back as a float, and then an operator changes it to an int, my code will continue to work, but the temperature is now zero.

magiconair: I am running into the same issue with our PLC programmers. The answer is "it depends". There is this concept of a TypeDictionary but that seems quite cumbersome and we don't have code to use that. Schema versioning is a difficult problem and it usually requires coordination. 

Also, I don't want to add too much boilerplate code to the library if you can write a simple function yourself to handle this situation.

magiconair: Also, the same thing happens when they add a new field and don't tell you about it.

magiconair:
I'll wait for @jkissinger for some feedback and then merge this unless there are some more general complaints. Even though this may not be perfect it moves us forward.

magiconair:
Also, @kung-foo if you can test this then this would be appreciated as well.

magiconair:
I'll merge this and tag the next version.

PR #244, opened 2019-07-19
Follow-up to PR #237

magiconair:
* Requests must not have a request header since it will be set by the library
* namespace ids can be uint16
* naming and formatting

See #236
See #237

### DIFF
--- a/examples/translate/translate.go
+++ b/examples/translate/translate.go
@@ -18,8 +18,8 @@ import (
 
 func main() {
 	endpoint := flag.String("endpoint", "opc.tcp://localhost:4840", "OPC UA Endpoint URL")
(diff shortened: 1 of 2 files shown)


magiconair:
@sdghchj I've missed some things in #237 like the `RequestHeader` which should not be there and the `uint16` namespace id. I've also cleaned up the naming and formatting. Can you check whether this is still OK?

sdghchj:
> @sdghchj I've missed some things in #237 like the `RequestHeader` which should not be there and the `uint16` namespace id. I've also cleaned up the naming and formatting. Can you check whether this is still OK?

@magiconair yes, you're right , my mistake, it ok.

magiconair:
Thank you.

Issue #248, opened 2019-07-24
potential memory exhaustion in ua.Variant decoder


kung-foo:
https://github.com/gopcua/opcua/blob/v0.1.4/ua/variant.go#L123-L127

`Variant.Decode` calls `reflect.MakeSlice` with an unchecked value. There is a potential (either malicious or accidental) for memory exhaustion.

magiconair:
What should we use as an upper bound?

—
Frank Schröder

kung-foo:
maybe something like `var MaxVariantArrayLength = 0xffff`? And if a user wants to override it, they can do it via:

```golang
func init() {
    ua.MaxVariantArrayLength = 0x1000000
}
```

magiconair:
Sounds reasonable. I still need to limit recursion as well. Want to send a PR?

magiconair:
Closed by #249

PR #249, opened 2019-07-24
add variant array limit

kung-foo:
fix for #248

### DIFF
--- a/ua/variant.go
+++ b/ua/variant.go
@@ -11,6 +11,11 @@ import (
 	"time"
 )
 
+var (
+	// MaxVariantArrayLength sets a limit on the number of elements in array
+	MaxVariantArrayLength = 0xffff
+)
+
 const (
 	// VariantArrayDimensions flags whether the array has more than one dimension
 	VariantArrayDimensions = 0x40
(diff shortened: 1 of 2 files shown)


magiconair (commented):
1. Could you please rebase the change to master so that the v0.1.4 fix is in. 
1. I think it'd be better to handle the error for the big array as a special test case instead of extending the `CodecTest` this way. 
1. Could you try to find a status code that expresses the error instead of the custom error?

Thx

kung-foo:
lol. if you force push (to first remove a merge commit) and it results in an "empty" PR, it seems like github automatically "merges" and closes it. :man_shrugging: will re-create.

kung-foo:
rebased and test code moved into own func

Regarding the error code, as a user, I want to be able to quickly figure out why I got the error. I could return something like `io.ErrShortBuffer`, but that isn't exactly right because it's not that the buffer is too small, it's that we (arbitrarily) added a safety limit. IMO, errors should clearly explain what is going on, and what the user needs to do to fix it.

magiconair:
So far we've been trying to stick to the available `StatusCode`s for consistency and I'm wondering whether we can use one of them. There is `StatusBadEncodingLimitsExceeded` which seems close but I also get your point. Question is how often that's going to happen to justify an exception for not using a status code. 

Could you have a quick look how FreeOPCUA and open62541 are handling this?

magiconair:
Maybe also squash the commits since now your change to `CodecTest` and the reverse is in. Unless I'm missing something.

kung-foo:
ahh, I see your point re the `StatusCode`. I wasn't familiar with that code. I'll change and squash.

Issue #240, opened 2019-07-16
Discussion on subscriptions


magiconair:
I am starting to work with subscriptions and would like to resurface @kung-foo 's suggestions in https://github.com/gopcua/opcua/pull/207#issuecomment-501418579. I think they seem like a good improvement/addition to the current approach.

Also, I'm having trouble shutting down a client with a subscription properly. It looks like it always times out when the context gets cancelled which is not what I'd expect. But I need to dig a bit deeper. Are your subscription clients shutting down quickly?

magiconair:
Also referring to a post https://github.com/gopcua/opcua/issues/239#issuecomment-511867189 from @jkissinger. Good article from Prosys about timeouts. https://www.prosysopc.com/blog/opc-ua-sessions-subscriptions-and-timeouts/

kung-foo:
What are your thoughts on how closely the golang API should follow the OPC API as defined by the network protocol?

My general approach is that I like having an opinionated, high level API that is designed from the perspective of the SDK user and follows the "zen" of the language. If the "opinion" is wrong, then I (as the user) can fall back to the lower level API that closely (or exactly) matches the network protocol.

An example is `MonitoringParameters.ClientHandle`. It doesn't seem like that ever needs to be exposed. From a user's perspective, it is just an implementation detail that they don't care about. Much in the same way that pagination APIs usually have some concept of a continuation token.

Elasticsearch's Python SDK has a package of "helpers" that hide some implementation details. Another example is the AWS CLI. It has `aws s3 ...` which has all the high level functions you expect (`cp`, `rm`, `ls`) and `aws s3api ...` which has functions which map directly to RESTful calls (`get-object`, `head-object`, `put-bucket-replication`, etc).

I don't know the network protocol well enough yet to know what to have a strong opinion on.

magiconair:
I think that's a good general guideline. Close to the API the lower you get and more abstractions the higher you get.

We already have functions in the client that map directly to the API calls with (hopefully) sane defaults. If that does not work you can just copy the function and send the request yourself.

For one of our projects I've just written a `NodeMonitor` based on your ideas in the comment which abstracts a lot of the subscription handling away including the `ClientHandle` but this was just a 30 min work of moving existing code around. I'm all for something more generic and robust. Things like these are definitely useful but could/should live in a separate package. 

If you can work with the existing api then add your code on top and make small tweaks where necessary. If you think the api or the implementation need a general refactor to make this work then I'd suggest two separate PRs.

```go
[...]

PR #252, opened 2019-07-30
Node monitor (extended subscription)

kung-foo:
This PR creates a new `NodeMonitor` struct for creating managed subscriptions. It attempts to hide implementation details from the user, and provide a simple, high-level interface for interacting with data change events.
 
see related discussion: https://github.com/gopcua/opcua/issues/240

The example code shows the two different methods for a subscription: callback-based, and channel-based.


kung-foo:
There are some naming/structure issues I'd still like to resolve. Possibly merge `NodeMonitor` and `Subscription` into a single struct and move `Subscribe(...)` and `ChanSubscribe(...)` to package level functions (both returning the new, unified struct).

magiconair:
This looks like a good start. Maybe call it `monitor.Subscription` instead of `monitor.New`. We might have other monitors. I'll need something that reads a list of nodes on a regular basis and that would fit in this quite nicely (e.g. `monitor.Nodes`).

Do we have to support both channel and callback mechanisms? Wouldn't channel be enough? 

Also, you can add nodes as strings and `NodeIDs` but when you set up the monitor you can use only strings. Don't have a good idea on how to solve that right now but might be something to keep in mind. With Channel and Callback you potentially get four different combinations. 

You probably don't want to `panic` when the subscription id doesn't match ... :)

kung-foo:
Thanks for the feedback. Regarding chan v callback, my favorite APIs are the ones that usually provide both. From this somewhat hyperbolic article: https://www.jtolio.com/2016/03/go-channels-are-bad-and-you-should-feel-bad/
> [...] your API can always be more general, always more flexible, and take drastically less resources if you use callbacks instead of channels.

kung-foo:
I should point out that the _current_ callback implementation is actually _less_ performant the the channel method because the callback code uses a channel behind the scenes. However, that could be removed with a little bit of work.

subscription.go
```diff
@@ -196,6 +198,29 @@ func (s *Subscription) sendNotification(ctx context.Context, data *PublishNotifi
 
 }
 
+// Stats returns a diagnostic struct with metadata about the current subscription
+func (s *Subscription) Stats() (*ua.SubscriptionDiagnosticsDataType, error) {
+	// TODO(kung-foo): once browsing feature is merged, attempt to get direct access to the
```
magiconair, line 203: does this refer to the more browsing branch or something else?

magiconair:
@kung-foo are you using this code right now or are you planning on using it?

subscription.go
kung-foo: yes. though I haven't tried it on this exact use case. So it is possible that I can't get access to the single diag node. Also, I noticed that with the FreeOPCUA server, no diag nodes are generated. Is there a standard way to indicate that that feature is not supported?

kung-foo:
Yes, we are using it, though not in production.

subscription.go
magiconair: I don't know.

magiconair:
How did you find the timer leak?

kung-foo:
re the timer leak: manual inspection. I just happen to remember that "idiosyncrasy" of the timer API.

magiconair:
@kung-foo How far are we from merging this? I'd like to start using it.

kung-foo:
I think we are good to merge. We can re-visit features like reconnecting to a subscription in a future PR.

magiconair:
Then lets do it.

fdamador:
Thank you for creating the monitor code which i've been testing.  

I have a need to pass the msg.value.value() values data back to my own custom struct array I have in the main() routine for a series of NodeIDs.  Can I do this by interrogating the subscription server?  Or do I have to send the NodeIds data (time, value, quality, etc) via channel?

PR #255, opened 2019-08-05
check for invalid/duplicate subscription id

### LINKED ISSUE #254: Prosys Simulation Server returns duplicate subscription IDs
When creating multiple subscriptions in quick succession, I have seen the same sub ID come back. This causes an overwrite here: https://github.com/gopcua/opcua/blob/v0.1.4/client.go#L556
Which then causes data change notifications to be dispatched to the wrong handler.

Quick fix is to check for an existing Subscription ID, and return an error. But since the subscription has already been created on the server, do we need to cleanup?

Or is this technically a valid response?

Log: [prosys.log](https://github.com/gopcua/opcua/files/3461440/prosys.log)

You can see two `CreateSubscriptionResponse` come back with `SubscriptionId=28`

image

kung-foo:
fixes #254

### DIFF
--- a/client.go
+++ b/client.go
@@ -553,6 +553,12 @@ func (c *Client) Subscribe(params *SubscriptionParameters) (*Subscription, error
 		c,
 	}
 	c.subMux.Lock()
+	if sub.SubscriptionID == 0 || c.subscriptions[sub.SubscriptionID] != nil {
+		// this should not happen
+		// see: Part 4 Section 5.13.2.2, Table 88 – CreateSubscription Service Parameters
+		c.subMux.Unlock()
+		return sub, ua.StatusBadSubscriptionIDInvalid
+	}
 	c.subscriptions[sub.SubscriptionID] = sub
 	c.subMux.Unlock()
 
--- a/client.go
+++ b/client.go
@@ -554,10 +554,10 @@ func (c *Client) Subscribe(params *SubscriptionParameters) (*Subscription, error
 	}
 	c.subMux.Lock()
 	if sub.SubscriptionID == 0 || c.subscriptions[sub.SubscriptionID] != nil {
-		// this should not happen
+		// this should not happen and is usually indicative of a server bug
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -553,6 +553,12 @@ func (c *Client) Subscribe(params *SubscriptionParameters) (*Subscription, error
 		c,
 	}
 	c.subMux.Lock()
+	if _, ok := c.subscriptions[sub.SubscriptionID]; ok {
```
magiconair: `SubscriptionID` is an `IntegerId` (Part 4, 7.14) which cannot be zero. So we should also check for `0` and I think you can simplify the check to 

```go
if c.subscriptions[sub.SubscriptionID] != nil {
	c.subMux.Unlock()
	return sub, ua.StatusBadSubscriptionIDInvalid
}
```

client.go
```diff
@@ -553,6 +553,12 @@ func (c *Client) Subscribe(params *SubscriptionParameters) (*Subscription, error
 		c,
 	}
 	c.subMux.Lock()
+	if sub.SubscriptionID == 0 || c.subscriptions[sub.SubscriptionID] != nil {
```
magiconair, line 556: We could check the subscription id right after the `StatusOK` check and shouldn't you return `nil` instead of `sub` on error?

kung-foo: Yeah, I thought about returning `nil`, but _technically_ subscription resources _have_ been created on the server. But realistically, it would be hard to cleanup because I think most operations use the ID. I usually think of `nil` as indicating that nothing was created.

magiconair: The clean-up will never be called b/c the caller checks the error and moves on unless we specifically document that behavior. Then there is nothing useful the caller can do with this. If there is useful cleanup you can do in the client then you should do it here. However, the whole point is that this checks for bugs and edge cases in the server. Lets just return `nil` here unless I'm missing smth substantial.

magiconair:
LGTM

PR #257, opened 2019-08-13
First stab at integration tests

magiconair:
This patch creates a separate integration test suite which uses the python server from https://github.com/FreeOPCUA until we have our own server implementation. 

The approach is similar to the `httptest` package in that it provides a `Server` which you initialize with a `path` to the Python server. It panics if the startup fails. Currently, the server is expected to run on `127.0.0.1:4840` and have no security. 

The python server needs between 1-2 seconds of startup. The `Server` waits a couple of seconds until the endpoint is available before continuing. Because of this the integration tests are hidden behind a build flag, `go test -tags=integration ./uatest`. Got the inspiration from https://medium.com/@povilasve/go-advanced-tips-tricks-a872503ac859

This should provide a platform for writing integration tests. 

The following things still need solving:

* How to ensure that the `opcua` package is installed? Do we care about the version? (I've added a `make install-py-opcua` target to make this easier)
* Not sure if the path resolution to the python server is done properly. My guess is that this breaks when the test is in a different package.

@alexbrdn and @kung-foo This could be useful for subscription testing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,9 @@ jobs:
       - restore_cache:
           keys:
             - go-mod-{{ checksum "go.sum" }}
+      - run: pip3 install opcua
       - run: go test -v $GO_TEST_FLAGS ./...
+      - run: go test -v -tags=integration ./uatest
       - run: go install ./...
       - save_cache:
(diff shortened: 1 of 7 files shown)


magiconair:
Note that this is PR has only one simple integration test to establish the platform.

PR #260, opened 2019-08-20
Do not send the certificate thumbprint for security mode None

### LINKED ISSUE #259: Siemens P7-1500 refuses connection
Today I've observed a Siemens P7-1500 returning a `0x80000000 Bad` error as response to an `OpenSecureChannelRequest` while establishing a connection with security mode `None`. We have not seen this behavior before. 

This happened after discovering the endpoints successfully with `GetEnpoints` and then choosing one with `SelectEndpoint` for the given security mode. If the client is not using `SelectEndpoint` the connection was established.

The issue is that the server returns a certificate for all endpoints - including for security mode `None` and the gopcua client sends the thumbprint of that certificate in the `OpenSecureChannelRequest`. It looks like the Siemens P7-1500 refuses to connect with `None` and a non-empty `ReceiverCertificateThumbprint`. 

To fix this the `ReceiverCertificateThumbprint` field in the `RequestHeader` should be empty when the security mode is `None`.

magiconair:
Fixes #259

### DIFF
--- a/uasc/message.go
+++ b/uasc/message.go
@@ -105,10 +105,19 @@ type Message struct {
 func NewMessage(srv interface{}, typeID uint16, cfg *Config) *Message {
 	switch typeID {
 	case id.OpenSecureChannelRequest_Encoding_DefaultBinary, id.OpenSecureChannelResponse_Encoding_DefaultBinary:
+		// Do not send the thumbprint for security mode None
+		// even if we have a certificate.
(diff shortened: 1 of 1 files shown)


magiconair:
Not sure if this is the best place to fix this. Somewhere hidden.

dwhutchison (approved):
LGTM

PR #232, opened 2019-07-01
fix(secure_channel): fix data race in Receive()

alexbrdn:
fixed by removing the offending line (see https://github.com/gopcua/opcua/issues/231#issuecomment-506775781)

closes #231

Issue #231, opened 2019-06-28
Data Race in SecureChannel (yet another)


alexbrdn:
With #227 applied and under the same conditions as in #225, I get the following data race in SecureChannel:
```
==================
WARNING: DATA RACE
Write at 0x00c00013806c by goroutine 17:
  github.com/gopcua/opcua/uasc.(*SecureChannel).Receive()
      .../opcua/uasc/secure_channel.go:354 +0x251
  github.com/gopcua/opcua.(*Client).monitorChannel()
      .../opcua/client.go:191 +0xbe

Previous write at 0x00c00013806c by main goroutine:
  github.com/gopcua/opcua/uasc.(*SecureChannel).sendAsyncWithTimeout()
      .../opcua/uasc/secure_channel.go:176 +0x481
  github.com/gopcua/opcua/uasc.(*SecureChannel).SendWithTimeout()
      .../opcua/uasc/secure_channel.go:154 +0xe3
  github.com/gopcua/opcua.(*Client).sendWithTimeout()
      .../opcua/client.go:439 +0xd3
  github.com/gopcua/opcua.(*Client).closeSession()
      .../opcua/client.go:428 +0x1bd
  github.com/gopcua/opcua.(*Client).CloseSession()
      .../opcua/client.go:396 +0x7c
  github.com/gopcua/opcua.(*Client).Close()
      .../opcua/client.go:211 +0x43
  main.main()
      .../opcua/examples/subscribe/subscribe.go:107 +0x119a

Goroutine 17 (running) created at:
  github.com/gopcua/opcua.(*Client).Dial()
      .../opcua/client.go:174 +0x2fd
  github.com/gopcua/opcua.(*Client).Connect()
      .../opcua/client.go:139 +0x17b
  main.main()
      .../opcua/examples/subscribe/subscribe.go:60 +0xa99
==================
Found 1 data race(s)
```
I guess this line from #211 was the last addition leading to data race: `s.cfg.RequestID = reqid`.

I don't think that Receive(), when executed in client, should change `s.cfg.RequestID`. Whether it should do the same when running as server, I'm not sure. @dwhutchison, help needed.

dwhutchison:
You're right; that was for the server and shouldn't execute for clients.  I put that in there so the server responses echo the request ID back to the client.  Feel free to delete it for now and I'll find a better way to do this in the server code

alexbrdn:
Addendum for the sake of completeness.

[...]

PR #261, opened 2019-09-01
Remove reflection from SecureChannel code

magiconair:
This patch removes reflection code in the secure channel.

The Secure Channel implementation relied on reflection to detect service
requests and responses and to get to their respective header fields.

This patch extends the code generator to generate helper methods for
getting and setting the RequestHeader and ResponseHeader fields so that
the reflection code can be removed with type safe code using interfaces.

### DIFF
--- a/uasc/secure_channel.go
+++ b/uasc/secure_channel.go
@@ -12,7 +12,6 @@ import (
 	"fmt"
 	"io"
 	"math"
-	"reflect"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -116,12 +115,12 @@ func (s *SecureChannel) hasState(n int32) bool {
 }
 
 // Send sends the service request and calls h with the response.
-func (s *SecureChannel) Send(svc interface{}, authToken *ua.NodeID, h func(interface{}) error) error {
+func (s *SecureChannel) Send(svc ua.Request, authToken *ua.NodeID, h func(interface{}) error) error {
 	return s.SendWithTimeout(svc, authToken, s.cfg.RequestTimeout, h)
 }
 
 // SendWithTimeout sends the service request and calls h with the response with a specific timeout.
-func (s *SecureChannel) SendWithTimeout(svc interface{}, authToken *ua.NodeID, timeout time.Duration, h func(interface{}) error) error {
+func (s *SecureChannel) SendWithTimeout(svc ua.Request, authToken *ua.NodeID, timeout time.Duration, h func(interface{}) error) error {
 	respRequired := h != nil
 
 	ch, reqid, err := s.SendAsync(svc, authToken, respRequired)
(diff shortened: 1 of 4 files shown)


magiconair:
Finally found a way to remove the `reflect` hack.

uasc/secure_channel.go
```diff
@@ -116,12 +115,12 @@ func (s *SecureChannel) hasState(n int32) bool {
 }
 
 // Send sends the service request and calls h with the response.
-func (s *SecureChannel) Send(svc interface{}, authToken *ua.NodeID, h func(interface{}) error) error {
+func (s *SecureChannel) Send(svc ua.Request, authToken *ua.NodeID, h func(interface{}) error) error {
```
dwhutchison: This should arguably be called `SendRequest` now

dwhutchison (commented):
I really like this change; things are much clearer.  I ran this against the server code and it's mostly compatible, but the following patch keeps everything working.  It's not extensively tested but it still functions with the basic operations I had working previously.
@alexbrdn, the only part I'm worried about is the data race; could you try replicate the data race from #231 with this patch?  I added the offending line back in, but wrapped it in a type check so it applies to requests only.

```diff
diff --git a/uasc/secure_channel.go b/uasc/secure_channel.go
index ad93b99..d5717d2 100644
--- a/uasc/secure_channel.go
+++ b/uasc/secure_channel.go
@@ -217,11 +217,11 @@ func (s *SecureChannel) sendAsyncWithTimeout(svc ua.Request, authToken *ua.NodeI
 	return resp, reqid, nil
 }
 
-// sendResponse sends a service response.
+// SendResponse sends a service response.
 // todo(fs): this method is most likely needed for the server and we haven't tested it yet.
 // todo(fs): it exists to implement the handleOpenSecureChannelRequest() method during the
 // todo(fs): refactor to remove the reflect code. It will likely change.
-func (s *SecureChannel) sendResponse(svc ua.Response) error {
+func (s *SecureChannel) SendResponse(svc ua.Response) error {
 	typeID := ua.ServiceTypeID(svc)
 	if typeID == 0 {
 		return fmt.Errorf("unknown service %T. Did you call register?", svc)
@@ -369,6 +369,11 @@ func (s *SecureChannel) Receive(ctx context.Context) Response {
 			} else {
 				debug.Printf("uasc %d/%d: recv %T", s.c.ID(), reqid, svc)
 			}
+			
+			// Revert data race fix from #232 with an additional type check
+			if _, ok := svc.(ua.Request); ok {
+				s.cfg.RequestID = reqid
+			}
 
 			switch svc.(type) {
 			case *ua.OpenSecureChannelRequest:
@@ -488,15 +493,13 @@ func (s *SecureChannel) receive(ctx context.Context) (uint32, interface{}, error
 			}
 
 			// If the service status is not OK then bubble
-			// that error up to the caller.
-			resp, ok := svc.(ua.Response)
-			if !ok {
-				return reqid, nil, fmt.Errorf("not a service response: %T", svc)
-			}
-			status := resp.Header().ServiceResult
-			debug.Printf("uasc %d/%d: res:%v", s.c.ID(), reqid, status)
-			if status != ua.StatusOK {
-				return reqid, svc, status
+			// that error up to the caller.  Skip the check for Requests
+			if resp, ok := svc.(ua.Response); ok {
+				status := resp.Header().ServiceResult
+				debug.Printf("uasc %d/%d: res:%v", s.c.ID(), reqid, status)
+				if status != ua.StatusOK {
+					return reqid, svc, status
+				}	
 			}
 			return reqid, svc, err
 		}
@@ -678,7 +681,7 @@ func (s *SecureChannel) handleOpenSecureChannelRequest(svc interface{}) error {
 		ServerNonce: nonce,
 	}
 
-	if err := s.sendResponse(resp); err != nil {
+	if err := s.SendResponse(resp); err != nil {
 		return err
 	}
``

uasc/secure_channel.go
magiconair: Yes.

magiconair:
@dwhutchison Could you push your rebased server branch once this is merged? I've started playing yesterday with a server implementation forgetting that you are already working on it. Maybe we can start using it for some tests or start writing tests for it while it's being developed.

magiconair:
I'll merge this. Maybe @alexbrdn is busy.

alexbrdn:
Sorry, guys. Been away for a while. I'll try to reproduce the data race and will let you know here.

magiconair:
No worries and thank you @alexbrdn :)

alexbrdn:
I can't reproduce the data race with code as is, yet can reliably reproduce it if I remove the type check. I think it's safe to say that this data race is fixed for good.

magiconair:
Yay!

PR #265, opened 2019-09-09
Add an IsConnected function to Client

fouwels:
Add an isConnected to the client - needed to determine whether a connection has been established prior to making a request.

Calling read/write functions on a non established connection from the library returns a panic!

Can call connect regardless on each request, but calling on an established connection returns a generic string error, so we cannot programmatically determine whether connect failed due to a failure, or due to being an existing connection, without parsing the result of err.Error(), which is a bad solution.

Need to either move to implementing proper (non string) error types, or add the attached function allowing it to be checked manually.

### DIFF
--- a/client.go
+++ b/client.go
@@ -154,6 +154,15 @@ func (c *Client) Connect(ctx context.Context) (err error) {
 	return nil
 }
 
+// IsConnected returns whether the a secure channel is currently established.
+func (c *Client) IsConnected(ctx context.Context) bool {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+
+	return c.sechan != nil
+}
+
 // Dial establishes a secure channel.
 func (c *Client) Dial(ctx context.Context) error {
 	if ctx == nil {


client.go
```diff
@@ -154,6 +154,15 @@ func (c *Client) Connect(ctx context.Context) (err error) {
 	return nil
 }
 
+// IsConnected returns whether the a secure channel is currently established.
+func (c *Client) IsConnected(ctx context.Context) bool {
+	if ctx == nil {
+		ctx = context.Background()
```
golangcibot, line 160: ineffectual assignment to `ctx` (from `ineffassign`)

magiconair:
Hmm, if the library panics when the connection is not established then we should address that. Something like `IsConnected` could be useful but I'd prefer if we address the panics first. Can you provide an example?

fouwels:
I've added an example as https://github.com/gopcua/opcua/issues/297

magiconair:
The `client.State()` method now provides the current connection state. You can check for `client.State() == opcua.Connected` if the client is connected.

PR #266, opened 2019-09-09
Variant custom getters

### LINKED ISSUE #245: ua.Variant should have custom getters for all supported types
See https://github.com/gopcua/opcua/pull/242#discussion_r305330777

magiconair:
Custom getter methods for all `Variant` types.

Fixes #245

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -16,3 +16,22 @@ func Errorf(format string, a ...interface{}) error {
 func New(text string) error {
 	return pkg_errors.New(Prefix + text)
 }
+
+// Equal returns true if the two errors have the same error message.
+//
+// todo(fs): the reason we need this function and cannot just use
+// todo(fs): reflect.DeepEqual(err1, err2) is that by using github.com/pkg/errors
+// todo(fs): the underlying stack traces change and because of this the errors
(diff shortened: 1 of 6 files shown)


magiconair:
Ah, wasn't done with the tests for all the types.

magiconair:
I'm also not sure what to do for `StatusCode`:

```go
func (m *Variant) StatusCode() StatusCode {
 	switch m.Type() {
 	case TypeIDStatusCode:
 		return m.value.(StatusCode)
 	default:
 		// todo(fs): should this be 0 == StatusOK ??
 		return StatusBad
 	}
 }
```

magiconair:
I've decided to return `StatusBadTypeMismatch` in case you call `Variant.StatusCode()` but the variant value is not a status code. Returning `StatusOK` feels really wrong and may hide some errors and `StatusBad` felt too generic. If you feel strongly in a different way or if this differs from other implementations then please let me know.

I also added some more tests for arrays in variants and while doing that added checks for negative lengths and dimensions to make the code more robust.

magiconair:
I'm going to merge this now but please feel free to comment on the decisions and I'll patch them later.

PR #271, opened 2019-09-16
Make counter roll-overs more robust

### LINKED ISSUE #189: Make SequenceNumber counter more robust
Follow-up to #184 and https://github.com/gopcua/opcua/issues/184#issuecomment-490278405

In short, `SequenceNumber` should not have large gaps and need to roll-over properly.

magiconair:
Fixes #189


dwhutchison (commented):
This looks good to me.  

This reminded me that the receiving logic needs to validate the received SequenceNumber.  I'll open an issue for that; no need to make this bigger than it needs to be.

PR #278, opened 2019-09-25
🚧 Add dial timeout and cleanup ResolveEndpoint

kung-foo:
This PR adds a default timeout to the underlying `Dial` calls. This affects both the name resolution and the actual TCP setup. I also cleaned up the resolve logic. I left the logic in place that explicitly resolves a host name to a single IP address because I don't know the original reasons (I don't see any need for it. I would rather just create a `ValidateEndpoint` and let the regular dialer do its thing.)

I originally tried to enable a timeout by adding a `context.Timeout` to the call to `Client.Dial` (https://github.com/Intelecy/opcua/blob/net-ctx/client.go#L158-L189), but that backfired as the incoming context is also used for signaling, and not just the dial itself (as the name would imply).

TODO: more testing as I've only tested this in one local configuration

### DIFF
--- a/uacp/conn.go
+++ b/uacp/conn.go
@@ -47,7 +47,8 @@ func Dial(ctx context.Context, endpoint string) (*Conn, error) {
 		return nil, err
 	}
 
-	dialer := net.Dialer{}
+	var dialer net.Dialer
+
 	c, err := dialer.DialContext(ctx, network, url.Host)
 	if err != nil {
 		return nil, err
--- a/uacp/conn.go
+++ b/uacp/conn.go
@@ -24,6 +24,8 @@ const (
 	DefaultSendBufSize    = 0xffff
 	DefaultMaxChunkCount  = 512
 	DefaultMaxMessageSize = 2 * MB
+
+	DefaultDialTimeout = time.Second * 10
 )
 
 // connid stores the current connection id. updated with atomic.AddUint32
@@ -36,11 +38,17 @@ func nextid() uint32 {
 
 func Dial(ctx context.Context, endpoint string) (*Conn, error) {
 	debug.Printf("Connect to %s", endpoint)
-	network, raddr, err := ResolveEndpoint(endpoint)
+
+	ctx, cancel := context.WithTimeout(ctx, DefaultDialTimeout)
+	defer cancel()
+
+	network, url, err := ResolveEndpoint(ctx, endpoint)
 	if err != nil {
 		return nil, err
 	}
-	c, err := net.DialTCP(network, nil, raddr)
+
+	dialer := net.Dialer{}
+	c, err := dialer.DialContext(ctx, network, url.Host)
 	if err != nil {
 		return nil, err
 	}
@@ -79,7 +87,7 @@ type Listener struct {
 // If the IP field of laddr is nil or an unspecified IP address, Listen listens
 // on all available unicast and anycast IP addresses of the local system.
 // If the Port field of laddr is 0, a port number is automatically chosen.
-func Listen(endpoint string, ack *Acknowledge) (*Listener, error) {
+func Listen(ctx context.Context, endpoint string, ack *Acknowledge) (*Listener, error) {
 	if ack == nil {
 		ack = &Acknowledge{
 			ReceiveBufSize: DefaultReceiveBufSize,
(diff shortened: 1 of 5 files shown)


uacp/endpoint.go
```diff
@@ -5,30 +5,47 @@
...
-	switch err.(type) {
-	case *net.DNSError:
-		return "", nil, fmt.Errorf("could not resolve address %s", addrString)
+
+	port := url.Port()
+	if port == "" {
+		port = defaultPort
+	}
+
+	resolver := net.Resolver{}
```
magiconair:
```go
var resolver net.Resolver
```

uacp/endpoint_test.go
```diff
@@ -5,72 +5,77 @@
...
-		}
-		if got, want := network, c.network; got != want {
-			t.Fatalf("got network %q want %q", got, want)
-		}
-		if got, want := addr.String(), c.addr.String(); got != want {
-			t.Fatalf("got addr %q want %q", got, want)
-		}
-		if got, want := errStr, c.errStr; got != want {
-			t.Fatalf("got error %q want %q", got, want)
+			if got, want := errStr, c.errStr; got != want {
```
magiconair, line 69: This handles both the `nil` and `not-nil` case without the `if` statement.

```go
if got, want := err, tt.err; !reflect.DeepEqual(got, want) {
    t.Fatalf("got error %v want %v", got, want)
}
```

magiconair (commented):
Looks good. Lets test it a bit more

uacp/endpoint_test.go
kung-foo: I was getting a panic above on `errStr = err.Error()` since `err` was `nil`.

kung-foo:
I am not sure what the proper value should be for this:

`DefaultDialTimeout = time.Second * 10`

uacp/endpoint_test.go
magiconair: yes, you need to refactor the test to have an `error` instead of an `errStr` in the test case definition.

magiconair:
Maybe by default we should not time out.

dwhutchison (commented):
Looks good; the biggest comment is how we can expose the timeout to the user.  This is especially needed if we remove the default timeout as @magiconair suggested.

uacp/conn.go
```diff
@@ -36,11 +38,18 @@ func nextid() uint32 {
 
 func Dial(ctx context.Context, endpoint string) (*Conn, error) {
 	debug.Printf("Connect to %s", endpoint)
-	network, raddr, err := ResolveEndpoint(endpoint)
+
+	ctx, cancel := context.WithTimeout(ctx, DefaultDialTimeout)
```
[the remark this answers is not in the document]
kung-foo: Currently the only way would be to set `uacp.DefaultDialTimeout` first. What I'd rather have is to be able to use a context with timeout here: https://github.com/Intelecy/opcua/blob/net-ctx/client.go#L158-L189.
But that requires more refactoring as the incoming context is used for the monitoring routine.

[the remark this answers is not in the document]
dwhutchison: Right, but `uacp.DefaultDialTimeout` is `const` so it can't be configured at runtime.  Changing it to be non-const means users would need to import `uacp` and modify a global variable to change the setting.

I agree that putting it in the client is the right spot which is why I think having `uacp.Dial` just accept a timeout parameter (or making a `uacp.DialWithTimeout` method and have `Dial` be a wrapper for it with the default timeout in place) could work.  Then the client's `Dial` method just changes to:
`conn, err := uacp.DialWithTimeout(ctx, c.endpointURL, c.cfg.DialTimeout)` and your changes still work (replacing `DefaultDialTimeout` with the parameter)

Alternatively, it could be possible to create a second context in the the client's `Dial` method with a timeout as long as it's not used as the parent for the monitor routine's context and that shouldn't cause issues. 
eg:
```go
func (c *Client) Dial(ctx context.Context) error {
...

    dialCtx, dialCancel := context.WithTimeout(ctx, c.cfg.DialTimeout)
    defer dialCancel()
    conn, err := uacp.Dial(dialCtx, c.endpointURL)
    if err != nil {
	    return err
    }

...

    ctx, c.cancelMonitor = context.WithCancel(ctx)
    go c.monitorChannel(ctx)

...
```
While this doesn't seem as clear to me, it might make having no timeout simpler.

uacp/endpoint_test.go
magiconair: we now have `errors.Equal`

PR #279, opened 2019-09-25
Add error prefix

### LINKED ISSUE #253: Error strings should be prefixed
To ease debugging, all errors returned by the `opcua` module should be prefixed with `opcua: `

Example:

```golang
var ErrSlowConsumer = errors.New("opcua: slow consumer. messages dropped")
```

kung-foo:
Add new `opcua/errors` package with wrappers for `fmt.Errorf` and `errors.New`.

Fixes: https://github.com/gopcua/opcua/issues/253

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 vendor/
 dist/
+__pycache__/
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ The current set of supported services is only for the high-level client.
 |                             | Write                         | Yes       |              |
 |                             | HistoryRead                   | Yes       |              |
 |                             | HistoryUpdate                 |           |              |
-| Method Service Set          | Call                          |           |              |
+| Method Service Set          | Call                          | Yes       |              |
 | MonitoredItems Service Set  | CreateMonitoredItems          | Yes       |              |
 |                             | DeleteMonitoredItems          | Yes       |              |
 |                             | ModifyMonitoredItems          |           |              |
--- a/client.go
+++ b/client.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"github.com/gopcua/opcua/debug"
+	"github.com/gopcua/opcua/errors"
 	"github.com/gopcua/opcua/id"
 	"github.com/gopcua/opcua/ua"
 	"github.com/gopcua/opcua/uacp"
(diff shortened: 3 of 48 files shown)


magiconair:
Should we use an Error type and wrap properly? Still like that stack traces from pkg/errors

kung-foo:
I can change it to:

```go
package errors

import (
	pkg_errors "github.com/pkg/errors"
)

// Prefix is the default error string prefix
const Prefix = "opcua: "

// Errorf is a wrapper for `fmt.Errorf`
func Errorf(format string, a ...interface{}) error {
	return pkg_errors.Errorf(Prefix+format, a...)
}

// New is a wrapper for `errors.New`
func New(text string) error {
	return pkg_errors.New(Prefix + text)
}
```

Which results in stack traces like:

```
opcua: hello
github.com/gopcua/opcua/errors.New
	/home/jonathan/gohack/github.com/gopcua/opcua/errors/errors.go:17
github.com/gopcua/opcua/errors.TestErrors
	/home/jonathan/gohack/github.com/gopcua/opcua/errors/errors_test.go:14
testing.tRunner
	/home/jonathan/.gvm/gos/go1.13/src/testing/testing.go:909
runtime.goexit
	/home/jonathan/.gvm/gos/go1.13/src/runtime/asm_amd64.s:1357
```

Which is ok, _except_ for the fact that the top of the trace is our convenience function.

There are a few other libs out there that allow a `skip` N frames parameter. :shrug:

magiconair:
I think this may need a rebase and force push

kung-foo:
oh, I know :) I like to do my dirty deeds in public.

magiconair:
Ha, re stack trace. Looks like there is a slick way of doing this. Check out the example in https://godoc.org/github.com/pkg/errors#hdr-Retrieving_the_stack_trace_of_an_error_or_wrapper

kung-foo:
The issue is that the stack frame skip is hardcoded here: https://github.com/pkg/errors/blob/master/stack.go#L166

_technically_ we'd want to bump that to 4 since we want to skip our own error func

magiconair:
hrm, unless there is a better way lets just use `pkg/errors`. I'll be at golab.io and am supposed to meet rsc and the rest of the gang. Wanted to talk about errors anyway.

magiconair:
Can you squash and rebase so that we can merge it?

magiconair:
Then I'll close v0.1.6 and release it.

kung-foo:
grr. i have seriously foobar'd this branch. abandoning.

PR #280, opened 2019-09-26
Fix custom timeouts not being utilized from SendRequestWithTimeout

### LINKED ISSUE #270: The sequence number is unknown to the server. StatusBadSequenceNumberUnknown (0x807A0000)
When I subscribe tags, it returns an error: The sequence number is unknown to the server. StatusBadSequenceNumberUnknown (0x807A0000)
What the means of this error?

dwhutchison:
fixes #270


magiconair:
Do we still need `SendAsync` after this? I think we can remove it.

dwhutchison:
> Do we still need `SendAsync` after this? I think we can remove it.

Good catch; I don't see it utilized anymore so I'll remove it.

PR #282, opened 2019-09-26
Fixed Browse Failure for Certain Kinds of OPC Nodes.

haodreams:
OPC UA Part 4
 7.37 ViewDescription
 The components of this parameter are defined in Table 175.
 The time date desired. The corresponding version is the one with the closest
 previous creation timestamp. Either the Timestamp or the viewVersion
 parameter may be set by a Client, but not both. If ViewVersion is set this
 parameter shall be null.


magiconair:
This was already fixed in 0.12 by 760b029

PR #286, opened 2019-09-27
Handle implied ns=0

### LINKED ISSUE #285: ua.ParseNodeID fails to parse strings without namespace
`ua.ParseNodeID` should handle a string like `i=2253`.

From Part 6 / 5.3.1.10:

> The NamespaceIndex formatted as a base 10 number.
> If the index is 0 then the entire 'ns=0;' clause shall be omitted.

Invalid test:
https://github.com/gopcua/opcua/blob/v0.1.5/ua/node_id_test.go#L134

kung-foo:
Fixes #285 

Note: `ParseNodeID` is still quite brittle. Things like whitespace and ordering will cause it to not parse.

### DIFF
--- a/ua/node_id.go
+++ b/ua/node_id.go
@@ -92,11 +92,17 @@ func ParseNodeID(s string) (*NodeID, error) {
 		return NewTwoByteNodeID(0), nil
 	}
 
+	var nsval, idval string
+
 	p := strings.SplitN(s, ";", 2)
-	if len(p) < 2 {
+	switch len(p) {
+	case 1:
+		nsval, idval = "ns=0", p[0]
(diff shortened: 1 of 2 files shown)


magiconair:
What do you mean with ordering? I could probably write a proper parser.

kung-foo:
I think `i=83;ns=0` would fail.

magiconair:
It should, shouldn't it? That's invalid, right?

magiconair:
If not, then of course the whole parser is wrong.

kung-foo:
ah, I guess the spec requires a strict order. it's a case of "not the way I would do it" :)

PR #291, opened 2019-10-02
Add common error prefix

### LINKED ISSUE #253: Error strings should be prefixed

kung-foo:
Fixes #253

### DIFF
--- a/ua/node_id_test.go
+++ b/ua/node_id_test.go
@@ -6,9 +6,10 @@ package ua
 
 import (
 	"encoding/base64"
-	"errors"
 	"reflect"
 	"testing"
+
+	"github.com/gopcua/opcua/errors"
 )
 
 func TestNodeID(t *testing.T) {
@@ -146,7 +147,7 @@ func TestParseNodeID(t *testing.T) {
 	for _, c := range cases {
 		t.Run(c.s, func(t *testing.T) {
 			n, err := ParseNodeID(c.s)
-			if got, want := err, c.err; !reflect.DeepEqual(got, want) {
+			if got, want := err, c.err; !compareErrors(got, want) {
 				t.Fatalf("got error %v want %v", got, want)
 			}
 			if got, want := n, c.n; !reflect.DeepEqual(got, want) {
@@ -244,7 +245,7 @@ func TestSetIntID(t *testing.T) {
 			}
 
 			err := tt.n.SetIntID(tt.v)
-			if got, want := err, tt.err; !reflect.DeepEqual(got, want) {
+			if got, want := err, tt.err; !compareErrors(got, want) {
 				t.Fatalf("got error %v want %v", got, want)
 			}
 			// if the test should fail and the error was correct
@@ -320,7 +321,7 @@ func TestSetStringID(t *testing.T) {
 			}
 
 			err := tt.n.SetStringID(tt.v)
-			if got, want := err, tt.err; !reflect.DeepEqual(got, want) {
+			if got, want := err, tt.err; !compareErrors(got, want) {
 				t.Fatalf("got error %q (%T) want %q (%T)", got, got, want, want)
 			}
 			// if the test should fail and the error was correct
@@ -377,7 +378,7 @@ func TestSetNamespace(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			err := tt.n.SetNamespace(tt.v)
-			if got, want := err, tt.err; !reflect.DeepEqual(got, want) {
+			if got, want := err, tt.err; !compareErrors(got, want) {
 				t.Fatalf("got error %v want %v", got, want)
 			}
 			// if the test should fail and the error was correct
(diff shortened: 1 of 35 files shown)


ua/node_id_test.go
```diff
@@ -391,3 +392,13 @@ func TestSetNamespace(t *testing.T) {
 		})
 	}
 }
+
+func compareErrors(err1, err2 error) bool {
```
magiconair, line 396: I think `reflect.DeepEqual` does exactly what `compareErrors` does. 

https://play.golang.org/p/iBDNJELixui

uapolicy/policyAes128Sha256RsaOaep.go
```diff
@@ -7,7 +7,7 @@ package uapolicy
 import (
 	"crypto"
 	"crypto/rsa"
-	"errors"
+	"github.com/gopcua/opcua/errors"
```
magiconair: maybe sort this properly

uapolicy/crypto_hmac.go
```diff
@@ -3,7 +3,7 @@ package uapolicy
 import (
 	"crypto"
 	"crypto/hmac"
-	"errors"
+	"github.com/gopcua/opcua/errors"
```
magiconair, line 6: sort pkg name

uapolicy/policyAes256Sha256RsaPss.go
```diff
@@ -7,7 +7,7 @@ package uapolicy
 import (
 	"crypto"
 	"crypto/rsa"
-	"errors"
+	"github.com/gopcua/opcua/errors"
```
magiconair: sort

uapolicy/policyBasic128Rsa15.go
```diff
@@ -7,7 +7,7 @@ package uapolicy
 import (
 	"crypto"
 	"crypto/rsa"
-	"errors"
+	"github.com/gopcua/opcua/errors"
```
magiconair: sort

uapolicy/policyBasic256.go
```diff
@@ -7,7 +7,7 @@ package uapolicy
 import (
 	"crypto"
 	"crypto/rsa"
-	"errors"
+	"github.com/gopcua/opcua/errors"
```
magiconair: sort

uapolicy/policyBasic256Sha256.go
```diff
@@ -7,7 +7,7 @@ package uapolicy
 import (
 	"crypto"
 	"crypto/rsa"
-	"errors"
+	"github.com/gopcua/opcua/errors"
```
magiconair: sort

uapolicy/securitypolicy.go
```diff
@@ -9,10 +9,10 @@ package uapolicy
 
 import (
 	"crypto/rsa"
-	"errors"
-	"fmt"
 	"sort"
 
+	"github.com/gopcua/opcua/errors"
+
```
magiconair, line 15: drop blank line

uasc/secure_channel.go
```diff
@@ -124,7 +124,7 @@ func (s *SecureChannel) SendRequest(req ua.Request, authToken *ua.NodeID, h func
 func (s *SecureChannel) SendRequestWithTimeout(req ua.Request, authToken *ua.NodeID, timeout time.Duration, h func(interface{}) error) error {
```
magiconair, line 124: this belongs to a different patch

magiconair (commented):
almost. Can you clean it up and then we merge?

ua/node_id_test.go
kung-foo: The underlying stack trace captures will be different:

https://play.golang.org/p/1WqB7u4BUf7

uasc/secure_channel.go
kung-foo: I have no idea how this is contaminating my branch. I even redid this PR from scratch.

uapolicy/securitypolicy.go
kung-foo: are you using a different fmt tool? `go fmt` doesn't report any issues this or `uapolicy/crypto_hmac.go`

ua/node_id_test.go
magiconair: Damn. That's not nice since the error itself doesn't change. It doesn't really matter where it occurs.

uapolicy/securitypolicy.go
magiconair: No, I'm just a stickler for these things. I believe in consistency and that developers don't read code but pattern match it. So any deviation from the structure slows you down b/c you fall back to the slower "read code" mode.

ua/node_id_test.go
magiconair: I've added a comment to this function referring to your play.golang.org link. My guess is that this will bite us at some point when users expect `reflect.DeepEqual` to work.

magiconair: Just got bitten by this in two ways. First, that `errors` now adds a prefix and then that the comparison did not work. I'll expose the compare function so that we can use that in tests but it would be nice if we could find a more generic solution.

Issue #239, opened 2019-07-15
Discussion on connection retries


jkissinger:
Are there plans for adding connection retries, or even existing functionality that I may have overlooked?

I have connection retries implemented at a high level, such as when first establishing the connection.  However, low level retries (for example, if the session times out or the server disconnects unexpectedly) should probably be part of the client itself.  One possibility I've considered is having a user defined error handler passed in when constructing the client and then all errors are first passed through that handler before being returned.

I'm new to Go and OPC UA so looking for any feedback or advice here.

magiconair:
I'm not aware that this is on anybody's list right now but that could be interesting. Subscriptions could be tricky since they establish state on the server (monitored items). 

For example, how would this work if you have an active subscription and the session times out or the connection gets interrupted. Would you expect the client to re-establish that subscription with all monitored items?

I think this is what the PubSub Service is trying to solve (including a robust failover).

jkissinger:
Yeah, for our use case we're definitely expecting the client to reestablish the connection and request any missed publications to be republished, similar to how it's described [here](https://www.prosysopc.com/blog/opc-ua-sessions-subscriptions-and-timeouts/):

> If communication breaks, the server can still record new notifications, assuming that it has got enough PublishRequests in advance from the client.

> In case you have 5 subscriptions and PublishRequestFactor=10, the client will send 50 PublishRequests to the server. In case of a communication break the server can send 50 PublishResponses back to the client in a normal rate. Considering a PublishingInterval of 1 s, the server can send 10 notifications for each subscription, if their items are really changing. But if the break is longer than 10 s, the server will start losing data. Unless the monitored items have a QueueSize long enough to record more samples.

> Once communication is restored, the client will first call Republish to get the missing samples: after that it can send more PublishRequests and the server can continue generating new PublishResponses.

I'm guessing from your response, that if this was added to an enhancement list it would be pretty low priority?

magiconair:
I think you misread my response then. We should add robustness features to the client and/or the library but I am not aware anybody working on this right now and I think there are some edge cases to consider. 

Contributions are more than welcome and we should focus first on adding what people need. So if you need this then I'm all for it.

I'll be dealing with the exact same problem in the coming weeks so this is probably a good time for a proposal.

jkissinger:
That's great news.  I would love to contribute personally, however we have limited resources and I'm told our priorities are elsewhere at the moment.

[...]

PR #298, opened 2019-11-04
Reconnection and publish engine implementation

unknownet:
Hello,
I've performed some modifications of your library to follow more closely the OPC standard and so I added the following:
 - OPC UA client reconnection procedure
 - Publish engine
 - Session refactorized
 - Option AutoReconnect added
 - Secure channel run context moved from client to secure channel
 - Subscription publish mechanism removed, now implemented in publish
engine
 - Debug message added

I followed this diagram, that you can find in the [lastest OPC UA specification](https://opcfoundation.org/developer-tools/specifications-unified-architecture) __(Part 4 - fig36 p112)__.
 
reconnection_patern

__!! FYI !!__
This implementation could be better with a more robust event handling system but I didn't wanted to alter too much the existing architecture.
I would suggest something similar to [this](https://github.com/unknownet/notify/blob/develop/notification.go) as a notification mechanism.

### DIFF
--- a/client.go
+++ b/client.go
@@ -14,6 +14,7 @@ import (
 	"sort"
 	"sync"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	"github.com/gopcua/opcua/debug"
@@ -80,6 +81,18 @@ func (a bySecurityLevel) Len() int           { return len(a) }
 func (a bySecurityLevel) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
 func (a bySecurityLevel) Less(i, j int) bool { return a[i].SecurityLevel < a[j].SecurityLevel }
 
+type ClientNotification uint
+
+const (
+	ClientNotificationStartReconnection ClientNotification = iota
+	ClientNotificationReconnection
+	ClientNotificationReconnectionCancel
+	ClientNotificationReconnectionAttemptHasFailed
+	ClientNotificationAfterReconnection
+	ClientNotificationConnectionReestablished
+	ClientNotificationConnectionLost
+)
+
 // Client is a high-level client for an OPC/UA server.
 // It establishes a secure channel and a session.
 type Client struct {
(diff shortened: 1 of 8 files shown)


magiconair:
Thank you. I will have a look. Is it necessary to have all these changes in a single commit or would it be possible to make a set of smaller changes?

Out of curiosity: what motivated this change? Is this fixing an existing problem or cleanup?

magiconair:
I had a look at this and I think this PR is too complex to just get merged. I think it is also refactoring too many things at the same time which makes it hard to impossible to review. 

Could you please split this up into smaller commits and/or PRs and first open an issue on what you are intending to change, why this needs fixing and what your approach is? It does not have to be an epic but should provide some context. 

Then we have a better baseline for reviewing the change.

unknownet:
I understand this modification is a bit intrusive and I'm sorry about that...

My first motivation was to create an input plugin for [Telegraf](https://www.influxdata.com/time-series-platform/telegraf/).
Your package works fine out of the box for collecting data. However, I needed to be able to maintain the connection for serveral months to a Siemens Sinumerik OPC server that is cutting the connection (as intended in the OPC standard). As I obviously didn't wanted to set a lifetime of 10 years and not loose data each time the connection failed, I had to implement all the standard reconnection process. Moreover, the Siemens Sinumerik OPC server in use is intended to be shutdown and/or rebooted at random time interval and I wanted to be able to reconnect each time automaticaly (again without missing data).

To answer your question for smaller changes per commit/pull request, I've already squashed the temporary branch and in doing so deleted it. To be honnest, at the beggining, I only wanted to implement the reconnection mechanism, but to do it correcly I had to pile up other changes sush as the publish engine...

I could remove the session keep alive and the debug messages, but I doubt it would make it easier to understand, if you have any suggestion as how I could make it easier I'll be happy to do it.

magiconair:
Ok. We will have a look then. Anybody else from the community want to lend a hand? Let’s try to do this properly. @unknownet I assume you’re open to comments and changes.

magiconair:
I can probably do a first sweep today or tomorrow.

kung-foo:
I am interested in reviewing. But I agree with @magiconair, if we can break this into multiple PRs then it becomes a lot easier to review.

magiconair:
Building reviewable PRs is work, I understand that. One thing to consider @unknownet is that if you don't invest the time to provide us with a change that is easy to review then you somewhat expect us to do that for you.

unknownet:
Yes I am fully aware of your last point, I am trying split into multiple PRs while I am writing, and to come up with a suggestion of how to do it, in a way that could optimize both our time.
At the time same time keep a functional state with each PR (obvisously).

As a first uncomplete draft, I would start by implementing the backbone of the reconnection mechanism (the bare minimum) and the session reactivation.
reconnection-reactivate_patern

I haven't figure the rest yet, does it sound more feasable ?

magiconair:
I think that makes more sense. Refactoring in a separate commit before you add the feature is a good strategy.

Also, please don't get me wrong. I think this will be definitely a useful feature and you're also not the first one asking for or attempting to write it. We'll help you to get this over the line.

magiconair:
See #239

unknownet:
Thanks you for your support,
I've already seen the issue #239 and this is what motivated me to do this PR.

Btw the first batch of features has been done in PR #299.

PR #301, opened 2019-11-11
do not panic on arrays dimensions greater than zero

### LINKED ISSUE #300: Variant.String() panics if underlying type is []string
Using the convenience function `String()` when the underlying go type is not a literal `string` will result in a panic.

```go
func TestVariant(t *testing.T) {
	v := ua.MustVariant([]string{"a", "b", "c"})
	println(v.String())
}
```

```
--- FAIL: TestVariant (0.00s)
panic: interface conversion: interface {} is []string, not string [recovered]
	panic: interface conversion: interface {} is []string, not string
```

This is because we only check the opc type before casting and do not check array length or dimension.

At minimum we should change the convenience functions to return `($type, error)`. But I also think we should revisit the whole `Variant` interface since I am finding it a bit awkward to work with.

kung-foo:
fixes #300 
note: this is a short term fix until a better interface can be designed.

### DIFF
--- a/ua/variant.go
+++ b/ua/variant.go
@@ -458,6 +458,10 @@ func (m *Variant) set(v interface{}) error {
 // todo(fs): this should probably be StringValue or we need to handle all types
 // todo(fs): and recursion
 func (m *Variant) String() string {
+	if m.ArrayLength() > 0 {
+		return ""
+	}
+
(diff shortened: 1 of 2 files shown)


magiconair (approved):
LGTM

PR #302, opened 2019-11-14
Feature/reconnect recreate

unknownet:
Hello,

In addition to the previous PL #299, this one add the possibilty to dynamically recreate the session and/or the subscription.
reconnection_patern_recreate
I plan to implement two additional PR:
 - Keep alive _(send a heartbeat if no value come from subscriptions)_
 - Publish Pool _(multiple pending publish request to allow the server to respond in burst mode)_

PR #299, opened 2019-11-07
Reconnection and session reactivation

unknownet:
Hello,
As agreed in my previous PR #298, this is a first batch of features to handle reconnection and session reactivation after the secure channel has disconnected.
I've made significant changes from PR #298 to minimize intrusive code and to follow your choice of architecture.

This Pull Request implements the following scheme:

image

 - Configuration AutoReconnect added
 - AutoReconnect state machine implemented for session reactivation
 - Repair subscriptions (republish) implemented
 - Suspend and Resume subscription implemented

I Hope that this Pull Request would be easier to digest :D and of course feel free to modify it (I can adjust the following PRs) !

PR #304, opened 2020-01-02
feat: automatically renew security token of secure channel before its lifetime expires

### LINKED ISSUE #303: Client should reconnect on expiry of secure channel lifetime
Client configuration includes the option for requested secure channel lifetime, which is sent to the server in the OpenSecureChannelRequest. The client is supposed to open a new secure channel after the lifetime of the current one expires. As it stands right now, this is left to the user of API to deal with.  However, there's even no way yet for the user to learn what's the revised lifetime from the server was, because it's never read from OpenSecureChannelResponse.

I'd like the client to automatically open a new secure channel when the lifetime of the existing one expires.

EDITED TO ADD:
I shouldn't have spoken of opening a secure channel above, but rather of the necessity to renew the security token of the secure channel before the token expires.

alexbrdn:
- store the security token lifetime from OpenSecureChannelResponse in Config
- schedule renewing the token before it expires

closes #303

### DIFF
--- a/client.go
+++ b/client.go
@@ -92,6 +92,9 @@ type Client struct {
 	// sessionCfg is the configuration for the session.
 	sessionCfg *uasc.SessionConfig
 
+	// conn is the open connection
+	conn *uacp.Conn
+
 	// sechan is the open secure channel.
 	sechan *uasc.SecureChannel
 
@@ -165,27 +168,33 @@ func (c *Client) Dial(ctx context.Context) error {
 	if c.sechan != nil {
 		return errors.Errorf("secure channel already connected")
 	}
-	conn, err := uacp.Dial(ctx, c.endpointURL)
+	var err error
+	c.conn, err = uacp.Dial(ctx, c.endpointURL)
 	if err != nil {
 		return err
 	}
-	sechan, err := uasc.NewSecureChannel(c.endpointURL, conn, c.cfg)
+	c.sechan, err = uasc.NewSecureChannel(c.endpointURL, c.conn, c.cfg)
 	if err != nil {
-		_ = conn.Close()
+		_ = c.conn.Close()
 		return err
 	}
-	c.sechan = sechan
 	ctx, c.cancelMonitor = context.WithCancel(ctx)
 	go c.monitorChannel(ctx)
+	err = c.openSecureChannel(ctx, c.sechan.Open)
+	if err != nil {
+		return err
+	}
+	return nil
+}
 
-	if err := sechan.Open(); err != nil {
+func (c *Client) openSecureChannel(ctx context.Context, open func() error) error {
+	if err := open(); err != nil {
 		c.cancelMonitor()
-
-		_ = conn.Close()
+		_ = c.conn.Close()
 		c.sechan = nil
 		return err
 	}
-
+	c.scheduleRenewingToken(ctx)
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -708,6 +717,20 @@ func (c *Client) HistoryReadRawModified(nodes []*ua.HistoryReadValueID, details
...
+func (c *Client) scheduleRenewingToken(ctx context.Context) {
+	timer := time.NewTimer(time.Duration(0.75*float64(c.cfg.Lifetime)) * time.Millisecond) // 0.75 is from Part 4, Section 5.5.2.1
+
+	go func() {
+		select {
+		case <-ctx.Done():
+			timer.Stop()
+		case <-timer.C:
+			debug.Printf("renewing security token...")
+			c.openSecureChannel(ctx, c.sechan.Renew)
```
golangcibot: Error return value of `c.openSecureChannel` is not checked (from `errcheck`)

kung-foo:
LGTM
Is there any overlap and/or mismatch with the proposals in https://github.com/gopcua/opcua/pull/302 or https://github.com/gopcua/opcua/pull/299?

alexbrdn:
The renewal of token is necessary when the connection is still alive, but the token is about to expire. Those other PRs deal with reconnecting when the connection is cut and republishing missed changes to subscriptions. In the long run (pun intended), both are necessary.

kung-foo:
Cool. I tested the PR against the python and nodejs servers as well as kepware (forcing a 5 second lifetime). No issues with maintaining subscriptions.

unknownet:
I Think that I have encountered an issue directly caused by the automatic session renewal.
I have a client with 6 subscriptions and about 40 monitored item (total),
and after the session security token was renewed, I get the following errors:

```console
2020-02-04T12:13:20Z E! [inputs.opcua] Notification error: The message encoding/decoding limits imposed by the stack have been exceeded. StatusBadEncodingLimitsExceeded (0x80080000)
```

I suspect the following explanation lead to this previous error:

1. A new SecureChannel is created in the server (renew), when the previous one was at 75% of his Lifetime.
2. We no longer read from the first SecureChannel, since it's replaced by the second one in the monitorChannel function.

The norme state:

Part 4, 5.5.2.1:
[...] Each SecureChannel exists until its explicitly closed or until the last token has expired and the overlap period has elapsed. [...]
[...] Clients should accept Messages secured by an expired SecurityToken for up to 20% of the token lifetime [...]

So I suppose the former SecureChannel keeps getting notification from the server and never get flushed, wich result to a StatusBadEncodingLimitsExceeded.

alexbrdn:
There is no new SecureChannel created when the token is renewed. The channel only gets a refreshed security token from the server.

unknownet:
Very true and thank you btw for your reactivity !
I suppose I mixed up the SecureChannel and the SecurityToken.

The reason why I came up with my silly asumption is because I've run into a renew error with my setup:

```toml
security_policy = "Basic256Sha256"
security_mode = "SignAndEncrypt"
auth_mode = "UserName"
```

The error returned by the server is __"BadEncodingLimitExceeded"__,

And that's what append with the request (Wireshark renew request packet):
image

Whereas with the Ua expert Client and the same configuration I get this (the request work in this case):
image

Hope this could help !

unknownet:
I think I might have found to reason why this error occurs,
By the way if you prefer I could open an issue rather then writing on the pull request.

https://github.com/gopcua/opcua/blob/9ba7e84e96413982af1a56edff6f784c2bea75ac/uasc/secure_channel_crypto.go#L21-L24

In the function SignAndEcrypt the variable __"isAsymmetric"__ is based on the status of the SecureChannel connection, when there was no renew request involved, __isAsymmetric__ could be only for the first __"openChannelRequest"__ which was the first request after the SecureChannel was created, but it is not true anymore hence the __SecurityPolicyUri__ being Encrypted, and so causing the Malformed Packet.

I'am going to test it in the days to come I juste wanted to share this If it could give you a head start in case you were willing to investigate this before I do.

alexbrdn:
Thanks for looking into this problem. I do think it's worth it to create an extra issue for this and include the above information in it. And whatever is necessary to reproduce it.

Of course, if you have a fix for it by now, no one will complain if you create a PR.

kung-foo:
I agree with the new issue. I (think) I've just bumped into this on a Kepware box rejecting openChannelRequest messages during renewal.

kung-foo:
I'll open a new one.

PR #308, opened 2020-01-10
chore: move connection state from Config to SecureChannel

alexbrdn:
As [discussed](https://github.com/gopcua/opcua/pull/299#issuecomment-572923643), this moves connection state fields to unexported fields in `SecureChannel`

PR #312, opened 2020-01-22
Feature/reconnect recreate

unknownet:
Hello,

Same as PL #302 - rebased from latested version of PL #299
 
This one add the possibilty to dynamically recreate the session and/or the subscription.

image

~~<s> Keep alive (send a heartbeat if no value come from subscriptions) </s>~~ No longer needed since PL #308
~~<s> Publish Pool (multiple pending publish request to allow the server to respond in burst mode) </s>~~ Misunderstanding publish pool isn't meant for burst server reponse

> Spec 1.03 part 4 5.13.5 Publish
 [..] in high latency networks, the Client may wish to pipeline Publish requests to ensure cyclic reporting from the Server. Pipe-lining involves sending more than one Publish request for each Subscription before receiving a response. For example, if the network introduces a delay between the Client and the Server of 5 seconds and the publishing interval for a Subscription is one second, then the Client will have to issue Publish requests every second instead of waiting for a response to be received before sending the next request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ The current set of supported services is only for the high-level client.
 |                             | ModifySubscription            |           |              |
 |                             | SetPublishingMode             |           |              |
 |                             | Publish                       | Yes       |              |
-|                             | Republish                     |           |              |
+|                             | Republish                     | Yes       |              |
 |                             | DeleteSubscriptions           | Yes       |              |
 |                             | TransferSubscriptions         |           |              |
 
--- a/client.go
+++ b/client.go
@@ -157,7 +157,7 @@ func (c *Client) Connect(ctx context.Context) (err error) {
 }
 
 func (c *Client) sessionClosed() bool {
-	return c.Session() != nil
+	return c.Session() == nil
 }
 
 // Dial establishes a secure channel.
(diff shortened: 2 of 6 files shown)


AndreySerebrovC2M:
@magiconair, could you please share your plans about this PR? Do you need some help to test it? Or you think it's not a good idea?

magiconair:
Didn't have a chance to look at it yet. Last time I looked at the previous attempts they were making too many changes and it looked like it would make sense to have a state machine for the entire connection handling and not just the re-connect logic. 

I'll have a look but not this week (vacation).

ymmemo:
Hi everyone, I have encountered the same question about losing data due to the delay. I would like to know if the newly updated version will contain ‘pipeline Publish’ feature.

unknownet:
> Hi everyone, I have encountered the same question about losing data due to the delay. I would like to know if the newly updated version will contain ‘pipeline Publish’ feature.

Hi, yymemo, I understand the need, but the feature you  mentioned is not part of this PR, in fact this PR entirely rely on the PR #299 and make no sens on its own.
Perhaps it would be better to close it until further progress on PR #299 to avoid any confusion.

ymmemo:
> Hi, yymemo, I understand the need, but the feature you mentioned is not part of this PR, in fact this PR entirely rely on the PR #299 and make no sens on its own.
> Perhaps it would be better to close it until further progress on PR #299 to avoid any confusion.

Ok, Thanks for your reply^^, I have understood the subject.

PR #314, opened 2020-01-22
Decouple dial context from monitor context

### LINKED ISSUE #313: Bug: Connection closes unexpectedly when dialing with timeout context
Assume the following code:

```go
ctx, cancel := context.WithTimeout(context.Background(), 5 * time.Second)
defer cancel()

c := NewClient("localhost:4840")
c, err := c.Dial(ctx)
if err != nil {...}
c.Whatever...
# connection stalls after 5 seconds and gets eventually terminated by the server
```

What happens is that the `monitor` go routine uses the dial context as parent context and since it is a timeout context it fires eventually. However, by that time you are not supposed to use it anymore. Since only the monitor routine dies the application continues except that it no longer forwards packets it receives to their handlers and the app stalls.

This was a fun chase .... :)

magiconair:
This patch decouples the context for the monitor go routine which
dispatches incoming messages to handler functions from the dial context.

The coupling had the effect that the monitor go routine terminates when
the dial context is closed. This can happen even after a successful
connection when a timeout context is used to implement a dial timeout.
The timeout context fires at some later point when it is no longer used
which then terminates the monitor routine and the application hangs.

Fixes #313

### DIFF
--- a/client.go
+++ b/client.go
@@ -178,13 +178,15 @@ func (c *Client) Dial(ctx context.Context) error {
 		_ = c.conn.Close()
 		return err
 	}
-	ctx, c.cancelMonitor = context.WithCancel(ctx)
-	go c.monitorChannel(ctx)
-	err = c.openSecureChannel(ctx, c.sechan.Open)
-	if err != nil {
-		return err
-	}
-	return nil
+
+	// Issue #313: decouple the dial context from the monitor context
+	// mctx must *not* be a child context of 'ctx'. Otherwise, the
+	// monitor go routine terminates whenever the dial context is done
+	// which may get triggered unexpectedly by a timer context.
+	var mctx context.Context
+	mctx, c.cancelMonitor = context.WithCancel(context.Background())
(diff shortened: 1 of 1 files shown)


magiconair:
Could you @alexbrdn and/or @kung-foo have a look whether there is a reason not to do it like this?

Thank you

alexbrdn:
I wonder a little why would you pass a timeout context to Dial (or Connect) if you didn't want the connection to die when the context expires? Other than that, I have no objections.

kung-foo:
I've always assumed the dial context only controls the time to connect. See https://golang.org/pkg/net/#Dialer.DialContext

alexbrdn:
It looks very much like that it's my assumption about context use in this case that was wrong.

magiconair:
The dial context should indeed only control the time to connect but it was also controlling the application message pump. 

I’ve added this to our code just to be robust and not because we had an issue but then everything went south and it took a while to figure out why.

PR #317, opened 2020-02-11
Add option to set the session name

magiconair:


magiconair:
Is there anything to keep in mind when setting the session name? Do they have to be unique every time the client connects?

uasc/config.go
```diff
@@ -114,6 +114,10 @@ type SessionConfig struct {
 	// The SignatureData type is defined in 7.32.
 	UserTokenSignature *ua.SignatureData
 
+	// SessionName is an optional name of the session.
+	// The default is a random value.
```
kung-foo: `UnixNano` is not random

magiconair: No, it isn't. Clarified the documentation. Question is whether it has to be.

PR #321, opened 2020-02-18
refactor/tweak SubscriptionParameters usage

### LINKED ISSUE #320: StatusBadSequenceNumberUnknown when subscribing to a node in Kepware
Using the default configuration to subscribe/monitor a node in Kepware (KEPServerEX.V6) results in:

`2020/02/18 09:29:58 error: sub=80 err=The sequence number is unknown to the server. StatusBadSequenceNumberUnknown (0x807A0000)`

When trying the same subscription using `examples/subscribe` it works. The difference is that in the `subscribe` example, it sets `SubscriptionParameters.Interval` to 500ms vs the default 100ms.

The monitor feature doesn't expose the interval parameter and instead leaves it at 100ms.

So for Kepware, 100ms fails, 500ms works. :man_shrugging:

kung-foo:
* Exposes `opcua.SubscriptionParameters` to the monitor function
* Separates notification channel from `SubscriptionParameters`

Fixes #320

### DIFF
--- a/monitor/subscription.go
+++ b/monitor/subscription.go
@@ -69,7 +69,11 @@ func NewNodeMonitor(client *opcua.Client) (*NodeMonitor, error) {
 	return m, nil
 }
 
-func newSubscription(m *NodeMonitor, notifyChanLength int, nodes ...string) (*Subscription, error) {
+func newSubscription(m *NodeMonitor, params *opcua.SubscriptionParameters, notifyChanLength int, nodes ...string) (*Subscription, error) {
+	if params == nil {
+		params = &opcua.SubscriptionParameters{}
+	}
+
 	s := &Subscription{
 		monitor:          m,
 		closed:           make(chan struct{}),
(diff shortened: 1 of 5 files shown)


monitor/subscription.go
```diff
@@ -69,7 +69,11 @@ func NewNodeMonitor(client *opcua.Client) (*NodeMonitor, error) {
 	return m, nil
 }
 
-func newSubscription(m *NodeMonitor, notifyChanLength int, nodes ...string) (*Subscription, error) {
+func newSubscription(m *NodeMonitor, params *opcua.SubscriptionParameters, notifyChanLength int, nodes ...string) (*Subscription, error) {
```
magiconair, line 72: Was wondering what `notifyChanLength` does here until I noticed it was there before. Looks a bit strange but we're looking at this from a different angle anyway.

magiconair (commented):
LGTM

monitor/subscription.go
kung-foo: It is the size of an internal channel used for buffering results that will be sent out to the receiver.

The reason that it is parameterized (though still private), is that there are two users: `NodeMonitor.Subscribe` and `NodeMonitor.ChanSubscribe`.

In the first one (callback-based), we the library take responsibility for buffering data-change messages that are sent out via callback. Hence the larger (8192) size.

In the later, we assume the client will take responsibility for draining their user-provided channel fast enough. Therefore the internal channel only needs to be big enough to handle moving messages between underlying source channel and the user-provided channel.

PR #323, opened 2020-02-18
Fix GUID implementation

### LINKED ISSUE #322: guid encoding / decoding might be wrong
Hey,

this is kind of related to an already closed issue https://github.com/gopcua/opcua/issues/272. Sorry for bringing this topic up again but I think the guid implementation is wrong.

Here are the relevant parts from the spec 

https://reference.opcfoundation.org/v104/Core/docs/Part6/5.1.3/
https://reference.opcfoundation.org/v104/Core/docs/Part6/5.2.2/#5.2.2.6

As far as I understand `Data4` is just a byte slice. It's not a `uint64`. That matters because of little / big endian.

The example from the spec says

`“72962B91-FA75-4AE6-8D28-B404DC7DAF63”` is 

```go
[]byte{
  // data1 (inverse order)
  0x91, 0x2b, 0x96, 0x72,
  // data2 (inverse order)
  0x75, 0xfa,
  // data3 (inverse order)
  0xe6, 0x4a,
  // data4 (same order)
  0x8d, 0x28, 0xb4, 0x04, 0xdc, 0x7d, 0xaf, 0x63
}
```

As you can see for `data1`, `data2`, `data3` endianness matters. The data4 bytes are simply attached to the previous bytes without thinking about endianness.

Our test case `"AAAABBBB-CCDD-EEFF-0101-0123456789AB"` https://github.com/gopcua/opcua/blob/fe267a1e2ac56111d50460e8e107718ac1e33d1c/ua/datatypes_test.go#L93 should be changed to

```go
[]byte{
  0xbb, 0xbb, 0xaa, 0xaa, 0xdd, 0xcc, 0xff, 0xee,
  0x01, 0x01, 0x01, 0x23, 0x45, 0x67, 0x89, 0xab
}
```

Then it fails and we have to change the encoding / decoding.

I used your test for my JavaScript implementation and the stuff from the spec as well. No matter what I did one of them was always failing. That was pretty frustrating :smile: So one is wrong at the moment. It's either the spec or the current implementation.

Best regards,
Mirco

magiconair:
Fixes #322


zemirco:
lgtm :+1: 

Am I allowed to merge? Thank you for inviting me to the gopcua organization \o/.

magiconair:
Hit it.

PR #326, opened 2020-03-02
WIP: rework secure channel logic

### LINKED ISSUE #324: Unable to renew security token when security mode is not none
Note: see previous discussion in #304 

On Kepware with security mode set to `Sign` and policy to `http://opcfoundation.org/UA/SecurityPolicy#Basic128Rsa15`, when the renewal is triggered, I see the outgoing `OpenSecureChannelRequest`:

image

followed immediately by:

image

kung-foo:
> Note: WIP

The goal of this PR is twofold:
1) enable proper secure channel renewal
2) cleanup broken/unused/racy code paths in the secure channel implementation

The biggest change is that a `SecureChannel` can now handle multiple secure channel IDs and token IDs. This is because the spec requires that existing crypto state be available _after_ the renewal process has completed. This is ensure that network delays, retransmission and other timing issues do not break existing messages. So we must keep them around, and optionally try to use them if decryption fails.

> Note: we only have access to the `secureChannelID` and not the `secureTokenID` before decryption. This seems to indicate we need to work our way backwards through the crypto state 

```
C: OPN() -> SC(c=1, t=1)      # new secure channel with id=1 and token_id=1
S: MSG1(c=1, t=1)             # message encrypted with id=1 and token_id=1
S: MSG2(c=1, t=1)
S: MSG3(c=1, t=1)
C: OPN(renew) -> SC(c=1, t=2) # renewal. new token_id=2
S: MSG4(c=1, t=1)             # message encrypted with old token due to network latency
S: MSG5(c=1, t=2)
```

## Changes

* All "crypto state" (keys, nonces, coutners, etc) are moved into a `channelState`struct
* Move `monitorChannel` functionality out of `Client` and into `SecureChannel` and rename it to `dispatch` since that better matches what it actually does
* Remove broken public `sc.Receive` functionality since we are using a dispatch style of multiplexing (no one appeared to be using it)
* Remove spawning a separate go routine just to deliver a single struct to a chan. (instead make the `response` chan with length 1)
* Added some message stats to `channelState`
* two scheduled functions: `scheduleRenewal` to renew a token, and `scheduleExpiration` to age out state
* attempt to reduce request/response-type-specific code that was littering the encoding/decoding paths.
* `SecureChannel` no longer closes the underlying `Client`
* lots of `go run --race ...` cleanup

## TODO

* Review `sync.Lock` usage -- Current code has separate locks for each struct level var. There is probably some that can be collapsed to a shared lock. Also the `channelState` is lockable which may or may not be useful.
* Ideally rather than pass in a concrete `Client` we could create an interface. This would facilitate testing by enabling a mock client.
* ~~Tests are completely broken~~
* Review the `RequestHandle` field of a `RequestHeader`. I am currently using the `requestID`since they seem to always be the same. Not sure it that is correct.


kung-foo:
FYI, this PR is not dead :)
I've been testing it in the real world to see how it behaves. Looking good so far.

magiconair:
Let me know when you're happy with this so that we can review it.

uasc/secure_channel_instance.go
```diff
@@ -0,0 +1,220 @@
...
+	"github.com/gopcua/opcua/ua"
+	"github.com/gopcua/opcua/uapolicy"
+)
+
+type instanceState int
+
+const (
+	channelOpening instanceState = iota
+	channelActive
+	channelClosed
```
golangcibot: `channelClosed` is unused (from `deadcode`)

uasc/secure_channel.go
```diff
@@ -433,250 +344,82 @@ func (s *SecureChannel) readChunk() (*MessageChunk, error) {
...
-			b, err := mergeChunks(all)
-			if err != nil {
-				return reqid, nil, errors.Errorf("chunk merge error: %v", err)
-			}
-
-			if uint32(len(b)) > s.c.MaxMessageSize() {
-				return reqid, nil, errors.Errorf("message too large: %d > %d", uint32(len(b)), s.c.MaxMessageSize())
-			}
+	instances := s.getInstancesBySecureChannelID(m.MessageHeader.SecureChannelID)
+	if instances == nil || len(instances) == 0 {
```
golangcibot: S1009: should omit nil check; len() for nil slices is defined as zero (from `gosimple`)

uasc/secure_channel.go
```diff
@@ -692,119 +435,336 @@ func (s *SecureChannel) openSecureChannel(requestType ua.SecurityTokenRequestTyp
...
+func (s *SecureChannel) scheduleExpiration(instance *channelInstance) {
+	// https://reference.opcfoundation.org/v104/Core/docs/Part4/5.5.2/#5.5.2.1
+	// Clients should accept Messages secured by an expired SecurityToken for up to 25 % of the token lifetime.
+	const expireAfter = 1.25
+	when := instance.createdAt.Add(time.Second * time.Duration(instance.revisedLifetime.Seconds()*expireAfter))
+
+	debug.Printf("channelID %d/%d will expire at %s", instance.secureChannelID, instance.securityTokenID, when.UTC().Format(time.RFC3339))
 
-	var err error
+	t := time.NewTimer(when.Sub(time.Now()))
```
golangcibot: S1024: should use time.Until instead of t.Sub(time.Now()) (from `gosimple`)

kung-foo:
I've tested this against a production Kepware instance for the last two weeks. In my case, I have a continuous subscription running, and a relatively short lifetime of 60 minutes. No issues, and I have verified that the secure channel does indeed need to use the previous crypto material when the renewal happens concurrent to a subscription notification.

I've also tested against a node-js test server, which I may create a separate PR so we can start thinking about more complete tests.

kung-foo:
I'll squash and re-write once everyone is happy.

kung-foo:
I've triggered an invalid renewal state with Kepware. Need to investigate.

uasc/message_test.go
```diff
@@ -22,11 +22,13 @@ func TestMessage(t *testing.T) {
 					cfg: &Config{
 						SecurityPolicyURI: "http://gopcua.example/OPCUA/SecurityPolicy#Foo",
 					},
-					requestID:       1,
-					sequenceNumber:  1,
+				}
+				instance := &channelInstance{
+					sc: s,
+					sequenceNumber:  0,
```
magiconair, line 28: has the logic changed for getting the sequence numbers so that we can start at 0 now? I'm all for it, just curious.

uasc/secure_channel.go
```diff
@@ -1,12 +1,7 @@
-// Copyright 2018-2020 opcua authors. All rights reserved.
-// Use of this source code is governed by a MIT-style license that can be
-// found in the LICENSE file.
-
 package uasc
```
magiconair: Missing copyright header.

uasc/secure_channel.go
```diff
@@ -114,258 +94,202 @@ func NewSecureChannel(endpoint string, c *uacp.Conn, cfg *Config) (*SecureChanne
...
+	return s, nil
 }
 
-func (s *SecureChannel) setState(n int32) {
-	atomic.StoreInt32(&s.state, n)
+func (s *SecureChannel) reset() {
+	s.closingMu.Lock()
+	defer s.closingMu.Unlock()
+
+	// note: we _don't_ reset s.requestID
```
magiconair, line 122: why? If you've already explained it maybe point to that comment.

uasc/secure_channel.go
```diff
@@ -114,258 +94,202 @@ func NewSecureChannel(endpoint string, c *uacp.Conn, cfg *Config) (*SecureChanne
...
+	s.reset()
 
-func (s *SecureChannel) Lifetime() uint32 {
-	return s.lifetime
+	return s, nil
 }
 
-func (s *SecureChannel) setState(n int32) {
-	atomic.StoreInt32(&s.state, n)
+func (s *SecureChannel) reset() {
```
magiconair, line 118: Why do we re-use a secure channel instance? I'm sure there is a reason but this deviates from the way normal Go programs function so it would be good to explain why.

uasc/secure_channel.go
```diff
@@ -114,258 +94,202 @@ func NewSecureChannel(endpoint string, c *uacp.Conn, cfg *Config) (*SecureChanne
...
+			if !ok {
+				debug.Printf("uasc %d/%d: no handler for %T", s.c.ID(), resp.ReqID, resp.V)
+				continue
+			}
+
+			debug.Printf("sending %T to handler\n", resp.V)
+			select {
+			case ch <- resp:
+			default:
+				// this should never happen since the chan is of size one
```
magiconair, line 175: Maybe log it so that we see it if does?

uasc/secure_channel.go
```diff
@@ -114,258 +94,202 @@ func NewSecureChannel(endpoint string, c *uacp.Conn, cfg *Config) (*SecureChanne
...
+			return resp
+		}
+	}
 }
 
 func (s *SecureChannel) readChunk() (*MessageChunk, error) {
 	// read a full message from the underlying conn.
 	b, err := s.c.Receive()
-	if err == io.EOF || s.hasState(secureChannelClosed) {
+	if err == io.EOF || len(b) == 0 { // || s.hasState(secureChannelClosed) {
```
magiconair: is the state check still relevant?

uasc/secure_channel.go
```diff
@@ -114,258 +94,202 @@ func NewSecureChannel(endpoint string, c *uacp.Conn, cfg *Config) (*SecureChanne
...
 		return nil, io.EOF
 	}
-	if errf, ok := err.(*uacp.Error); ok {
-		return nil, errf
-	}
+
```
magiconair, line 292: I recall adding that check to make the error handling between uacp and uasc somewhat simpler. I can try to dig up the issue that lead to that. Any particular reason for removing this?

uasc/secure_channel.go
```diff
@@ -114,258 +94,202 @@ func NewSecureChannel(endpoint string, c *uacp.Conn, cfg *Config) (*SecureChanne
...
 	if err != nil {
 		return nil, errors.Errorf("sechan: read header failed: %s %#v", err, err)
 	}
 
-	const hdrlen = 12
+	const hdrlen = 12 // TODO: move to pkg level const
```
magiconair, line 297: I think the reason I put this here is because I wanted a constant to have a name for a magic number but that number was only ever going to be used in this tiny context.

magiconair (commented):
This is quite an extensive change to the way we're handling requests. I got half way through so far. Need to read this a couple more times. 

Not sure if having a single commit for this makes it easier to review. However, I'm also not sure if you can refactor this incrementally.

I think what would help in any case is a longer description on how the secure channel implementation operates with references to the methods being used. That should probably be the `SecureChannel` type documentation. When the doc points to an internal method we should make sure it is named accordingly. Clarity is important.

Thank you so far. I'll continue a bit later.

Update: it isn't a single commit. my bad

uasc/secure_channel.go
kung-foo: It just seemed redundant since L289 returns the error anyway.

uasc/secure_channel.go
kung-foo: probably not. was left over from the refactor.

client.go
```diff
@@ -168,70 +164,33 @@ func (c *Client) Dial(ctx context.Context) error {
...
+
 	// try to close the session but ignore any error
 	// so that we close the underlying channel and connection.
 	_ = c.CloseSession()
-	if c.cancelMonitor != nil {
-		c.cancelMonitor()
-	}
 
-	return c.sechan.Close()
+	 _ = c.sechan.Close()
```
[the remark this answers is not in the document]
kung-foo: yeah, i'm torn as well. technically I should maybe wrap or append the first error. But no caller would ever check that.

Logging it is possible, but I think we should revisit logging and make it "first class" rather than the `debug.Log` method we have now.

uacp/conn.go
```diff
@@ -319,6 +319,8 @@ const hdrlen = 8
 // The size of b must be at least ReceiveBufSize. Otherwise,
 // the function returns an error.
 func (c *Conn) Receive() ([]byte, error) {
+	// TODO(kung-foo): allow user-specified buffer
+	// TODO(kung-foo): sync.Pool
```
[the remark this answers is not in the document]
kung-foo: Good to know. I haven't really profiled heap usage in our product, but it seemed like a good place to potentially optimize.

uapolicy/securitypolicy.go
```diff
@@ -139,6 +141,17 @@ func (e *EncryptionAlgorithm) NonceLength() int {
 	return e.nonceLength
 }
 
+func (e *EncryptionAlgorithm) MakeNonce() ([]byte, error) {
+	b := make([]byte, e.NonceLength())
+	// note: we use `rand.Reader` instead of `rand.Read(...)` to ensure that we don't accidentally switch to using
+	// math/rand (which has a default, fixed seed). Only crypto/rand exposes a global `io.Reader` var.
```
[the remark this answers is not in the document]
kung-foo: No test is needed because if the import is changed to `math/rand` there will be a compile time error.

uasc/secure_channel.go
kung-foo: oops.

uasc/message_test.go
kung-foo: There's a atomic get/increment which returns the _new_ value. hence starts with 1.

uasc/secure_channel.go
kung-foo: will do.

uasc/secure_channel.go
kung-foo: IIRC I came across the same value in another file/func.

uasc/secure_channel.go
kung-foo: I think part of the issue is that we expose `secchan.Open` as an instance function rather than a static. This means the API allows use to call `s.Open()`, `s.Close()` and then `s.Open()` again.

Not the best API. Maybe worth removing/changing.

unknownet:
I've tested your PR on a Siemens Sinumerik 840D server.
And I got the following response from the server: `BadRequestTypeInvalid`.
To fix it I found out I had to add the following line to `uasc/secure_channel.go`.

```diff
@@ -452,6 +461,7 @@ func (s *SecureChannel) open(ctx context.Context, instance 		
		// this seems racy. if another request goes out while the other open request is in flight then won't an error
		// be raised on the server? can the sequenceNumber be as "global" as the request ID?
		s.openingInstance.sequenceNumber = instance.sequenceNumber
+		s.openingInstance.secureChannelID = instance.secureChannelID
	}
```

But with that fix some asynchronous problems still remain, after the renewal of the TokenID, the server won't accept a new request with the old TokenID even though It hasn't expired.
Also from the moment the server replied to the request `open renew` the following `MSG`s will be encrypted with the newly created secure channel and sometime even before than the client has the time to save the new `SecureChannelInstance` as the `activeInstance`.

I came up with fix [patch/issue-324-sinumerik](https://github.com/unknownet/opcua/tree/patch/issue-324-sinumerik), but I am not satisfied with It, but It gets the job done.
Hopefully It will help you, I am sure you would find a more elegant approach.

kung-foo:
Hmm, my interpretation was that the server would continue to accept unexpired tokens, allowing a client to _not_ have to do the stop-the-world lock as in your patch. But I've only tested with NodeOPCUA and Kepware. I'll fire up my node test environment again and see if I can explicitly test messages with an older token.

magiconair:
Might be quicker if you rebase and squash this branch so that we can use it as a basis for the reconnect logic. Do you have time for that today @kung-foo ?

kung-foo:
Yeah, I should have time.

magiconair:
Doesn't look rebased yet

kung-foo:
Nope. I don't think my git skills are sufficient.

magiconair:
Do you need help?

kung-foo:
We can continue on Keybase so as not to clutter up the thread.

PR #328, opened 2020-03-09
uacp.Dial use context for dial

### LINKED ISSUE #327: opcua.Client.Dial never returns on invalid address
When trying to connect using a invalid address, like this:
```golang
	endpoint := "opc.tcp://example.com:48010"
	c := opcua.NewClient(endpoint)
	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)

	defer cancel()
	if err := c.Dial(ctx); err != nil {
		log.Println(err)
		return
	}
	defer c.Close()
	fmt.Println("Connected")
```
I expect that dial would return one of these:  
- context.DeadlineExceeded
- Error that the address is not valid

Instead it will hang and never return.

mmikalsen:
Use `net.Dialer{}.DialContext` instead of `net.DialTCP`. This allows the dial to be canceled by the caller, or by the use of context with a timeout.

Fixes https://github.com/gopcua/opcua/issues/327

### DIFF
--- a/uacp/conn_test.go
+++ b/uacp/conn_test.go
@@ -6,6 +6,7 @@ package uacp
 
 import (
 	"context"
+	"net"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


uacp/conn_test.go
```diff
@@ -6,46 +6,59 @@ package uacp
...
+		case <-done:
+		case err := <-acceptErr:
+			t.Fatalf("accept fail: %v", err)
+		case <-time.After(time.Second):
+			t.Fatal("timed out")
+		}
+	})
+
+	t.Run("Server does not exists", func(t *testing.T) {
+		ep := "opc.tcp://example.com:4840"
```
magiconair: Note that `example.com` resolves to an ip address. While unlikely that they run a server at 4840 it might be nicer to not spam them more. What about `127.0.0.99:56789`?

```sh
$ host example.com
example.com has address 93.184.216.34
example.com has IPv6 address 2606:2800:220:1:248:1893:25c8:1946
example.com mail is handled by 0 .
```

uacp/conn_test.go
```diff
@@ -6,46 +6,59 @@ package uacp
...
+		ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+		defer cancel()
+		_, err := Dial(ctx, ep)
+		if err.(*net.OpError).Err.Error() != "i/o timeout" {
```
magiconair: Can we use the new `errors.Is()` here?

mmikalsen: the `net.OpError.Err` is a `*poll.TimeoutError`, pool is an internal package, therefore we can't compare against that.

magiconair: Ah, too bad. Would have been nice to actually see a use case for this.

kung-foo: what does https://tip.golang.org/src/net/net.go?s=15864:15896#L489 return?

uacp/conn_test.go
mmikalsen: The error only occurs when the Address resolves, so for `127.0.0.99:56789` it will return `connect: connection refused`.

uacp/conn_test.go
mmikalsen: it returns `true`, I guess we can use that instead.

magiconair:
LGTM

PR #333, opened 2020-03-30
Make TCP buffers configurable

magiconair:
This patch adds methods to configure the receive and transmit
buffers of the underlying TCP connection.

### DIFF
--- a/client.go
+++ b/client.go
@@ -234,6 +234,26 @@ func (c *Client) Close() error {
 	return c.sechan.Close()
 }
 
+var errNotConnected = errors.New("not connected")
+
+// SetReadBuffer sets the operating system's TCP receive buffer
+// of the underlying UACP connection.
+func (c *Client) SetReadBuffer(bytes int) error {
+	if c.conn == nil {
+		return errNotConnected
+	}
+	return c.conn.SetReadBuffer(bytes)
+}
+
+// SetWriteBuffer sets the operating system's TCP transmit buffer
+// of the underlying UACP connection.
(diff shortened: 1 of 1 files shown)


magiconair:
We ran into this today and hard-coded the changes. Need to see whether this approach works as well or whether we need to call `SetReadBuffer/SetWriteBuffer` after `Dial` but before the `uacp` handshake for this to work.

kung-foo:
I wonder if we should just promote the underlying conn:

```
type Conn struct {
        net.Conn
	id  uint32
	ack *Acknowledge
}
```

or if we can assume it is _always_ `TCP`:

```
type Conn struct {
        net.TCPConn
	id  uint32
	ack *Acknowledge
}
```

magiconair:
`net.Conn` does not have the `SetRead/WriteBuffer` methods. We would have to type check twice. 

Unless we support Websockets it will be TCP and even then.

kung-foo:
Maybe this? https://github.com/gopcua/opcua/commit/6f756f38adbbba8904af8f0b7df0d1950a7799c5

magiconair:
This looks good. Let me test that.

kung-foo:
note: I didn't understand this bit right here: https://github.com/gopcua/opcua/commit/6f756f38adbbba8904af8f0b7df0d1950a7799c5#diff-e06a25cad06b1ecd848e968ae7d9984eL293-R270

There's some bad naming, and aliasing going on. And I don't think there's anything that tests that path.

magiconair:
This is for the reverse handshake but we don't have any tests for that I think.

PR #336, opened 2020-04-06
disable error handling of publish ACKs

### LINKED ISSUE #337: StatusBadSequenceNumberUnknown issues
In our app we are seeing a significant number of subscription failures with the `StatusBadSequenceNumberUnknown` error. After some digging we've seen two problems:

### ACK handling 

The current code does not handle the ACKs and error handling correctly IMO. If I understand the spec in Part 4, 5.3.13 correctly then 

a `PublishRequest` contains
* The list of messages which can be removed from the servers' retransmission queue `subscription Acknowledgements`

and a `PublishResponse` contains
* The data notification `notificationMessage`
* The list of sequence numbers which can be acknowledged `availableSequence Numbers`
* The status codes of the **previous** acknowledgements `results`

The current code checks the results for errors before forwarding a data notification (https://github.com/gopcua/opcua/blob/v0.1.11/client.go#L685-L700). However, the `results` in the `PublishResponse` are not related to the data notification at all. They are only relevant for the subscription ACKs which need to be handled in the main loop and the `StatusBadSequenceNumberUnknown` should not be bubbled up to the caller.

### Siemens PLC 

This is triggered by a Siemens PLC with the following sequence:

1. Server sends notif 1 and seq nr 1 for ACK
1. Client ACKs notif 1
1. Server sends notif 2 and seq nrs 1&2 for ACK
1. Client ACKs notif 1&2
1. Server sends notif 3 and seq nr 3 for ACK and `StatusBadSequenceNumberUnknown` for the Client ACKing notif 1 twice

This should not happen according to the spec IMO since the list of available sequence numbers for ACK'ing should be compiled after the server has processed the ACKs from the client. 

### Mitigation

I can think of several solutions but in all cases we need to remove the wrong error check from `notifySubscription`:

1. Ignore `BadSequenceNumberUnknown` and assume that this is because of the client sending a double ACK
1. Keep track of the ACKs and the sequence numbers and maybe trigger a republish

@alexbrdn @dwhutchison any of you have an opinion?

I don't know what happens if the client does not ACK the notifications at all.

magiconair:
The result status codes in the publish response are not related
to the notification message data but to the acknowledgements sent
in the previous PublishRequest.

It looks like the Siemens PLC is sending us sequence number that
we have ackknowledged in the previous request and then responds
with a bad sequence number error for one of the sequence numbers
we've acknowledged.

This is correct but not according to spec since the PLC should only
send us packets that have not been ack'ed and the that list should
be compiled *after* processing the PublishRequest with our sequence
numbers.

In any case, there is no correlation between the results and the
notification data so we should not check these errors here at all.

Fixes #337

### DIFF
--- a/client.go
+++ b/client.go
@@ -690,21 +690,21 @@ func (c *Client) notifySubscription(ctx context.Context, response *ua.PublishRes
 	}
 
 	// Check for errors
-	status := ua.StatusOK
-	for _, res := range response.Results {
-		if res != ua.StatusOK {
-			status = res
-			break
-		}
-	}
+	// status := ua.StatusOK
+	// for _, res := range response.Results {
+	// 	if res != ua.StatusOK {
+	// 		status = res
+	// 		break
+	// 	}
+	// }
 
-	if status != ua.StatusOK {
-		sub.sendNotification(ctx, &PublishNotificationData{
-			SubscriptionID: response.SubscriptionID,
-			Error:          status,
-		})
-		return
-	}
+	// if status != ua.StatusOK {
+	// 	sub.sendNotification(ctx, &PublishNotificationData{
+	// 		SubscriptionID: response.SubscriptionID,
+	// 		Error:          status,
+	// 	})
+	// 	return
+	// }
 
 	if response.NotificationMessage == nil {
 		sub.sendNotification(ctx, &PublishNotificationData{


magiconair:
I'll follow up with some more details after rolling out a hotfix.

See #337 for details. This patch only comments out the error handling.

magiconair:
I'll merge this into the v0.1.12 release and open another issue for following up.

magiconair:
We are going to touch subscriptions for the reconnection logic anyway.

PR #345, opened 2020-04-29
Allow adding params on Nodes added to subscription

### LINKED ISSUE #268: monitor.Subscription does not allow changing of monitoring parameters (queue size, sampling interval)
`Subscription.AddNodeIDs` does directly lookup `NewMonitoredItemCreateRequestWithDefaults` which contains standard settings for `QueueSize` = 10 and `SamplingInterval` = 0.0 and applies those to the subscription. 

I am no go expert but this behaviour cannot be override. `AddNodeIDs` should contain extra arguments to supply the monitoring parameters in the method signature.

Are pull requests welcome on this one?

bbergshaven:
Bumped into the same issue as this [Issue](https://github.com/gopcua/opcua/issues/268) and decided to make a PR. First contribution to this repo, so change requests / feedback is appreciated. 

- Make new type Request to contain parameters for the Node to Monitor
- Fix so that one can create multiple monitor Items per node. 
- Return MonitoredItemIDs so that it is possible to delete individual monitored items
- Keep original methods for nodes... (string) and *ua.NodeID... for backward compatibility and convenience.

### DIFF
--- a/monitor/subscription.go
+++ b/monitor/subscription.go
@@ -39,10 +39,29 @@ type NodeMonitor struct {
 	errHandlerCB     ErrHandler
 }
 
-// internal struct to manage various ids
-type itemIDs struct {
-	handle uint32 // client-provided
-	id     uint32 // from server
+// Item is a struct to manage Monitored Items
+type Item struct {
+	id     uint32     // from server
+	nodeID *ua.NodeID // from request
+	handle uint32     // client provided
+}
+
+// ID returns the MonitorItemID set by the server
+func (m *Item) ID() uint32 {
+	return m.id
+}
+
+// NodeID returns the NodeID for the Item
+func (m *Item) NodeID() *ua.NodeID {
+	return m.nodeID
+}
+
+// Request is a struct to manage a request to monitor a node
+type Request struct {
+	NodeID               *ua.NodeID
+	MonitoringMode       ua.MonitoringMode
+	MonitoringParameters *ua.MonitoringParameters
+	handle               uint32
 }
 
 // Subscription is an instance of an active subscription.
@@ -56,7 +75,7 @@ type Subscription struct {
 	closed           chan struct{}
 	mu               sync.RWMutex
 	handles          map[uint32]*ua.NodeID
-	nodeLookup       map[string]*itemIDs
+	itemLookup       map[uint32]Item
 }
 
 // NewNodeMonitor creates a new NodeMonitor
@@ -79,7 +98,7 @@ func newSubscription(m *NodeMonitor, params *opcua.SubscriptionParameters, notif
 		closed:           make(chan struct{}),
 		internalNotifyCh: make(chan *opcua.PublishNotificationData, notifyChanLength),
 		handles:          make(map[uint32]*ua.NodeID),
-		nodeLookup:       make(map[string]*itemIDs),
+		itemLookup:       make(map[uint32]Item),
 	}
 
 	var err error
@@ -238,6 +257,7 @@ func (s *Subscription) Dropped() uint64 {
 
 // AddNodes adds nodes defined by their string representation
 func (s *Subscription) AddNodes(nodes ...string) error {
+
 	nodeIDs, err := parseNodeSlice(nodes...)
 	if err != nil {
 		return err
@@ -247,53 +267,73 @@ func (s *Subscription) AddNodes(nodes ...string) error {
 
 // AddNodeIDs adds nodes
 func (s *Subscription) AddNodeIDs(nodes ...*ua.NodeID) error {
+	requests := make([]Request, len(nodes))
+
+	for i, node := range nodes {
+		requests[i] = Request{
+			NodeID:         node,
+			MonitoringMode: ua.MonitoringModeReporting,
+		}
+	}
+	_, err := s.AddMonitorItems(requests...)
+	return err
+}
+
+// AddMonitorItems adds nodes with monitoring parameters to the subscription
+func (s *Subscription) AddMonitorItems(nodes ...Request) ([]Item, error) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
 	if len(nodes) == 0 {
 		// some server implementionations allow an empty monitoreditemrequest, some don't.
 		// beter to just return
-		return nil
+		return nil, nil
 	}
 
 	toAdd := make([]*ua.MonitoredItemCreateRequest, 0)
 
-	for _, node := range nodes {
+	// Add handles and make requests
+	for i, node := range nodes {
 		handle := atomic.AddUint32(&s.monitor.nextClientHandle, 1)
+		s.handles[handle] = nodes[i].NodeID
+		nodes[i].handle = handle
 
-		s.handles[handle] = node
-		s.nodeLookup[node.String()] = &itemIDs{
-			handle: handle,
+		request := opcua.NewMonitoredItemCreateRequestWithDefaults(node.NodeID, ua.AttributeIDValue, handle)
+		request.MonitoringMode = node.MonitoringMode
+
+		if node.MonitoringParameters != nil {
+			request.RequestedParameters = node.MonitoringParameters
+			request.RequestedParameters.ClientHandle = handle
 		}
-
-		// log.Printf("node=%s handle=%d", node.String(), handle)
-
-		toAdd = append(toAdd, opcua.NewMonitoredItemCreateRequestWithDefaults(node, ua.AttributeIDValue, handle))
+		toAdd = append(toAdd, request)
 	}
-
 	resp, err := s.sub.Monitor(ua.TimestampsToReturnBoth, toAdd...)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	if resp.ResponseHeader.ServiceResult != ua.StatusOK {
-		return resp.ResponseHeader.ServiceResult
+		return nil, resp.ResponseHeader.ServiceResult
 	}
 
 	if len(resp.Results) != len(toAdd) {
-		return errors.Errorf("monitor items response length mismatch")
+		return nil, errors.Errorf("monitor items response length mismatch")
 	}
-
+	var monitoredItems []Item
 	for i, res := range resp.Results {
 		if res.StatusCode != ua.StatusOK {
-			return res.StatusCode
+			return nil, res.StatusCode
 		}
-		// note: this works _iff_ the order of the response is the same as the request
-		sid := toAdd[i].ItemToMonitor.NodeID.String()
-		s.nodeLookup[sid].id = res.MonitoredItemID
+		mn := Item{
+			id:     res.MonitoredItemID,
+			handle: nodes[i].handle,
+			nodeID: toAdd[i].ItemToMonitor.NodeID,
+		}
+		s.itemLookup[res.MonitoredItemID] = mn
+		monitoredItems = append(monitoredItems, mn)
 	}
 
-	return nil
+	return monitoredItems, nil
 }
 
 // RemoveNodes removes nodes defined by their string representation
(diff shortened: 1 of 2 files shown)


kung-foo:
Hi @bbergshaven, thanks for the PR!

Running a quick test using the `examples/monitor` I get:

```
$ go run monitor.go -endpoint opc.tcp://arrakis:53530/OPCUA/SimulationServer -node i=2258
2020/05/05 09:03:09 *http://opcfoundation.org/UA/SecurityPolicy#NoneMessageSecurityModeNone
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x59a837]

goroutine 53 [running]:
github.com/gopcua/opcua/monitor.(*Subscription).AddMonitorItems(0xc000246120, 0xc00009d530, 0x2, 0x2, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/jonathan/src/opcua/monitor/subscription.go:298 +0xf7
github.com/gopcua/opcua/monitor.(*Subscription).AddNodeIDs(0xc000246120, 0xc0000b4098, 0x1, 0x1, 0x1, 0x1)
	/home/jonathan/src/opcua/monitor/subscription.go:277 +0x154
github.com/gopcua/opcua/monitor.(*Subscription).AddNodes(0xc000246120, 0xc0002101d0, 0x1, 0x1, 0x0, 0x0)
	/home/jonathan/src/opcua/monitor/subscription.go:263 +0x92
github.com/gopcua/opcua/monitor.newSubscription(0xc0002124c0, 0xc0002bae58, 0x10, 0xc0002101d0, 0x1, 0x1, 0xc000236280, 0x80, 0x5d8180)
	/home/jonathan/src/opcua/monitor/subscription.go:108 +0x172
github.com/gopcua/opcua/monitor.(*NodeMonitor).ChanSubscribe(0xc0002124c0, 0x653240, 0xc0000f0940, 0xc0002bae58, 0xc000246060, 0xc0002101d0, 0x1, 0x1, 0x0, 0x0, ...)
	/home/jonathan/src/opcua/monitor/subscription.go:141 +0x69
main.startChanSub(0x653240, 0xc0000f0940, 0xc0002124c0, 0x5f5e100, 0x0, 0xc0002101d0, 0x1, 0x1)
	/home/jonathan/src/opcua/examples/monitor/monitor.go:122 +0xfb
created by main.main
	/home/jonathan/src/opcua/examples/monitor/monitor.go:90 +0xabb
exit status 2
```

monitor/subscription.go
```diff
@@ -39,10 +39,28 @@ type NodeMonitor struct {
...
-	id     uint32 // from server
+// Item is a struct to manage Monitored Items
+type Item struct {
+	id     uint32     // from server
+	nodeID *ua.NodeID // from request
+	handle uint32     // client provided
+}
+
+// ID returns the MonitorItemID set by the server
+func (m Item) ID() uint32 {
```
kung-foo: should `m` be a pointer? Or is there a reason these are value receivers?

monitor/subscription.go
```diff
@@ -247,53 +265,74 @@ func (s *Subscription) AddNodes(nodes ...string) error {
 
 // AddNodeIDs adds nodes
 func (s *Subscription) AddNodeIDs(nodes ...*ua.NodeID) error {
+
+	requests := make([]Request, len(nodes))
+
+	for _, node := range nodes {
+		requests = append(requests, Request{
```
kung-foo: This appends _after_ the pre-allocated nodes from L269. Either init an empty slice, or (as I prefer), use the index from `range` to set each element.

see: https://play.golang.org/p/PrCpdwROEdF

monitor/subscription.go
```diff
@@ -350,6 +407,7 @@ func (s *Subscription) RemoveNodeIDs(nodes ...*ua.NodeID) error {
 	return nil
 }
 
+// Stats returns statistic for the subscription
```
kung-foo: statistic**s**

monitor/subscription.go
```diff
@@ -307,25 +346,43 @@ func (s *Subscription) RemoveNodes(nodes ...string) error {
...
+	if len(nodes) == 0 {
+		return nil
+	}
+
+	var toRemove []Item
+
+	for _, node := range nodes {
+		for _, item := range s.itemLookup {
+			if item.nodeID == node {
+				toRemove = append(toRemove, item)
```
kung-foo, line 359: break on match

monitor/subscription.go
```diff
@@ -39,10 +39,28 @@ type NodeMonitor struct {
...
+	handle uint32     // client provided
+}
+
+// ID returns the MonitorItemID set by the server
+func (m Item) ID() uint32 {
+	return m.id
+}
+
+// NodeID returns the NodeID for the Item
+func (m Item) NodeID() ua.NodeID {
```
kung-foo: any reason you return a copy of the node id?

monitor/subscription.go
```diff
@@ -247,53 +265,74 @@ func (s *Subscription) AddNodes(nodes ...string) error {
...
-	for _, node := range nodes {
+	// Add handles
+	for i := range nodes {
 		handle := atomic.AddUint32(&s.monitor.nextClientHandle, 1)
-
-		s.handles[handle] = node
-		s.nodeLookup[node.String()] = &itemIDs{
-			handle: handle,
+		s.handles[handle] = nodes[i].NodeID
+		nodes[i].MonitoringParameters.ClientHandle = handle
```
kung-foo: It's too bad we have to change the incoming request parameters, but I guess it can't be avoided.

kung-foo (changes requested):

monitor/subscription.go
bbergshaven: Fixed this now. I see that you use pointers all over and changed this PR to follow the same style. Just out of curiosity, Is there a good reason for the extended pointer use?

monitor/subscription.go
bbergshaven: Changed to used the index as suggested.

monitor/subscription.go
bbergshaven: 🙈 Fixed.

monitor/subscription.go
bbergshaven: Fixed.

monitor/subscription.go
bbergshaven: Fixed to return a pointer. Same question here as above regarding the use of pointers for small objects like this? Is it to make NodeID nullable?

monitor/subscription.go
bbergshaven: I left all the other incoming methods there and only added a new that accepts the new type Request. So this should not break any usage of this package as far as I can see.

bbergshaven:
> Hi @bbergshaven, thanks for the PR!
> 
> Running a quick test using the `examples/monitor` I get:
> 
> ```
> $ go run monitor.go -endpoint opc.tcp://arrakis:53530/OPCUA/SimulationServer -node i=2258
> 2020/05/05 09:03:09 *http://opcfoundation.org/UA/SecurityPolicy#NoneMessageSecurityModeNone
> panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x59a837]
> 
> goroutine 53 [running]:
> github.com/gopcua/opcua/monitor.(*Subscription).AddMonitorItems(0xc000246120, 0xc00009d530, 0x2, 0x2, 0x0, 0x0, 0x0, 0x0, 0x0)
> 	/home/jonathan/src/opcua/monitor/subscription.go:298 +0xf7
> github.com/gopcua/opcua/monitor.(*Subscription).AddNodeIDs(0xc000246120, 0xc0000b4098, 0x1, 0x1, 0x1, 0x1)
> 	/home/jonathan/src/opcua/monitor/subscription.go:277 +0x154
> github.com/gopcua/opcua/monitor.(*Subscription).AddNodes(0xc000246120, 0xc0002101d0, 0x1, 0x1, 0x0, 0x0)
> 	/home/jonathan/src/opcua/monitor/subscription.go:263 +0x92
> github.com/gopcua/opcua/monitor.newSubscription(0xc0002124c0, 0xc0002bae58, 0x10, 0xc0002101d0, 0x1, 0x1, 0xc000236280, 0x80, 0x5d8180)
> 	/home/jonathan/src/opcua/monitor/subscription.go:108 +0x172
> github.com/gopcua/opcua/monitor.(*NodeMonitor).ChanSubscribe(0xc0002124c0, 0x653240, 0xc0000f0940, 0xc0002bae58, 0xc000246060, 0xc0002101d0, 0x1, 0x1, 0x0, 0x0, ...)
> 	/home/jonathan/src/opcua/monitor/subscription.go:141 +0x69
> main.startChanSub(0x653240, 0xc0000f0940, 0xc0002124c0, 0x5f5e100, 0x0, 0xc0002101d0, 0x1, 0x1)
> 	/home/jonathan/src/opcua/examples/monitor/monitor.go:122 +0xfb
> created by main.main
> 	/home/jonathan/src/opcua/examples/monitor/monitor.go:90 +0xabb
> exit status 2
> ```

Fixed this error and also added a waitGroup to the monitor example to avoid a null pointer exception on ctrl-c.

monitor/subscription.go
kung-foo: This is a pretty good summary: https://github.com/golang/go/wiki/CodeReviewComments#receiver-type

My approach tends to be: pointer unless I have an explicit reason to deviate (which generally only shows up when trying optimize the performance of a function),

monitor/subscription.go
kung-foo: Same as above. No reason to make a copy of the node.

monitor/subscription.go
kung-foo: Ah, not the methods themselves. Rather the `nodes` param in `AddMonitorItems(nodes ...Request)`. It is a little unfortunate that the `nodes` param _changes_ as a result of calling the method. :shrug:

kung-foo (approved):

deleted:
@bbergshaven I'm curious about how you indended this to be used.

It looks like you've modified `AddMonitorItems()` to use `MonitoringParameters` if supplied on each `Request` item, but all the public "subscribe" methods seem to call `newSubscription()`, which eventually calls `AddNodeIDs()` and constructs a bunch of `Request`s with empty `MonitoringParameters`:

```
for i, node := range nodes {
		requests[i] = Request{
			NodeID:         node,
			MonitoringMode: ua.MonitoringModeReporting,
		}
	}
```

Am I missing something? Is there some alternate way to construct subscriptions that lets you inject per-channel parameters?

bbergshaven:
Hello @deleted 

Some time since I did this fix, but the intention was to be able to extend a monitorItem with the required parameters to allow onChange, change queueSize etc. Below I've attached some snippets of our code that shows how we use it. The code consists of snippets from several functions and I have not testet it directly, but it should give you some insight into how you can construct Monitoring Items with parameters.

I hope this helps and I'll do my best to answer any additional questions.

Regards
Bernt-Johan 

```
func testSubscribe(ctx context.Context, c *opcua.Client, nid *ua.NodeID) error {
	var items []monitor.Request

	samplingInterval := 10 * time.Second

	// Regular monitoringitem
	items = append(items, monitor.Request{
		NodeID:         nid,
		MonitoringMode: ua.MonitoringModeReporting,
		MonitoringParameters: &ua.MonitoringParameters{
			SamplingInterval: float64(samplingInterval.Milliseconds()),
			QueueSize:        uint32(5),
			DiscardOldest:    true,
		},
	})

	// OnChange monitoringitem
	items = append(items, monitor.Request{
		NodeID:         nid,
		MonitoringMode: ua.MonitoringModeReporting,
		MonitoringParameters: &ua.MonitoringParameters{
			SamplingInterval: float64(1000),
			QueueSize:        uint32(10),
			DiscardOldest:    true,
			Filter: &ua.ExtensionObject{
				EncodingMask: ua.ExtensionObjectBinary,
				TypeID:       ua.NewFourByteExpandedNodeID(0, id.DataChangeFilter_Encoding_DefaultBinary),
				Value: ua.DataChangeFilter{
					Trigger: ua.DataChangeTriggerStatusValue,
				},
			},
		},
	})

	m, err := monitor.NewNodeMonitor(c)
	if err != nil {
		log.Fatal(err)
	}
	m.SetErrorHandler(func(c *opcua.Client, s *monitor.Subscription, err error) {
		log.Fatal(err)
	})

	ch := make(chan *monitor.DataChangeMessage)

	sub, err := m.ChanSubscribe(ctx, &opcua.SubscriptionParameters{Interval: samplingInterval, MaxKeepAliveCount: 5, LifetimeCount: 15}, ch)
	if err != nil {
		log.Fatal(err)
	}
	defer sub.Unsubscribe()

	_, err = sub.AddMonitorItems(items...)
	if err != nil {
		log.Fatal(err)
	}

	// loop and wait for results
	timeout := 10 * time.Second
	timer := time.NewTimer(timeout)
	for {
		select {
		case <-ctx.Done():
			if err := sub.Unsubscribe(); err != nil {
				return err
			}

			if err := c.Close(); err != nil {
				if errors.Is(err, io.EOF) {
					return nil
				}
				return err
			}

			return ctx.Err()
		case <-timer.C:
			_ = sub.Unsubscribe()
			_ = c.Close()
			return fmt.Errorf("timeout")
		case res := <-ch:
			timer.Reset(timeout) // reset timer
			if res.Error != nil {
				log.Print(res.Error)
				continue
			}
			data := res.Value
			timeStamp := res.SourceTimestamp
			tag := res.NodeID.String()

			fmt.Printf("Got data [%s] %s : %v", timeStamp, tag, data)

		}
	}

}
```

PR #353, opened 2020-07-01
fix(client): fix nil pointer in sendWithTimeout() after connection interrupt

### LINKED ISSUE #350: Nil pointer dereference when connection is lost
I am on v0.1.11

In `client.go` the function `openSecureChannel` is called periodically, when renewing the secure connection:

``` golang
func (c *Client) openSecureChannel(ctx context.Context, open func() error) error {
    if err := open(); err != nil {
        c.cancelMonitor()
        _ = c.conn.Close()
        c.sechan = nil
        return err
    }
    c.scheduleRenewingToken(ctx)
    return nil
}
```
If the connection is lost and this function is called, then `c.sechan` is set to `nil`. This leads to a nil pointer dereference, when calling `sendWithTimeout`. I overlooked other uses of `sechan` and cannot see any other access, that can happen after it is set to nil. Maybe `Close()` is another candidate.

Just checking if a simple check for `nil` is enough to fix this. Feedback is welcome.

alexbrdn:
- add check that sechan is not nil in `sendWithTimeout()`
- add some more nil checks where potential for nil pointer exists
- replace free text error "secure channel not connected" with appropriate standard error `StatusBadServerNotConnected`

closes #350

### DIFF
--- a/client.go
+++ b/client.go
@@ -196,8 +196,7 @@ func (c *Client) openSecureChannel(ctx context.Context, open func() error) error
 		c.sechan = nil
 		return err
 	}
-	c.scheduleRenewingToken(ctx)
-	return nil
+	return c.scheduleRenewingToken(ctx)
 }
 
 func (c *Client) monitorChannel(ctx context.Context) {
@@ -289,7 +288,7 @@ type Session struct {
 // See Part 4, 5.6.2
 func (c *Client) CreateSession(cfg *uasc.SessionConfig) (*Session, error) {
 	if c.sechan == nil {
-		return nil, errors.Errorf("secure channel not connected")
+		return nil, ua.StatusBadServerNotConnected
 	}
 
 	nonce := make([]byte, 32)
@@ -372,6 +371,9 @@ func anonymousPolicyID(endpoints []*ua.EndpointDescription) string {
 //
 // See Part 4, 5.6.3
 func (c *Client) ActivateSession(s *Session) error {
+	if c.sechan == nil {
+		return ua.StatusBadServerNotConnected
+	}
 	sig, sigAlg, err := c.sechan.NewSessionSignature(s.serverCertificate, s.serverNonce)
 	if err != nil {
 		log.Printf("error creating session signature: %s", err)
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -230,8 +229,10 @@ func (c *Client) Close() error {
 	if c.cancelMonitor != nil {
 		c.cancelMonitor()
 	}
-
-	return c.sechan.Close()
+	if c.sechan != nil {
+		return c.sechan.Close()
+	}
+	return io.EOF
```
kung-foo: you return `ua.StatusBadServerNotConnected` everywhere else `sechan` is nil. Any reason to deviate here?

magiconair:
Don’t think the status code would make it over the wire. Gut feel is to use status codes for error responses on active connections and standard errors for the rest. If unsure we should try to keep it consistent though. 

—
Frank Schröder

kung-foo:
`io.EOF` is not returned from `Close()` on `os.File` or `io.PipeWriter` (just spot checked a few) if the underlying handle is `nil` or is already closed. So maybe just return `nil`, `os.ErrClosed`, or something similar to `io.ErrClosedPipe`?

client.go
alexbrdn: I reckon, when one attempts to close a `Client` and there's no `sechan` to close, we can assume that everything is closed and return `io.EOF`, which is what `sechan` would return to signal a successful closure.

alexbrdn:
I would actually be in favor of returning `nil`. Just tried to be consistent with `sechan.Close()`

kung-foo:
of course all of this is moot since no one ever checks the result of a close call :)

magiconair:
I vaguely remember a blog post how returning an error on Close is wrong in POSIX since you can’t really do anything about it anyway.

 If there is no clear indication from the OPC protocol then I suggest we try to be consistent with the Go net package if we can
—
Frank Schröder

alexbrdn:
`net.TCPConn` returns `syscall.EINVAL` ("invalid argument"). What do you think? Should we go with that?

magiconair:
OK, https://www.daemonology.net/blog/2011-12-17-POSIX-close-is-broken.html has nothing to do with this. 

Lets use `ua.StatusBadServerNotConnected` and move on.

magiconair (approved):

magiconair:
@alexbrdn merge at will. Thank you!

PR #355, opened 2020-07-05
WIP: resurrect server branch

### LINKED ISSUE #177: Server Support
I've pushed a WIP branch for the first kick at a server implementation.  The focus on this initial code is to show the `uacp` package changes that need to occur to support listening for unsolicited messages and how that integrates into both the client and the server code.  I anticipate these changes being the first to be merged, then the server code can follow once it's fleshed out.
The primary difference is that I moved the goroutine read loop out of the uasc package and into each of the client and server packages.  The uasc package still deals with decoding/decrypting the messages, but returns the service up to the application layer.  The exception to this are the Open and Close SecureChannel requests; those are dealt with directly in the uasc package without notifying the client/server.

The server implementation has preliminary support for opening and closing secure channels, auto-builds endpoints, and opening/closing/activating sessions.  Authorization has config for the purpose of  endpoint generation, but is otherwise not implemented yet.  The session management is still in a very early state and should be normalized with the client implementation (they're currently totally separate).  Other services are just placeholders for now.  To properly implement further services, the server needs to implement an address space, which is what I've started looking into next.

magiconair:
I've rebased @dwhutchison's `feature/server` branch and want to give it some love. Right now it compiles and I think I've integrated it with the latest changes. There are some obvious things in context handling I've seen but I need to understand the code a bit better and then want to add an AddressSpace and XML NodeSet2 import.

This is **DONE** when we can start a server, show the default nodes and can read/write/browse nodes. Crypto and security might come later depending on the current state of the code. Subscriptions will come later for sure. 

Note that there is also a `magiconair-server` branch which contains my initial attempt in writing a server. It exists only for completeness and we will end up with one implementation.

Fixes #177


kmpm:
Any chance of getting this rebased with current main?
Then there might be a chance of me trying to help out but it seems quite far behind right now.

magiconair:
I can try later. Would be great if you can help

—
Frank Schröder

magiconair:
This is a bit more work but I'll keep trying.

sruehl:
I stumbled over this again while considering using a golang based opcua server for unit tests (seem easier than to abstract every call to gopcua). Any chance to revive this again before it drifts to far away from main @magiconair?

magiconair:
Closing in favor of #737

PR #357, opened 2020-07-17
Add SetTriggering functionality and example

marcus-orchard:
Basic changes to allow monitored item triggering for a given subscription with an example.

### DIFF
--- a/examples/trigger/trigger.go
+++ b/examples/trigger/trigger.go
@@ -1,0 +1,175 @@
+// Copyright 2018-2020 opcua authors. All rights reserved.
+// Use of this source code is governed by a MIT-style license that can be
+// found in the LICENSE file.
+
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/gopcua/opcua"
+	"github.com/gopcua/opcua/debug"
+	"github.com/gopcua/opcua/id"
+	"github.com/gopcua/opcua/ua"
+)
+
+func main() {
+	var (
+		endpoint      = flag.String("endpoint", "opc.tcp://morchard-mbp1.local:53530/OPCUA/SimulationServer", "OPC UA Endpoint URL")
+		policy        = flag.String("policy", "", "Security policy: None, Basic128Rsa15, Basic256, Basic256Sha256. Default: auto")
+		mode          = flag.String("mode", "", "Security mode: None, Sign, SignAndEncrypt. Default: auto")
+		certFile      = flag.String("cert", "", "Path to cert.pem. Required for security mode/policy != None")
+		keyFile       = flag.String("key", "", "Path to private key.pem. Required for security mode/policy != None")
+		triggerNodeID = flag.String("trigger", "", "node id to trigger with")
+		reportNodeID  = flag.String("report", "", "node id value to report on trigger")
+		filter        = flag.String("filter", "timestamp", "DataFilter: status, value, timestamp.")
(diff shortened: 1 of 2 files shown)


examples/trigger/trigger.go
```diff
@@ -0,0 +1,175 @@
...
+
+	"github.com/gopcua/opcua"
+	"github.com/gopcua/opcua/debug"
+	"github.com/gopcua/opcua/id"
+	"github.com/gopcua/opcua/ua"
+)
+
+func main() {
+	var (
+		endpoint      = flag.String("endpoint", "opc.tcp://morchard-mbp1.local:53530/OPCUA/SimulationServer", "OPC UA Endpoint URL")
```
kung-foo: odds that my laptop is called `morchard-mbp1` is quite low :)

kung-foo (changes requested):

subscription.go
```diff
@@ -112,6 +112,22 @@ func (s *Subscription) Unmonitor(monitoredItemIDs ...uint32) (*ua.DeleteMonitore
 	return res, err
 }
 
+func (s *Subscription) SetTriggering(triggeringItemId uint32, add, remove []uint32) (*ua.SetTriggeringResponse, error) {
```
kung-foo: should be `triggeringItemID`

examples/trigger/trigger.go
```diff
@@ -0,0 +1,175 @@
...
+	go sub.Run(ctx) // start Publish loop
+
+	triggeringServerID, triggeredServerID := subRes.Results[0].MonitoredItemID, subRes.Results[1].MonitoredItemID
+	tRes, err := sub.SetTriggering(triggeringServerID, []uint32{triggeredServerID}, []uint32{})
+
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	log.Print(tRes)
```
kung-foo: This just dumps `&{0xc00020f6d0 [The monitoring item id does not refer to a valid monitored item. StatusBadMonitoredItemIDInvalid (0x80420000)] [] [] []}`. Is that useful?

examples/trigger/trigger.go
```diff
@@ -0,0 +1,175 @@
...
+
+	subRes, err := sub.Monitor(ua.TimestampsToReturnBoth, miCreateRequests...)
+	if err != nil || subRes.Results[0].StatusCode != ua.StatusOK {
+		log.Fatal(err)
+	}
+
+	go sub.Run(ctx) // start Publish loop
+
+	triggeringServerID, triggeredServerID := subRes.Results[0].MonitoredItemID, subRes.Results[1].MonitoredItemID
+	tRes, err := sub.SetTriggering(triggeringServerID, []uint32{triggeredServerID}, []uint32{})
```
kung-foo: last param (`remove`) can be `nil`.

examples/trigger/trigger.go
marcus-orchard: lol, dang. ok, most of these look like they suffer from classic late Friday checkin syndrome. my apologies, I updated this to match the default in the other examples.

examples/trigger/trigger.go
marcus-orchard: i changed it to check the status and throw an error if not ok instead.

subscription.go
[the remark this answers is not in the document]
marcus-orchard: that makes me more comfortable anyways, was trying to live by the "when in rome..." adage when I skipped the commenting originally.

PR #358, opened 2020-07-24
Register all known extension objects

### LINKED ISSUE #277: questions regarding extension objects
Couple of extension object related questions

1) Is there any reason we don't register _all_ of them? Or are we expecting the user to register only what they need? For example, I need to add this to my code:

```go
func init() {
	ua.RegisterExtensionObject(ua.NewNumericNodeID(0, id.BuildInfo_Encoding_DefaultBinary), new(ua.BuildInfo))
}
```

2) I am wondering if we can make this copy-pasta-error-prone-code at little bit less verbose:

```go
node := client.Node(ua.NewNumericNodeID(0, id.Server_ServerStatus_BuildInfo))

v, err := node.Value()
if err != nil {
	return err
}

if eo, ok := v.Value().(*ua.ExtensionObject); ok && eo.TypeID.NodeID.IntID() == id.BuildInfo_Encoding_DefaultBinary {
	bi := eo.Value.(*ua.BuildInfo)
	...
}
```

magiconair:
Fixes #277


go.mod
```diff
@@ -5,6 +5,6 @@ go 1.12
 require (
 	github.com/pascaldekloe/goe v0.1.0
 	github.com/pkg/errors v0.8.1
-	golang.org/x/crypto v0.0.0-20190621222207-cc06ce4a13d4
-	golang.org/x/sys v0.0.0-20190624142023-c5567b49c5d0 // indirect
+	golang.org/x/crypto v0.0.0-20200622213623-75b288015ac9
```
kung-foo, line 8: intentional?

kung-foo: ah. nm.

PR #361, opened 2020-08-13
Support different RSA key sizes between client and server

jvstein:
I'm connecting to a KepWare server using the older security settings ("Basic256" or "Basic128Rsa15"). I used the `examples/crypto/` sample to generate my certificate with 2048 and found the server has a 1024 bit certificate. The connection kept failing until I changed my client certificate to 1024 bits, which lead me to this issue. These changes got me working with mismatched key sizes.

I added a new `remoteSignatureLength` field to the `EncryptionAlgorithm` struct, mirroring the `signatureLength` field, and used it in the `verifyAndDecrypt` function. The `signAndEncrypt` function was left unchanged.


kung-foo:
Thanks for the PR!

PR #368, opened 2020-08-24
Reconnection and session reactivation

magiconair:
~~I've squashed the changes from #299 and rebased them on top of `master`. Changes are attributed to @unknownet which I've invited as a contributor to this project so that we can collaborate on getting this change over the line.~~

This is no longer the case. We have taken the changes from #326 and the `feature/auto-reconnect-fix-secure-channel-clean` from @unknownet, rebased them and cleaned them up for testing and review.

Fixes #298
Closes #299

### DIFF
--- a/client.go
+++ b/client.go
@@ -316,6 +316,7 @@ func (c *Client) dispatcher(ctx context.Context) {
 						c.state.Store(Open)
 						rState = RecreateSubscription
 
+					// todo(fs): shouldn't this be state RepairSubscription?
 					case RestoreSubscription:
 
 						if err := c.repairSubscriptions(c.SubscriptionIDs()); err != nil {
@@ -325,46 +326,48 @@ func (c *Client) dispatcher(ctx context.Context) {
 						}
 						rState = None
 
+					// todo(fs): shouldn't this be state RestoreSubscription?
 					case RecreateSubscription:
 
 						subIDs := c.SubscriptionIDs()
-						res, err := c.transferSubscriptions(subIDs)
-						subsToRestore := []uint32{}
 						subsToRepair := []uint32{}
+						subsToRestore := []uint32{}
 
+						res, err := c.transferSubscriptions(subIDs)
 						if err != nil {
 							debug.Printf("Transfer subscriptions failed: %v", err)
 							subsToRestore = subIDs
-							err = nil
 						} else {
-							for id := range res.Results {
-								transferResult := res.Results[id]
+							for i := range res.Results {
+								transferResult := res.Results[i]
 								if transferResult.StatusCode == ua.StatusBadSubscriptionIDInvalid {
-									debug.Printf("Warning suscription %d should be recreated", id)
-									subsToRestore = append(subsToRestore, subIDs[id])
+									debug.Printf("Warning suscription %d should be recreated", subIDs[i])
+									subsToRestore = append(subsToRestore, subIDs[i])
 								} else {
 									debug.Printf(
-										"Subscription %d can be repaired and available",
-										transferResult.AvailableSequenceNumbers[id],
+										"Subscription %d can be repaired with sequence number %d",
+										subIDs[i],
+										transferResult.AvailableSequenceNumbers[i],
 									)
-									subsToRepair = append(subsToRepair, subIDs[id])
+									subsToRepair = append(subsToRepair, subIDs[i])
 								}
 							}
 						}
 
 						if len(subsToRepair) > 0 {
-							if err = c.repairSubscriptions(subsToRepair); err != nil {
+							if err := c.repairSubscriptions(subsToRepair); err != nil {
 								debug.Printf("Repair subscriptions failed: %v", err)
 								subsToRestore = append(subsToRestore, subsToRepair...)
 							}
 						}
 						if len(subsToRestore) > 0 {
-							if err = c.restoreSubscriptions(subsToRestore); err != nil {
+							if err := c.restoreSubscriptions(subsToRestore); err != nil {
 								debug.Printf("Restore subscripitions failed: %v", err)
 								rState = RecreateSession
 								continue
 							}
 						}
+
 						rState = None
 
 					case NonRecoverable:
--- a/client.go
+++ b/client.go
@@ -112,10 +112,9 @@ type Client struct {
 	// session is the active session.
 	session atomic.Value // *Session
 
-	// map of active subscriptions managed by this client. key is SubscriptionID
-	// access guarded by subMux
-	subscriptions map[uint32]*Subscription
-	subMux        sync.RWMutex
+	// subs is the set of active subscriptions by id.
+	subs   map[uint32]*Subscription
+	subMux sync.RWMutex
 
 	// state of the client
 	state atomic.Value // ConnectState
@@ -141,11 +140,11 @@ type Client struct {
 func NewClient(endpoint string, opts ...Option) *Client {
 	cfg, sessionCfg := ApplyConfig(opts...)
 	c := Client{
-		endpointURL:   endpoint,
-		cfg:           cfg,
-		sessionCfg:    sessionCfg,
-		sechanErr:     make(chan error, 1),
-		subscriptions: make(map[uint32]*Subscription),
+		endpointURL: endpoint,
+		cfg:         cfg,
+		sessionCfg:  sessionCfg,
+		sechanErr:   make(chan error, 1),
+		subs:        make(map[uint32]*Subscription),
 	}
 	c.state.Store(Closed)
 	return &c
@@ -241,7 +240,7 @@ func (c *Client) dispatcher(ctx context.Context) {
 			}
 
 			// Stop all subscriptions
-			for _, sub := range c.subscriptions {
+			for _, sub := range c.subs {
 				sub.stop()
 			}
 
@@ -381,8 +380,8 @@ func (c *Client) dispatcher(ctx context.Context) {
 				<-c.sechanErr
 			}
 
-			// Resume all subscriptions
-			for _, sub := range c.subscriptions {
+			// Resume all subs
+			for _, sub := range c.subs {
 				sub.resume()
 			}
 		}
@@ -421,7 +420,7 @@ func (c *Client) SubscriptionIDs() []uint32 {
 	defer c.subMux.Unlock()
 
 	var ids []uint32
-	for id := range c.subscriptions {
+	for id := range c.subs {
 		ids = append(ids, id)
 	}
 	return ids
@@ -431,12 +430,12 @@ func (c *Client) SubscriptionIDs() []uint32 {
 // with the same parameters to replace the previous ones
 func (c *Client) restoreSubscriptions(ids []uint32) error {
 	for _, id := range ids {
-		if _, exist := c.subscriptions[id]; !exist {
+		if _, exist := c.subs[id]; !exist {
 			debug.Printf("Cannot restore subscription %d", id)
 			continue
 		}
 
-		sub := c.subscriptions[id]
+		sub := c.subs[id]
 
 		debug.Printf("Restoring subscription %d", id)
 		if err := sub.restore(); err != nil {
@@ -468,13 +467,12 @@ func (c *Client) transferSubscriptions(ids []uint32) (*ua.TransferSubscriptionsR
 }
 
 // repairSubscriptions repairs all the subscriptions of subscriptionIDs given,
-// if no subscriptionIDs repair all subscriptions
 func (c *Client) repairSubscriptions(ids []uint32) error {
 	c.subMux.RLock()
 	defer c.subMux.RUnlock()
 
 	for _, id := range ids {
-		sub, ok := c.subscriptions[id]
+		sub, ok := c.subs[id]
 		if !ok {
 			return errors.Errorf("invalid subscription id %d", id)
 		}
@@ -962,13 +960,13 @@ func (c *Client) Subscribe(params *SubscriptionParameters, notifyCh chan *Publis
 	}
 
 	c.subMux.Lock()
-	if sub.SubscriptionID == 0 || c.subscriptions[sub.SubscriptionID] != nil {
+	if sub.SubscriptionID == 0 || c.subs[sub.SubscriptionID] != nil {
 		// this should not happen and is usually indicative of a server bug
 		// see: Part 4 Section 5.13.2.2, Table 88 – CreateSubscription Service Parameters
 		c.subMux.Unlock()
 		return nil, ua.StatusBadSubscriptionIDInvalid
 	}
-	c.subscriptions[sub.SubscriptionID] = sub
+	c.subs[sub.SubscriptionID] = sub
 	c.subMux.Unlock()
 
 	return sub, nil
@@ -982,17 +980,17 @@ func (c *Client) registerSubscription(sub *Subscription) error {
 
 	c.subMux.Lock()
 	defer c.subMux.Unlock()
-	if _, ok := c.subscriptions[sub.SubscriptionID]; ok {
+	if _, ok := c.subs[sub.SubscriptionID]; ok {
 		return errors.Errorf("SubscriptionID %d already registered", sub.SubscriptionID)
 	}
 
-	c.subscriptions[sub.SubscriptionID] = sub
+	c.subs[sub.SubscriptionID] = sub
 	return nil
 }
 
 func (c *Client) forgetSubscription(id uint32) {
 	c.subMux.Lock()
-	delete(c.subscriptions, id)
+	delete(c.subs, id)
 	c.subMux.Unlock()
 }
 
@@ -1000,10 +998,10 @@ func (c *Client) notifySubscriptionsOfError(ctx context.Context, res *ua.Publish
 	c.subMux.RLock()
 	defer c.subMux.RUnlock()
 
-	subsToNotify := c.subscriptions
+	subsToNotify := c.subs
 	if res != nil && res.SubscriptionID != 0 {
 		subsToNotify = map[uint32]*Subscription{
-			res.SubscriptionID: c.subscriptions[res.SubscriptionID],
+			res.SubscriptionID: c.subs[res.SubscriptionID],
 		}
 	}
 	for _, sub := range subsToNotify {
@@ -1015,7 +1013,7 @@ func (c *Client) notifySubscriptionsOfError(ctx context.Context, res *ua.Publish
 
 func (c *Client) notifySubscription(ctx context.Context, response *ua.PublishResponse) {
 	c.subMux.RLock()
-	sub, ok := c.subscriptions[response.SubscriptionID]
+	sub, ok := c.subs[response.SubscriptionID]
 	c.subMux.RUnlock()
 	if !ok {
 		debug.Printf("Unknown subscription: %v", response.SubscriptionID)
--- a/client.go
+++ b/client.go
@@ -487,40 +487,41 @@ func (c *Client) repairSubscriptions(ids []uint32) error {
 }
 
 func (c *Client) repairSubscription(sub *Subscription) error {
-	debug.Printf("RepairSubscription  for SubscriptionId %d", sub.SubscriptionID)
-	if err := c.subscriptionRepublish(sub); err != nil {
+	debug.Printf("repairing subscription %d", sub.SubscriptionID)
+	if err := c.republishSubscription(sub); err != nil {
 		status, ok := err.(ua.StatusCode)
 		if !ok {
 			return err
 		}
+
 		switch status {
 		case ua.StatusBadSessionIDInvalid:
 			return nil
 		case ua.StatusBadSubscriptionIDInvalid:
-			debug.Printf("Republish failed, subscriptionId is not valid anymore on server side.")
-			return errors.Errorf("Republish failed, subscriptionId is not valid anymore on server side.")
+			debug.Printf("Republish failed since subscription %d is invalid", sub.SubscriptionID)
+			return errors.Errorf("republish failed since subscription %d is invalid", sub.SubscriptionID)
 		}
 	}
 	return nil
 }
 
-// subscriptionRepublish send republish request for the given subscription
+// republishSubscription send republish request for the given subscription
 // republish should end with a StatusCode BadMessageNotAvailable
 // wich implies that there's no more message to restore
-func (c *Client) subscriptionRepublish(sub *Subscription) error {
+func (c *Client) republishSubscription(sub *Subscription) error {
 	for {
 		req := &ua.RepublishRequest{
 			SubscriptionID:           sub.SubscriptionID,
 			RetransmitSequenceNumber: atomic.LoadUint32(&sub.lastSequenceNumber) + 1,
 		}
 
-		debug.Printf("Republish Request for subscription %d retransmitSequenceNumber=%d",
+		debug.Printf("Republishing subscription %d and sequence number %d",
 			req.SubscriptionID,
 			req.RetransmitSequenceNumber,
 		)
 
 		if c.sessionClosed() {
-			debug.Printf("Republish aborted")
+			debug.Printf("Republishing subscription %d aborted", req.SubscriptionID)
 			return ua.StatusBadSessionClosed
 		}
 
@@ -528,9 +529,10 @@ func (c *Client) subscriptionRepublish(sub *Subscription) error {
 		if err != nil {
 			if err == ua.StatusBadMessageNotAvailable {
 				// No more message to restore
+				debug.Printf("Republishing subscription %d OK", req.SubscriptionID)
 				return nil
 			}
-			debug.Printf("Republish request ends with: %v", err)
+			debug.Printf("Republishing subscription %d failed: %v", req.SubscriptionID, err)
 			return err
 		}
 
@@ -540,7 +542,7 @@ func (c *Client) subscriptionRepublish(sub *Subscription) error {
 		}
 
 		if status != ua.StatusOK {
-			debug.Printf("Republish request ends with: %v", status)
+			debug.Printf("Republishing subscription %d failed: %v", req.SubscriptionID, status)
 			return status
 		}
 	}
--- a/client.go
+++ b/client.go
@@ -250,6 +250,7 @@ func (c *Client) dispatcher(ctx context.Context) {
 				case <-ctx.Done():
 					c.state.Store(Broken)
 					return
+
 				default:
 
 					switch rState {
@@ -260,26 +261,19 @@ func (c *Client) dispatcher(ctx context.Context) {
 						c.sechan.Close()
 						c.sechan = nil
 
-						var scheduleRecreateSecureChan, recreateSecureChan func()
-
-						scheduleRecreateSecureChan = func() {
-							timer := time.NewTimer(c.cfg.ReconnectInterval)
-							select {
-							case <-ctx.Done():
-								return
-							case <-timer.C:
-								debug.Printf("Trying to recreate secure channel")
-								recreateSecureChan()
-							}
-						}
-
-						recreateSecureChan = func() {
-							if err := c.Dial(ctx); err != nil {
-								scheduleRecreateSecureChan()
-							}
-						}
 						debug.Printf("Trying to recreate secure channel")
-						recreateSecureChan()
+						for {
+							if err := c.Dial(ctx); err != nil {
+								select {
+								case <-ctx.Done():
+									return
+								case <-time.After(c.cfg.ReconnectInterval):
+									debug.Printf("Trying to recreate secure channel")
+									continue
+								}
+							}
+							break
+						}
 						debug.Printf("Secure channel recreated")
 						rState = RestoreSession
 
@@ -305,11 +299,13 @@ func (c *Client) dispatcher(ctx context.Context) {
 
 						debug.Printf("Trying to recreate session")
 						s, err := c.CreateSession(c.sessionCfg)
-						if err == nil {
-							err = c.ActivateSession(s)
-						}
 						if err != nil {
 							debug.Printf("Recreate session failed: %v", err)
+							rState = RecreateSecureChannel
+							continue
+						}
+						if err := c.ActivateSession(s); err != nil {
+							debug.Printf("Reactivate session failed: %v", err)
 							rState = RecreateSecureChannel
 							continue
 						}
@@ -360,6 +356,7 @@ func (c *Client) dispatcher(ctx context.Context) {
 								subsToRestore = append(subsToRestore, subsToRepair...)
 							}
 						}
+
 						if len(subsToRestore) > 0 {
 							if err := c.restoreSubscriptions(subsToRestore); err != nil {
 								debug.Printf("Restore subscripitions failed: %v", err)
--- a/client.go
+++ b/client.go
@@ -953,18 +953,14 @@ func (c *Client) Subscribe(params *SubscriptionParameters, notifyCh chan *Publis
 	}
 
 	sub := &Subscription{
-		res.SubscriptionID,
-		time.Duration(res.RevisedPublishingInterval) * time.Millisecond,
-		res.RevisedLifetimeCount,
-		res.RevisedMaxKeepAliveCount,
-		notifyCh,
-		params,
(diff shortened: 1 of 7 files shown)


magiconair:
I've started to work on this a bit more to understand the code. First changes are some minor cleanups like renames, comments, whitespace. 

However, I think the `monitorChannel` and `reconnectProcess` need to be merged since they are racy. They both use `c.secchan` but `reconnectProcess` changes it without locking. Also, they are both looping over the state channel which looks too similar.

unknownet:
Hi magiconair,

Just to make sure that we are synchronized, this branch is a bit old and I have done serveral branches since then:
 - [feature/auto-reconnect](https://github.com/unknownet/opcua/tree/feature/auto-reconnect): I've mentioned creating this branch https://github.com/gopcua/opcua/pull/299#issuecomment-637078279, it's been an attempt to enhance the reconnection lower level mechanism and also to create a unit test.
 - [feature/auto-reconnect-fix-secure-channel](https://github.com/unknownet/opcua/tree/feature/auto-reconnect-fix-secure-channel): this branch is basically `feature/auto-reconnect` with #326
merged into it.

Regarding this PR:
> However, I think the `monitorChannel` and `reconnectProcess` need to be merged since they are racy. They both use `c.secchan` but `reconnectProcess` changes it without locking. Also, they are both looping over the state channel which looks too similar.

I am pretty sure that you are right about that, if I remember correctly, I took kind of the same direction with `feature/auto-reconnect`.
Any how if You could take a look at this commit ([feature/auto-reconnect#b446f31](https://github.com/unknownet/opcua/commit/b446f31a5c74c7feb109f56c0beb2f1a7f59f697)) maybe It could help with the change you want to make by merging `monitorChannel` and `reconnectProcess`.
I suppose the way the client handles the reconnection at this level is the most critical part. Once you set for an approach It will not be hard to add the rest on top of that.

>Also, If I could add a thought, I don't know how much progress has been made for the PR #326 but if It is ready to be merged, perhaps It should be merged before this PR because it will cause some conflict.

magiconair:
I was just starting out this way as well...

image

magiconair:
I'm wondering if it makes sense to use your auto-reconnect-fix-secure-channel branch as a starting point for 0.2 and move forward so that we have something to test. @kung-foo @alexbrdn what do you think?

magiconair:
Note that we also need to fix the problem of #337 for which we have a temp fix in #336

magiconair:
@unknownet which of your branches is implementing the reconnection logic or comes closest? We have a PLC which keeps closing/interrupting the TCP connection and I need to integrate this logic quite soon.

unknownet:
The branch [auto-reconnect-fix-secure-channel](https://github.com/unknownet/opcua/tree/feature/auto-reconnect-fix-secure-channel) is the one I use in production and so far the reconnection seams to work every time. If my memory is good there are some hack with `locker` mechanism I abuse of in `uasc/secure_channel.go` for instance. I would be a good thing to get rid of this pattern.
But both `auto-reconnect-fix-secure-channel` and `feature/auto-reconnect` implements auto reconnection, I just haven't tested `feature/auto-reconnect` for a while and I can't remember for sure if every thing was ok with it.

magiconair:
@unknownet Could I ask you to rebase the https://github.com/unknownet/opcua/tree/feature/auto-reconnect-fix-secure-channel branch to master and then push this branch to this repo? I have invited you as a contributor and it would be great if we could add your knowledge to this project.

magiconair:
I think it probably makes more sense to first rebase #326 from @kung-foo and use that as a basis for your changes. I'd like us to get to a point where we can test this at least on a branch.

I'll try to clean up #326

alexbrdn:
> I'm wondering if it makes sense to use your auto-reconnect-fix-secure-channel branch as a starting point for 0.2 and move forward so that we have something to test.

Yes and yes to using #326 as the base for reconnect

magiconair:
@kung-foo is working on getting #326 into shape.

magiconair:
I've cleaned up #326 into #370. Merging master back into a feature branch does not work well with rebasing ...

unknownet:
From what I understand the next moves for this PR are:
 - Wait for the merging of #370 into master
 - rebase the two commit of this branch from master
 - force push those rebased commits into this branch
 - rebase `feature/feature/auto-reconnect-fix-secure-channel` from this branch _(no squash ?)_
 - merge `feature/feature/auto-reconnect-fix-secure-channel` into this branch
 - correct/adjust the branch
 - squash and merge into master

Is that correct ?

magiconair:
I can change this PR to be based on #370 and then just cherry-pick your changes on top. Then we force push into this PR so that we don't create another one. Then we'll squash and merge #370 when @kung-foo is back around 14:00 and can move forward with this one.

magiconair:
I'm assuming that we only need what is in your `feature/auto-reconnect-fix-secure-channel` branch (minus what is in #370).

unknownet:
> I'm assuming that we only need what is in your `feature/auto-reconnect-fix-secure-channel` branch (minus what is in #370).

Yes, we probably only need that branch, since it's based on `feature/auto-reconnect anyway. One thing though, all the clean up that you've done has probably been in vain. Also, after 14:00 I won't be available before 18:00 (except for questions).

kung-foo:
#370 is merged into master now

magiconair:
Awesome. Thank you @kung-foo

unknownet:
It just stike me, but I remember that I've tested  #326 on a Siemens Sinumerik 840D and It gave me `BadRequestTypeInvalid` error.
The issue was that the client was sending query encrypted with the previous TokenID even though it was not expired yet.

https://github.com/gopcua/opcua/pull/326#issuecomment-645288224

I haven't check but does this branch fix it, or should I open an issue relative to Siemens Sinumerik support ?
I am not aware if this issue only relate to Siemens or is a requirement of Opc UA fundation.

magiconair:
Darn, @unknownet did I just overwrite your changes? I went through the branch and cherry-picked your code on top of master and force pushed it to this branch. So in theory this branch should be the `feature/auto-reconnect-fix-secure-channel-clean` but rebased on `master`. Can you please check?

magiconair:
I think the Sinumerik fix should be in 1237633

magiconair:
Once we've verified that the code works and contains all relevant changes then I'll start a cleanup sweep through the code. Then we can do a review.

magiconair:
Refactoring notes:

* `resumeC` has two functions: resume and quit and the latter triggers the cancellation of outstanding publish requests. We should decouple them with two channels `resume` and `quit` which allows us to drop the `publishCancel` and probably also the wait group.

unknownet:
> Darn, @unknownet did I just overwrite your changes? I went through the branch and cherry-picked your code on top of master and force pushed it to this branch. So in theory this branch should be the `feature/auto-reconnect-fix-secure-channel-clean` but rebased on `master`. Can you please check?

Yes, everything seems to be ok, I run a `git clone -b 298-reconnect-reactivate` and then a `git reset --soft 6d59c31` and the only differences I've seen aren't mine.

kung-foo:
> It just stike me, but I remember that I've tested #326 on a Siemens Sinumerik 840D and It gave me `BadRequestTypeInvalid` error.
> The issue was that the client was sending query encrypted with the previous TokenID even though it was not expired yet.
> 
> [#326 (comment)](https://github.com/gopcua/opcua/pull/326#issuecomment-645288224)
> 
> I haven't check but does this branch fix it, or should I open an issue relative to Siemens Sinumerik support ?
> I am not aware if this issue only relate to Siemens or is a requirement of Opc UA fundation.

So when I first tested this change it was against a Kepware server and node js server, it didn't require the change to the `secureChannelID`. However when I tested against Prosys this evening it failed with `BadSecureChannelIdInvalid`. Applying @unknownet 's fix seemed to work. Anyone opposed to a PR with just that fix since the rest of the secchan fixes are already in master?

magiconair:
I've only done cosmetic changes so far. Will work on the quit/resume logic tomorrow. 💤  now

magiconair:
@kung-foo please do. (Separate PR)

kung-foo:
cherrypicked here: https://github.com/gopcua/opcua/pull/371

magiconair:
@unknownet I'm a bit uncomfortable on how you use `sync.WaitGroup` since they are designed to be synchronizing go routines. You should call `wg.Add(1)` in the main go routine and then `wg.Done()` in the child go routine. Then you `wg.Wait()` in the main go routine again. 

This code feels strange since you are calling `wg.Add()` and `wg.Done()` in the same go routine.

```go
	s.pendingReq.Add(1)
	respRequired := h != nil

	ch, err := s.sendAsyncWithTimeout(req, reqID, instance, authToken, respRequired, timeout)
	s.pendingReq.Done()
	if err != nil {
		return err
	}
```

and 

```go
	for {
		sctx, s.publishCancel = context.WithCancel(ctx)
		s.publishWg.Add(1)
		s.run(sctx)
		s.publishWg.Done()

		select {
		case <-ctx.Done():
			return
		case _, ok := <-s.resumeC:
			if !ok {
				return
			}
		}
	}
```

Also, in the second case you would create a new context on every iteration. We're probably not doing this often but it still smells like a resource leak. Also, if you get a `cancel()` function you must call it. That's why there is usually a `defer cancel()` in most code.

Third, the use of `resumeC` is for checking if the channel was closed, right?

Some of the code has a bit of a C smell to it and the number of locks we're using on the secure channel makes me also a bit uncomfortable. Feels like overkill but maybe I don't understand the code well enough yet. 

When I'm done with the cleanup please look through this carefully.

unknownet:
> @unknownet I'm a bit uncomfortable on how you use `sync.WaitGroup` since they are designed to be synchronizing go routines. You should call `wg.Add(1)` in the main go routine and then `wg.Done()` in the child go routine. Then you `wg.Wait()` in the main go routine again.
> 
> This code feels strange since you are calling `wg.Add()` and `wg.Done()` in the same go routine.
> 
> ```go
> 	s.pendingReq.Add(1)
> 	respRequired := h != nil
> 
> 	ch, err := s.sendAsyncWithTimeout(req, reqID, instance, authToken, respRequired, timeout)
> 	s.pendingReq.Done()
> 	if err != nil {
> 		return err
> 	}
> ```
> 
> and
> 
> ```go
> 	for {
> 		sctx, s.publishCancel = context.WithCancel(ctx)
> 		s.publishWg.Add(1)
> 		s.run(sctx)
> 		s.publishWg.Done()
> 
> 		select {
> 		case <-ctx.Done():
> 			return
> 		case _, ok := <-s.resumeC:
> 			if !ok {
> 				return
> 			}
> 		}
> 	}
> ```


__I agree It certainly is a missuse of the WaitGroup mechanism, the goal of this was to make sure that once stop was called there were no publish ongoing anymore.__


> Also, in the second case you would create a new context on every iteration. We're probably not doing this often but it still smells like a resource leak. Also, if you get a `cancel()` function you must call it. That's why there is usually a `defer cancel()` in most code.

__For sure, that was a misstake recreating a context in every iteration, indeed useless and dangerous, but in my defense run is blocking and it only iterate with stop/resume is called. I am sure there's a better and obvious pattern for stopping and resuming the run subscription, probably just by closing and rerunning it.__

> Third, the use of `resumeC` is for checking if the channel was closed, right?

__No, `resumeC` is meant to unblock the waiting select so that the loop restart and rerun the publish loop, once again the all thing is part of an hazardous design.__

```go
		select {
		case <-ctx.Done():
			return
		case _, ok := <-s.resumeC:
			if !ok {
				return
			}
		}
 ```


> Some of the code has a bit of a C smell to it and the number of locks we're using on the secure channel makes me also a bit uncomfortable. Feels like overkill but maybe I don't understand the code well enough yet.
> 
> When I'm done with the cleanup please look through this carefully.

__I was still on a early stage of the learning curve of Golang when I wrote that, I've settled couple of times for design I was not happy with hopping to improve It or being help to do so afterward, just know that I am gratful for your eyesight as It helps me progress.
As of `secure channel` all the modification in this file was meant to fix the Sinumerik issue, as I explain here: [issuecomment-681909843](https://github.com/gopcua/opcua/pull/371#issuecomment-681909843), the cherrypick fix is probably not suffisant, and I suggest you drop the modification in this file for this branch (that's also why I had two branch in the first place).
The only part in secure channel that was not related to this issue was the following [ref](https://github.com/gopcua/opcua/blob/38ade1b8a1566f9765232c64379ccf32d0e5e735/uasc/secure_channel.go#L358-361):__

```go
	// do not wrap this error since it hides conn error
	if _, ok := err.(*uacp.Error); ok {
		return nil, err
	}
```

magiconair:
No worries and I didn't mean that as a criticism. In order to refactor this - if necessary - I need to understand what this is trying to accomplish. From what I understand, we need to be able to pause, resume and stop the subscriptions as well as handle any context cancellations. 

When a subscription is paused you want to ensure that the current publish request is completed. You want to achieve the same thing for `stop`?

I'll try to push something today and I'm looking forward to your comments.

unknownet:
It's ok I was not offended ^^ it was just to give you context so that you don't overthink when something looks odd. Your sum up is correct even if I am not sur anymore if ensuring that the publish request is complete is an absolute necessity, It felt safer that way.

magiconair:
OK, so the last commit is an attempt to refactor this approach. Not sure if this is much cleaner but please have a look. I'll replace the log statements once we're comfortable with this.

Thank you so much for writing an integration test for this. This made this refactoring so much easier!

unknownet:
Commit | Note
---------|-------
38ade1b8a1566f9765232c64379ccf32d0e5e735 | OK, Nothing to add 
06f32765fa4623c10e24b3c0e991b9c3a255abfe | OK, Nothing to add

625b0da0d553c5b55839ee032f96e19d0a0214e5
The pause/resume mechanism is much more readable, outside/inside label aside I came across the same approach and I tought I could do better but I am not sure anymore.
The only remaining problem I can think of is since the `pause` method is not blocking it could lead to unexpected error on recreating securechannel (to be confirmed).

magiconair:
`pause` should be blocking since the channel is not buffered and `run()` is not run in a separate go routine. At least that's my thinking. There is no way `paused` can be picked up by `Run()` without `run()` to complete first.

Unless I am missing something or you had something else in mind.

magiconair:
The only change I've added was to use `close(s.stopch)` again in order to unblock the `pause()` `resume()` calls but maybe that's not necessary.

unknownet:
> `pause` should be blocking since the channel is not buffered and `run()` is not run in a separate go routine. At least that's my thinking. There is no way `paused` can be picked up by `Run()` without `run()` to complete first.
> 
> Unless I am missing something or you had something else in mind.

Yes it should be and It is ^^ sorry I misread, it's all right!

unknownet:
> The only change I've added was to use `close(s.stopch)` again in order to unblock the `pause()` `resume()` calls but maybe that's not necessary.

Can't hurt anyway. Now that I think of it, We should keep the same mecanism in secure_channel:

`secure_channel` L711:
```go
	select {
	case <-s.disconnected:
		s.popHandler(reqID)
		return io.EOF
	case resp := <-ch:		case resp := <-ch:
		if resp.Err != nil {			if resp.Err != nil {
			if resp.V != nil {				if resp.V != nil {
```
Otherwise the pending publish wait for the timeout and return an `ua.StatusBadTimeout`, if I am not misstaken we want to purge those publish or the `pause` is gonna be delayed by the timeout

unknownet:
Hi, I took the liberty to push a commit [63e7260](https://github.com/unknownet/opcua/commit/63e7260212dab0fcf988005afcf342fd69a06ede) in my fork that cleans `secure_channel` and make client reconnection mechanism a bit more consistent.
If you were several commit ahead feel to ask me to update it.
Most of the code review parts are covered by this commit.

Also I opened up a new PR #380 which have the parts that I've just removed from `secure_channel`.

unknownet:
Also I've noticed something odd with the reconnection unit test, for every test:
 - connection_failure
 - securechannel_failure
 - session_failure
 - subscription_failure

the error received by the secure channel is `EOF`.
I am sure at __99%__ that when I've developed the unit test I could see from `secure_channel` the different errors:

 - EOF
 - BadSecureChannelIdInvalid
 - BadSessionIdInvalid
 - BadSubscriptionIdInvalid

Could it be that the behaviour of the client or server (doubtful) has changed ? or the unit test has never actually worked in the first place ?

magiconair:
I think what bugs me a bit is that the changes seem to be all over the place and there isn't enough documentation on the methods and in the methods to follow the logic. This makes it quite hard to review the flow unless you have a thorough understanding of the protocol and all its edge cases. 

We have `master`, the #380 fix against `master` which is a significant refactor in itself, the https://github.com/unknownet/opcua/commit/63e7260212dab0fcf988005afcf342fd69a06ede commit against this branch and this branch. 

We have four different send methods with long argument lists which all call each other: `SendRequest`, `SendRequestWithTimeout`, `sendRequestWithTimeout` and `sendAsyncWithTimeout`. I think the first two are clear but for the rest it isn't - at least to me when I'm reading the code. The new code adds more locks which other changes remove again. 

Writing code that works is only one part of the story. We need to write code that others can understand so that project remains maintainable - especially in this crucial part of the protocol.

I want this to land but since this is a significant change in the core logic we need to iterate on one thing IMO and spend more focus on readability. I need your help to drive this @unknownet and I'm more than happy to help with readability and refactoring. But for that I need more comments and documentation. In the end I'd like you, @kung-foo and @alexbrdn and myself to be comfortable with this change and the way it is written. 

I think what I would like to see is that we focus our efforts on this branch so that we have one reference point to test with. We rebase it against `master` to make sure that it contains the latest changes. I will try to deploy this in our factory setup and monitor it closely. 

We also add more comments to the methods and cases in the state machines which explain how we can get into this state. I also suggest to fold the various send methods into a smaller set and maybe add separate structure we pass around for sending requests, e.g. 

```go
type SecChannelRequest struct {
	Req       ua.Request
	AuthToken *ua.NodeID
	Timeout   time.Duration
	Handler   func(interface{}) error

	reqID        uint32
	instance     *channelInstance
	respRequired bool
}
```

If you need to test something, create a local branch from this one and work on it until it is ready to be added here.

Another option would be to accept that `master` is somewhat broken for a while but I am not a big fan of this since we're all busy and cannot control how much time we can spend on this project. Then we mess things up for everybody else. 

So let us try to make this the branch for the next major release, test it and ask people to test it. Does that sound like a good strategy for moving forward?

magiconair:
Can we maybe come up with a checklist of things we want to accomplish and which are still missing?

magiconair:
rebased against master

unknownet:
> We have four different send methods with long argument lists which all call each other: `SendRequest`, `SendRequestWithTimeout`, `sendRequestWithTimeout` and `sendAsyncWithTimeout`. I think the first two are clear but for the rest it isn't - at least to me when I'm reading the code. The new code adds more locks which other changes remove again.

If by that you mean the locks in `secure_channel`:
```go
	reqLocker  *conditionLocker	
	rcvLocker  *conditionLocker	
	pendingReq sync.WaitGroup
```

Those locks have been initially added because you've merged my branch [feature/auto-reconnect-fix-secure-channel](https://github.com/unknownet/opcua/tree/feature/auto-reconnect-fix-secure-channel) wich contained a fix for the Siemens Sinumerik. Later on we choose to strip away the Sinumerik fix in a separated PR.
So I supposed that all the logic involved for the Sinumerik fix had to be removed from this branch.

> We have master, the #380 fix against master which is a significant refactor in itself, the unknownet@63e7260 commit against this branch and this branch.

I know, but unless if I have made a mistake this branch and #380 does not depend on one another, so I suppose we can postpone #380 after resolving this PR.

>We also add more comments to the methods and cases in the state machines which explain how we can get into this state. I also suggest to fold the various send methods into a smaller set and maybe add separate structure we pass around for sending requests, e.g.

I am ok with using a smaller set of send methods, although the last [commit](https://github.com/unknownet/opcua/commit/63e7260212dab0fcf988005afcf342fd69a06ede) I've pushed has removed any changes to the send methods and I am not sure if it's a good idea to make those changes in this PR, perhaps in #380.
I suppose @kung-foo is probably making changes in `secure_channel` because of this issue #376 and I don't know if it's constructive to make changes in parallel.

> So let us try to make this the branch for the next major release, test it and ask people to test it. Does that sound like a good strategy for moving forward?

Yes, sounds good to me !

> Can we maybe come up with a checklist of things we want to accomplish and which are still missing?

Ok, I'll do that ! And add more comments o.c.

unknownet:
The reason why I called the state `restoreSubscription` was because I had the point of view of the server in mind, which now that I think of it is wrong considering that this is the client...

```go
	// todo(fs): shouldn't this be state RepairSubscription?
	case restoreSubscription:

		if err := c.repairSubscriptions(c.SubscriptionIDs()); err != nil {
			debug.Printf("Restore subscription failed: %v", err)
			reconn = recreateSecureChannel
			continue
		}
		c.state.Store(Connected)
		reconn = none

	// todo(fs): shouldn't this be state RestoreSubscription?
	case recreateSubscription:
```	
So yes you rigth it's more consistent to rename `restoreSubscription` into `repairSubscription` and `recreateSubscription` into `restoreSubscription`.

magiconair:
If we need to make changes to the send methods to make this change clearer then let's do it. First I need more comments and documentation to understand what is going on.

unknownet:
I have added 2 comments commit in my fork 23502f8f3b41f3996946a70bee1ec5b4932e5b29 and ddcaaa42defa04862f80d546746728259a3fff9c how should I proceed with the modifications, I don't want to make a mess ?

unknownet:
Also If we rename `restoreSubscription` and `recreateSubscription`, we should do the same with `restoreSession` and `recreateSession` to be consistent.

magiconair:
These are good. Just cherry-pick them onto this branch and push. If you want to keep a separate copy for experiments just create another branch in this repo. You should have permissions to do so.

unknownet:
> These are good. Just cherry-pick them onto this branch and push. If you want to keep a separate copy for experiments just create another branch in this repo. You should have permissions to do so.

I've tried to push my commit onto this branch and I've got this:

```console
remote: Permission to gopcua/opcua.git denied to unknownet.
fatal: unable to access 'https://github.com/gopcua/opcua.git/': The requested URL returned error: 403
```

magiconair:
I thought I've invited you but apparently not. Invite just went out.

unknownet:
Oops error L51

```go
sort.Sort(sort.Reverse(bySecurityLevel(endpoints)))
```
I forgot, I suppose you rebased the commits, should I change it and force push ?

unknownet:
To clear things up a bit:

opc_reconnection

unknownet:
Now that I think of it `restoreSubscription` does not restore the subscriptions it's a deceptive name.
The restoration of subscription come along with the restoration of the session.
`restoreSubscription` truely is `republishSubscriptions`...

here is a suggestion of new state names:

name | observation
-------|---------------
`recreateSecureChannel` | :heavy_check_mark:
`restoreSession` | :grey_question: `reactivateSession`
`recreateSession` | :grey_question: `restoreSession`
`restoreSubscription` | :grey_question: `republishSubscriptions`
`recreateSubscription` | :grey_question: `restoreSubscriptions`
`nonRecoverable` |  :heavy_check_mark:

magiconair:
Republish as in resend `PublishRequests`?

unknownet:
> Republish as in resend `PublishRequests`?

Yes, if I understand correctly `republish` is meant for asking the server to reply to the pending `PublishRequest` that were waiting for a response just before the connection crashed.

>This Service requests the Subscription to republish a NotificationMessage from its retransmission queue. If the Server does not have the requested Message in its retransmission queue, it returns an error response.

magiconair:
I've pushed some cleanups which should be self-explanatory. The only real change is in df26018 which starts the connection monitor (previously `dispatcher`) only after the initial connection has been established. This seems more in line with the OPC/UA connection model but I'm not sure if this helps in the real world. Was there a reason you've started the connection monitor before the session had been created @unknownet? Wouldn't this create a race condition with the initial connection setup or was that the intention?

I have some more thoughts on naming which I'll post in a minute.

magiconair:
@unknownet I think I've found one more issue. When folding the error handling code into a single switch statement I've noticed that not all errors are handled. The previous code handles `io.EOF`, `syscall.ECONNREFUSED` and `*uacp.Error` but nothing else. I've changed that in 0ae46f9

magiconair:
Also, the `fallthrough` for `ua.StatusBadCertificateInvalid` should just trigger a new secure channel. Wouldn't it be clearer to just use `action = recreateSecureChannel` or was there another intention when you used `fallthrough`?

unknownet:
> Was there a reason you've started the connection monitor before the session had been created @unknownet? Wouldn't this create a race condition with the initial connection setup or was that the intention?

Yes, I suppose my intention was to trigger the reconnection if the CreateSession or ActivateSession were to fail, I think It make more sens if the reconnection start only after Connection as succedded.

magiconair:
Some thoughts about naming and the state model:

I've struggled a while with the names and am still not fully clear but would argue that we `restore` existing things and `create` new things. For example, we first have to create a new secure channel to restore the previous session and so forth. 

The edge case are subscriptions which can be restored in two different ways: either by republishing items for an existing subscription or by creating a new one. Also, when republishing fails we still need to recreate the subscription. 

So I think your suggestions to rename `restoreSubscription` to `republishSubscriptions` and `recreateSubscription` to `restoreSubscriptions` is correct. This also means renaming the `repairSubscriptions` methods to `republishSubscriptions`.

However, when looking at that code I've noticed that when we transfer we fail to republish all subscriptions if one of them fails. Shouldn't we only recreate the subs which failed to republish?

```go
if len(subsToRepair) > 0 {
	if err := c.repairSubscriptions(subsToRepair); err != nil {
		debug.Printf("Repair subscriptions failed: %v", err)
		subsToRestore = append(subsToRestore, subsToRepair...)
	}
}
```

The second issue is in the `restoreSubscription/republishSubscriptions` path. It says that we recreate the subscriptions but I don't see that. This should actually be an endless loop `restoreSubsciption -> createSecureChannel -> restoreSession -> restoreSubscription`. Am I missing something?

```go
case restoreSubscription:
	// try to repair the previous subscriptions on the server
	// otherwise restore it

	if err := c.repairSubscriptions(c.SubscriptionIDs()); err != nil {
		debug.Printf("Restore subscription failed: %v", err)
		action = recreateSecureChannel
		continue
	}
	c.state.Store(Connected)
	action = none
```

I am going to push some changes which reflect this discussion in a minute also with some `todo`s. Lets use this as a basis for discussion.

unknownet:
> Also, the fallthrough for ua.StatusBadCertificateInvalid should just trigger a new secure channel. Wouldn't it be clearer to just use action = recreateSecureChannel or was there another intention when you used fallthrough?

No I only though this way it would be clearer that `ua.StatusBadCertificateInvalid` was unimplemented and should be added at a certain point

unknownet:
> The edge case are subscriptions which can be restored in two different ways: either by republishing items for an existing subscription or by creating a new one. Also, when republishing fails we still need to recreate the subscription.

If I may add I have an issue an a production client, that causes to double the number of subscription on the server at some reconnexion (maybe every), which eventually lead to connection erreur `tooManyMonitoredItems` so I suppose that something is  wrong here. I probably recreate a subscription even if the last one has been recovered.

unknownet:
> So I think your suggestions to rename restoreSubscription to republishSubscriptions and recreateSubscription to restoreSubscriptions is correct. This also means renaming the repairSubscriptions methods to republishSubscriptions.

Yes

magiconair:
Then have a look at the latest change which should address that.

unknownet:
> The second issue is in the restoreSubscription/republishSubscriptions path. It says that we recreate the subscriptions but I don't see that.

No, it doesn't... I was going to change that

> This should actually be an endless loop restoreSubsciption -> createSecureChannel -> restoreSession -> restoreSubscription. Am I missing something?

I suppose I figured the only way this method could return an error would be if there was another disconnection `io.EOF`, that case we would want to fall back into `createSecureChannel` and restart the reconnection process.

I think in every state It could come in handy to check error for `io.EOF` and fall back to `createSecureChannel` the handle a disconnection while reconnecting

unknownet:
Also the all logic of how I've implemented `sendRepublishRequests` isn't correct.

At some point I believed that `RetransmitSequenceNumber` was a new `SequenceNumber` you feed the server so that It can be used to resend any `NotificationMessage` back with the given `SequenceNumber`.
So I thought by storing the latest acknowledged `SequenceNumber` It could use it to ask for republish and so doing avoiding conflict with previous `SequenceNumber`.

While in fact `RetransmitSequenceNumber` requires a specific `SequenceNumber` that was already sent to the server.
That means subscription should store all the `publish` `SequenceNumber` and release them with acknowlegment.
So that when the connection falls It can retransmit only the right `SequenceNumber`s.
But that would require an access to request `SequenceNumber` in subscription.

magiconair:
I think the `EOF` check is a good idea. The sequence number issue is related to #369

unknownet:
I've rebased my experimental branch on your latest modifications.

Perhaps, there are some commits that could be useful to cherry-pick or take into account:

commits | suggestion
----------|-------------
0567b4e92c84d1192e1027db69cc253ce5f3e7a1 | remove double use of cancel (already in defer)
5da13c7b06a0112ff454ea96e952a484f63b4fb7 | inital state should be `disconnected` not `closed`
06ccbc5fdbb92c21ac11e98a14d99af7efa582f9 | `SubscriptionIDs` renamed, should be `subscriptionIDs`

antlad:
@magiconair @unknownet Guys any progress on this? Maybe some help needed?

magiconair:
Not yet. As mentioned in #384 I've hurt my foot unfortunately and got underwater in day job. Living a bit on pain killers.

The main thing on my plate is to fix the handling of subscription acks in #369 so that we know which messages should be republished. I'm sure we'll find other edge cases but in terms of missing/broken functionality this is the one thing missing. 

@unknownet if you think these commits are useful then please cherry-pick them onto this branch. 

I would like to get this branch into a testable state and in terms of functionality the subscription re-publish is the only thing in the way. @antlad it would be great if you can help out with this.

magiconair:
I've started working on the subscription ack handling.

magiconair:
I have started a separate discussion on the handling of subscriptions in general in #385 . This looks like it might be a larger refactoring.

magiconair:
Just to sync: we currently have this PR and the `298-reconnect-reactivate` branch as the basis for the auto-reconnect logic. Implementing this correctly also means to refactor the subscription logic. 

I think having a single publish loop is the way to go. Therefore, we should test if #385 works and merge it into this PR. I will try to do a final review of the transfer and republish logic over the weekend. 

This would mean that we need to see what we still need from #383 and #381 (@unknownet).

* #368 this branch
* #385 refactored subscriptions based on #368 (single publish loop, fixes #369, @magiconair)
* #383 refactored subscriptions and hardened reconnect based on #368 (@unknownet)
* #381 changes to reconnect logic (no conditionLocker, initial state `disconnected`, @unknownet) <- do we still need this?

Is the secure channel race condition a separate problem? That's how I understand it so far.

* #381 (@unknownet)

magiconair:
rebased to master (v0.1.13)

unknownet:
>#383 refactored subscriptions and hardened reconnect based on #368 (@unknownet)

I have to take some time to compare this branch actual state with #368 and determine if my changes are still relevant.  

>#381 changes to reconnect logic (no conditionLocker, initial state disconnected, @unknownet) <- do we still need this?
Is the secure channel race condition a separate problem? That's how I understand it so far.

You're right `conditionLocker` is part of a seperate problem, that's why I came up with this PR, to get ride of all those racy problematics patch that perhaps only relates to Sinumerik Opc Ua server. I suppose It would be better to deal with it in a second time.
Plus I suggested a better implementation with the PR #380.
To answer your question I believe we still need this and I think It would be a better to apply It and reconsider a new approach or maybe just merge #380.

magiconair:
Since we are not getting any feedback on these branches I suggest to merge #368  and #385 to master and tag v0.2.0. At this point I'm OK with some issues since then we have a basis to work on. In the current state we are not making any progress.

Unless there are objections I'm going ahead with this early next week. I'll make sure we dogfood this at Northvolt as well.

CC @unknownet @kung-foo @alexbrdn

PR #371, opened 2020-08-26
fix for secure channel renew with opc sinumerik server

kung-foo:
this also fixes secure channel renew with Prosys simulation server

### DIFF
--- a/uasc/secure_channel.go
+++ b/uasc/secure_channel.go
@@ -457,6 +457,7 @@ func (s *SecureChannel) open(ctx context.Context, instance *channelInstance, req
 		// this seems racy. if another request goes out while the other open request is in flight then won't an error
 		// be raised on the server? can the sequenceNumber be as "global" as the request ID?
 		s.openingInstance.sequenceNumber = instance.sequenceNumber
+		s.openingInstance.secureChannelID = instance.secureChannelID
 	}
 
 	// trigger cleanup after we are all done


magiconair (approved):

unknownet:
From what I understand that's no gonna be suffisant to fix the Sinumerik problem.
Let's consider those 2 scenarii:
 - The first one on the left side send a request after the token renewal has been sent to the server.
the server can still receive response and use the first token to decode messages.
But It wont accept the request encoded with the new TokenId after the renewal request.

- The second one stop sending the request after a renewal request has been sent. Once the new Token Id has been returned by the server, the client use it to send the pending requests and the future ones.

secure_channel_scenario

The fix I came up with was not ideal and I don't like the "stop-the-world lock" approach attempt I made, It feels a bit sloppy.
This was the basic implementation idea I had:
fix attempt

unknownet:
I've commited the implementation I described above [commit 6838034](https://github.com/unknownet/opcua/commit/68380345d6167dd67bf70674ec04e0f40074c662), I am gonna test it tomorrow.

unknownet:
I've commited a working fix for my test with Sinumerik ccce94dc9d7fad02a5c06778436492920e7a7573, the last commit include a fix for concurrent access to `activeChannelInstance` between dispatcher (wich read it) and renew (wich write it).
The issue was due to the fact that `activeChannelInstance` could be written after dispatcher receive a new response that require the new `activeChannelInstance` to be decoded.

PR #372, opened 2020-08-26
add DialTimeout and cleanup ResolveEndpoint

kung-foo:
* cleanup and rebase of https://github.com/gopcua/opcua/pull/278.
* implemented suggestion from @dwhutchison in https://github.com/gopcua/opcua/pull/278#discussion_r331599367


uacp/endpoint_test.go
```diff
@@ -38,11 +38,12 @@ func TestResolveEndpoint(t *testing.T) {
 			"",
 		},
 		{ // Valid, hostname resolved
-			"opc.tcp://localhost:4840/foo/bar",
+			// note: xip.io is hosted by Basecamp (see: https://github.com/basecamp/xip-pdns)
```
magiconair: This is awesome. I never knew.

kung-foo: I suppose this should be in an integration test rather than a normal unit test since it requires a mostly working network stack. :shrug:

kung-foo: xip is dead. long live https://sslip.io/

magiconair:
@kung-foo I've added the `DialTimeout` and the `Dialer` to #468 . Maybe you can refactor this so that we get the `ResolveEndpoint` cleanup?

magiconair:
Still want to clean this one up @kung-foo ?

kung-foo:
yeah, let me take a crack at resolving the conflicts and cleaning up.

uacp/conn.go
```diff
@@ -26,6 +26,7 @@ const (
 	DefaultSendBufSize    = 0xffff
 	DefaultMaxChunkCount  = 512
 	DefaultMaxMessageSize = 2 * MB
+	DefaultDialTimeout    = time.Second * 10
```
magiconair: Could you change this to `10 * time.Second` so that it's similar to the line above?

uacp/endpoint_test.go
```diff
@@ -5,66 +5,78 @@
...
 	for _, c := range cases {
-		network, addr, err := ResolveEndpoint(c.input)
-		require.Equal(t, c.network, network, "network not equal")
-		require.Equal(t, c.addr.String(), addr.String(), "addr not equal")
+		var errStr string
+		network, u, err := ResolveEndpoint(context.Background(), c.input)
 		if err != nil {
-			require.ErrorContains(t, err, c.errStr)
+			errStr = err.Error()
+			if got, want := errStr, c.errStr; got != want {
```
magiconair: Could you change this to use `require` since I've removed `got, want` almost everywhere.

magiconair (changes requested):
Two tiny nitpicks and then we're ready to go. Should we call this v0.8.0 because of the breaking change for `ResolveEndpoint`?

uacp/conn.go
kung-foo: was going to fix this and noticed we don't use `DefaultDialTimeout` anywhere...
The `Config` + `Dialer` is a bit confusing/inconsistent. Will patch that up as well.

kung-foo: refactored a bit. should have less surprises when using the various options. testing timeouts is still a bit iffy since it's hard to test a dial or resolve timeout without a special server. but i'll save that for a re-work of the e2e testing.

kung-foo:
and yes, i'm fine with 0.8 due to the api change

uacp/conn.go
magiconair: Out of curiosity: wouldn't you just need to mock a `net.Conn` ?

kung-foo: can you mock a "syn->ack" where the ack never comes back? i guess I'm not clear exactly what the tcp dial timeout covers. i generally think of the `net.Conn` as a "high" level interface.

magiconair: Right. Well, that would be a fun exercise for a different time.

magiconair (approved):
LGTM

kung-foo:
well, that only took 5 years...

magiconair:
Better late than never 😁

PR #379, opened 2020-09-05
return most secure endpoint instead of least secure

### LINKED ISSUE #378: opcua.SelectEndpoint selects endpoint with lowest security level
Given the following two endpoints:
```
SecLevel    SecPolicy                                                 MsgSecMode
1           http://opcfoundation.org/UA/SecurityPolicy#Basic128Rsa15  MessageSecurityModeSign
2           http://opcfoundation.org/UA/SecurityPolicy#Basic128Rsa15  MessageSecurityModeSignAndEncrypt
```
and running

```go
opcua.SelectEndpoint(endpoints, "Basic128Rsa15", ua.MessageSecurityModeFromString("auto"))
``` 
will actually select the endpoint with the lowest security level instead of the one with the highest. This is because the endpoints are sorted on increasing security level, should be the other way around https://github.com/gopcua/opcua/blob/60a88cb3d109f2f394392ef6a3ac79bcdd745a35/client.go#L49

should be changed to `sort.Sort(sort.Reverse(bySecurityLevel(endpoints)))` 

(line 52 mentions that, in case multiple are possible, the one with the highest security is returned, which is not the case.)

coussej:
fixes #378

### DIFF
--- a/client.go
+++ b/client.go
@@ -46,7 +46,7 @@ func SelectEndpoint(endpoints []*ua.EndpointDescription, policy string, mode ua.
 		return nil
 	}
 
-	sort.Sort(bySecurityLevel(endpoints))
+	sort.Sort(sort.Reverse(bySecurityLevel(endpoints)))
 	policy = ua.FormatSecurityPolicyURI(policy)
 
 	// don't care -> return highest security level


magiconair (approved):
Thank you!

coussej:
It's a small contribution, but it's something :-)

kung-foo:
hmm, this will break existing code due to assumptions on the return value. For example, several of the examples in this repo used work out of the box with various simulators (since sims usually expose lots of different security configs). Now the examples fail because they don't (by default) require things like a cert or credz. I'm all in favor of raising the default security of the community in general, but is this a breaking change? I see that the docs said it returns the most secure. So maybe we just say this was a bug and now it's fixed?

magiconair:
Hmm. Didn’t think about that. Can we fix the examples that they default to the previous behavior?

In HTTP if also assume to start with mutual TLS. 

—
Frank Schröder

coussej:
@kung-foo Indeed, I noticed this 'bug' because the docs on the SelectEnpoint function explicitly mentions that the most secure endpoint will be returned:

image

magiconair:
Question is whether we should fix the docs or the behavior. Does the old behavior make sense in any way?

kung-foo:
The more I look at the API the more it seems broken.

The use case seems to be, "always select the most secure option". But in reality, since there are different requirements depending on the result, it should be "select the most secure option _given_ I have supplied...". i.e, does the caller have a user/pass? client certs?

The whole thing is predicated on a `SecurityLevel` byte which is used to sort. However, there isn't a `1-N` level, meaning as right now, it's `[0,1,2]`. So depending on how it's returned from the server, the sorting op isn't even stable since there duplicated `SecurityLevel`s.

Maybe we should mark it as deprecated (not that golang has anything like that) and remove it in some future release?

PR #380, opened 2020-09-05
Fix Sinumerik, avoid sending request with previous TokenId

unknownet:
Second part of __"fix for secure channel renew with opc sinumerik server"__ #371

Problem description __[left side]__:

The client send a request after the token renewal has been sent to the server.
the server can still receive response and use the first token to decode messages.
But It wont accept the request encoded with the new TokenId after the renewal request.

Resolution __[right side]__:

The client should stop sending the request after a renewal request has been sent. Once the new Token Id has been returned by the server, the client use it to send the pending requests and the future ones.

image

Implementation:

Use a Read/Write mutex to keep allowing multiple `SendRequestWithTimeout` while stopping them when tokenId renewal occurs.

image

### DIFF
--- a/uasc/secure_channel.go
+++ b/uasc/secure_channel.go
@@ -64,8 +64,8 @@ type SecureChannel struct {
 	activeInstance *channelInstance
 	instancesMu    sync.Mutex
 
-	// handles maps request IDs to response channels
-	handlers   map[uint32]chan *response
+	// handles maps request IDs to response callbacks
+	handlers   map[uint32]func(*response)
 	handlersMu sync.Mutex
 
 	// chunks maintains a temporary list of chunks for a given request ID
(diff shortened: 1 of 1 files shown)


unknownet:
Quick note, I've used `openingMu` for my implementation, I believe I should either rename it, as it doesn't serve exactly the same purpose or I should add another mutex _(not sure we still have use for openingMu)_.
Either way I haven't found a suitable name for it yet.

uasc/secure_channel.go
```diff
@@ -485,13 +480,31 @@ func (s *SecureChannel) open(ctx context.Context, instance *channelInstance, req
...
+	ch := make(chan error, 1)
+	err = s.sendAsyncWithTimeout(req, reqID, s.openingInstance, nil, true, s.cfg.RequestTimeout, func(r *response) {
+		if r.Err != nil {
+			ch <- r.Err
+			return
+		}
+		resp, ok := r.V.(*ua.OpenSecureChannelResponse)
 		if !ok {
-			return errors.Errorf("got %T, want OpenSecureChannelResponse", v)
+			ch <- errors.Errorf("got %T, want OpenSecureChannelResponse", r.V)
```
magiconair: Is there a standard status code that we can use in case this bubbles up to the caller?

magiconair:
Hi @unknownet, can we pick this up again? Is this still relevant?

magiconair:
Is this related to #398 ?

magiconair:
Closing b/c of age. @unknownet if you think there is still code in here which should be on `main` then please let us know.

PR #385, opened 2020-09-29
Handle subscription ACKs and republish requests

### LINKED ISSUE #369: Handle subscription ACKs from the server correctly
The short story is that we should keep track of the `PublishResponse`s ids we are ACK'ing and ignore if the server tells us to ACK a message twice.

This is a follow-up of #337 and the partial fix in #336.

magiconair:
The first attempt to handle subscription ACKs and republish
requests correctly. Not tested.

Fixes #369

### DIFF
--- a/client.go
+++ b/client.go
@@ -161,8 +161,8 @@ func NewClient(endpoint string, opts ...Option) *Client {
 		sessionCfg:     sessionCfg,
 		sechanErr:      make(chan error, 1),
 		subs:           make(map[uint32]*Subscription),
-		pausech:        make(chan struct{}, 1),
-		resumech:       make(chan struct{}, 1),
+		pausech:        make(chan struct{}, 2),
+		resumech:       make(chan struct{}, 2),
 		pendingAcks:    []*ua.SubscriptionAcknowledgement{},
 		publishTimeout: uasc.MaxTimeout,
 	}
@@ -175,20 +175,14 @@ func NewClient(endpoint string, opts ...Option) *Client {
 type reconnectAction uint8
 
 const (
-	// none, no reconnection action
-	none reconnectAction = iota
-	// createSecureChannel, recreate secure channel action
-	createSecureChannel
-	// restoreSession, ask the server to repair session
-	restoreSession
-	// recreateSession, ask the client to repair session
-	recreateSession
-	// republishSubscriptions, ask the server to repair the previous subscription
-	republishSubscriptions
-	// restoreSubscriptions moves subscriptions from one session to another
-	restoreSubscriptions
-	// abortReconnect, the reconnecting is not possible
-	abortReconnect
+	none reconnectAction = iota // no reconnection action
+
+	createSecureChannel   // recreate secure channel action
+	restoreSession        // ask the server to repair session
+	recreateSession       // ask the client to repair session
+	restoreSubscriptions  // republish or recreate subscriptions
+	transferSubscriptions // move subscriptions from one session to another
+	abortReconnect        // the reconnecting is not possible
 )
 
 // Connect establishes a secure channel and creates a new session.
@@ -285,7 +279,7 @@ func (c *Client) monitor(ctx context.Context) {
 
 					case ua.StatusBadSubscriptionIDInvalid:
 						// the subscription has been rejected by the server
-						action = restoreSubscriptions
+						action = transferSubscriptions
 
 					case ua.StatusBadCertificateInvalid:
 						// todo(unknownet): recreate server certificate
@@ -306,6 +300,12 @@ func (c *Client) monitor(ctx context.Context) {
 
 			c.pauseSubscriptions()
 
+			var (
+				subsToRepublish []uint32 // subscription ids for which to send republish requests
+				subsToRecreate  []uint32 // subscription ids which need to be recreated as new subscriptions
+				availableSeqs   map[uint32][]uint32
+			)
+
 			for action != none {
 
 				select {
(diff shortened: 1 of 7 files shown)


magiconair:
@alexbrdn @unknownet @kung-foo this is my first attempt at addressing the subscription ACKs and republish requests. I will try to test it on a real server and see if we can build an integration test for that as well. I'm grateful for some insight. This feels quite complex.

magiconair:
After reading Part 4, 5.13.5 Publish again I think the approach of having multiple publish loops is wrong since there is no guarantee that the publish responses from one loop will only contain responses from that subscription. If we keep that in mind then we can use the `Subscription` type to keep the state of the subscription but the actual publish loop would need to run in the client if there is at least one active subscription. Am I missing something?

> This Service is used for two purposes. First, it is used to acknowledge the receipt of NotificationMessages for one or more Subscriptions. Second, it is used to request the Server to return a NotificationMessage or a keep-alive Message. Since Publish requests are not directed to a specific Subscription, they may be used by any Subscription. 5.13.1.2 describes the use of the Publish Service

magiconair:
I think that would mean that we would no longer have to do this in an application but that the publish loop runs automatically by the client.

```go
go sub.Run(ctx) // start Publish loop
```

magiconair:
I think where this is an issue is that the `s.publishTimeout()` result is used when sending the `PublishRequest`s. However, this cannot really work if sending the publish requests is done for all subscriptions together. My gut feel is that using the `publishTimeout` is wrong here.

```go
	var res *ua.PublishResponse
	err := s.c.sendWithTimeout(req, s.publishTimeout(), func(v interface{}) error {
		return safeAssign(v, &res)
	})
```

```go
func (s *Subscription) publishTimeout() time.Duration {
	timeout := time.Duration(s.RevisedMaxKeepAliveCount) * s.RevisedPublishingInterval // expected keepalive interval
	if timeout > uasc.MaxTimeout {
		return uasc.MaxTimeout
	}
	if timeout < s.c.cfg.RequestTimeout {
		return s.c.cfg.RequestTimeout
	}
	return timeout
}
```

unknownet:
> @alexbrdn @unknownet @kung-foo this is my first attempt at addressing the subscription ACKs and republish requests. I will try to test it on a real server and see if we can build an integration test for that as well. I'm grateful for some insight. This feels quite complex.

I can probably run it on a Sinumerik 840D (production machine) next week, if It can help.

unknownet:
> After reading Part 4, 5.13.5 Publish again I think the approach of having multiple publish loops is wrong since there is no guarantee that the publish responses from one loop will only contain responses from that subscription. If we keep that in mind then we can use the `Subscription` type to keep the state of the subscription but the actual publish loop would need to run in the client if there is at least one active subscription. Am I missing something?
> 
> > This Service is used for two purposes. First, it is used to acknowledge the receipt of NotificationMessages for one or more Subscriptions. Second, it is used to request the Server to return a NotificationMessage or a keep-alive Message. Since Publish requests are not directed to a specific Subscription, they may be used by any Subscription. 5.13.1.2 describes the use of the Publish Service

I believe what you are saying make sens, If I recall correctly that's how it's implemented in the OpcUa ansi C Stack and in nodeJs [PublishEngine](https://github.com/node-opcua/node-opcua/blob/c9bfb4f020d87cf76bc96ba55a66e9cf7a67fe8f/packages/node-opcua-client/source/private/client_publish_engine.ts).

magiconair:
@alexbrdn when you first implemented the subscription code was there a reason you implemented it this way (one publish loop per subscription)? I feel like we're missing something.

magiconair:
The other question I have is when would we trigger a republish request? Is that only used when transferring a subscription or should we expect that a we may not get a notification and request it again if the server still has it? When should that happen?

magiconair:
Hmm, Part 4, 5.3.16.1 says the following which indicates that republishing is only relevant during reconnect

> See 6.7 for a description of the issues and strategies regarding reconnect handling and Republish

magiconair:
I have moved all subscription related code to `client_sub.go` and refactored the code to have a single publish loop for all subscriptions. The tests pass and the auto-reconnect works with our Siemens SIMATIC S7-1500. Doesn't seem to support `TransferSubscription` though. Need to check. 

I've also improved the debug logging a bit by adding a prefix debug logger to create more consistent log messages with less boilerplate. So I would encourage to test this.

magiconair:
I still need to review the republish and restore logic again. Will go through the todos tomorrow. 💤 now.

magiconair:
Rebased to latest `298-reconnect-reactivate`

alexbrdn:
> @alexbrdn when you first implemented the subscription code was there a reason you implemented it this way (one publish loop per subscription)? 

IIRC, the main idea was to allow starting multiple publish loops per subscription. One reason to start multiple publish loops is to reduce latency on slow network.

On the other hand, I fully agree, that publish loops don't have to "belong" to subscriptions. Multiple loops per client make more sense.

magiconair:
@alexbrdn are you on https://keybase.io?

alexbrdn:
No. Haven't heard of this one before.

I've sent you an email to take it off of here.

magiconair:
Since we are not getting any feedback on these branches I suggest to merge #368 and #385 to master and tag v0.2.0. At this point I'm OK with some issues since then we have a basis to work on. In the current state we are not making any progress.

Unless there are objections I'm going ahead with this early next week. I'll make sure we dogfood this at Northvolt as well.

CC @unknownet @kung-foo @alexbrdn

PR #389, opened 2020-10-05
Add option to override default dialer

antlad:
dialer.DialContext doesn't help when host is unreachable. So it is possible to wait on dial about minutes. Moving this config or dialer from top levels will touch to much places. So possible this workaround will help in some cases.  Sometimes it is helpfull to setup hard timeout via dialer config for dial attempt.


magiconair:
Hi @antlad, thank you for the PR. This is a good idea but I would like to avoid another global variable. I've provided a different approach in #432. Could you test if that solves your problem?

antlad:
@magiconair Thanks for notice this MR and your solution. Sadly only now I got time to test it. So I think no longer issues in main branch latest commit. Now it is stopping connect to unreachable host in time. No longer need for fork for us - so closing MR.

PR #440, opened 2021-06-21
Update subscription.go

barisvelioglu:
could not delete nodes by name from the subscription. Doesnt match by references so I put String() for comparison


magiconair:
Thank you!

Issue #444, opened 2021-07-14
ua.NewExtensionObject does not set the correct NodeID


siscia:
The call to `ua.NewExtensionObject` does not return a valid `ExtensionObject`.

The returned `ExtensionObject` contains a default `ObjectTypeID` (always set to zero).

The implementation of `ua.ExtensionObjectTypeID` (https://github.com/gopcua/opcua/blob/main/ua/extension_object.go#L117-L132) can be enhanced, for instance, by checking into the type register before to return the `TypeID`

PR #446, opened 2021-07-15
Set the NodeID to the ExtensionObjects to the correct value #444

siscia:
The NodeID of the ExtensionObjects was set to a default, zero, value.
Even in cases where the type was actually registered by the library.

### DIFF
--- a/ua/extension_object.go
+++ b/ua/extension_object.go
@@ -127,6 +127,9 @@ func ExtensionObjectTypeID(v interface{}) *ExpandedNodeID {
 	case *ServerStatusDataType:
 		return NewFourByteExpandedNodeID(0, id.ServerStatusDataType_Encoding_DefaultBinary)
 	default:
+		if id := eotypes.Lookup(v); id != nil {
+			return &ExpandedNodeID{NodeID: id}
+		}
 		return NewTwoByteExpandedNodeID(0)
 	}
 }
--- a/ua/extension_object.go
+++ b/ua/extension_object.go
@@ -15,7 +15,7 @@ var eotypes = NewTypeRegistry()
 // RegisterExtensionObject registers a new extension object type.
 // It panics if the type or the id is already registered.
 func RegisterExtensionObject(typeID *NodeID, v interface{}) {
-	if err := eotypes.Register(typeID.String(), v); err != nil {
+	if err := eotypes.Register(typeID, v); err != nil {
 		panic("Extension object " + err.Error())
 	}
 }
@@ -73,7 +73,7 @@ func (e *ExtensionObject) Decode(b []byte) (int, error) {
 		return buf.Pos(), body.Error()
 	}
 
-	typeID := e.TypeID.NodeID.String()
+	typeID := e.TypeID.NodeID
 	e.Value = eotypes.New(typeID)
 	if e.Value == nil {
 		return buf.Pos(), errors.Errorf("invalid extension object with id %s", typeID)
--- a/uatest/method_server.py
+++ b/uatest/method_server.py
@@ -16,7 +16,7 @@ def square(parent, variant):
 def sum_of_squares(parent, variant):
     # print("sum_of_square methos call with parameter: ", variant.Value)
     ret = 3*3 + 8*8
-    return [ua.Variant(ret, ua.VariantType.Boolean)]
+    return [ua.Variant(ret, ua.VariantType.Int64)]
 
 if __name__ == "__main__":
     server = Server()
--- a/uatest/method_server.py
+++ b/uatest/method_server.py
@@ -1,6 +1,24 @@
 #!/usr/bin/env python3
 
 from opcua import ua, Server
+from opcua.ua import uatypes
+
+class Complex(uatypes.FrozenClass):
+    ua_types = [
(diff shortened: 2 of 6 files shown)


siscia:
Should we add test for this?

siscia:
Test are not passing!

siscia:
Call to make fails with:

```
=== RUN   TestCallMethod
cryptography is not installed, use of crypto disabled
cryptography is not installed, use of crypto disabled
Endpoints other than open requested but private key and certificate are not set.
Listening on 0.0.0.0:4840
=== RUN   TestCallMethod/ns=2;s=main
=== RUN   TestCallMethod/ns=2;s=main#01
=== RUN   TestCallMethod/ns=2;s=main#02
    method_test.go:71: opcua: unsupported type: int
Exception raised while parsing message from client, closing
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/opcua/server/binary_server_asyncio.py", line 75, in _process_data
    ret = self.processor.process(hdr, buf)
  File "/usr/local/lib/python3.9/site-packages/opcua/server/uaprocessor.py", line 86, in process
    msg = self._connection.receive_from_header_and_body(header, body)
  File "/usr/local/lib/python3.9/site-packages/opcua/common/connection.py", line 315, in receive_from_header_and_body
    return self._receive(chunk)
  File "/usr/local/lib/python3.9/site-packages/opcua/common/connection.py", line 346, in _receive
    self._check_incoming_chunk(msg)
  File "/usr/local/lib/python3.9/site-packages/opcua/common/connection.py", line 290, in _check_incoming_chunk
    raise ua.UaError("Received chunk: {0} with wrong sequence expecting: {1}, received: {2}, spec says to close connection".format(chunk, self._peer_sequence_number, num))
opcua.ua.uaerrors._base.UaError: Received chunk: MessageChunk(Header(type:b'MSG', chunk_type:b'F', body_size:51, channel:6), SequenceHeader(SequenceNumber:7, RequestId:7 ), SymmetricAlgorithmHeader(TokenId:13 ), 39 bytes) with wrong sequence expecting: 5, received: 7, spec says to close connection
^CException ignored in: <module 'threading' from '/usr/local/Cellar/python@3.9/3.9.5/Frameworks/Python.framework/Versions/3.9/lib/python3.9/threading.py'>
Traceback (most recent call last):
  File "/usr/local/Cellar/python@3.9/3.9.5/Frameworks/Python.framework/Versions/3.9/lib/python3.9/threading.py", line 1428, in _shutdown
  ```

File: uatest/method_server.py [hunk 1]
```diff
@@ -13,13 +13,21 @@ def square(parent, variant):
     variant.Value *= variant.Value
     return [variant]
 
+def sum_of_squares(parent, variant):
+    # print("sum_of_square methos call with parameter: ", variant.Value)
+    ret = 3*3 + 8*8
+    return [ua.Variant(ret, ua.VariantType.Boolean)]
```
magiconair: The type should be `ua.VariantType.Int64` not `Boolean`

ua/extension_object.go
```diff
@@ -127,6 +127,16 @@ func ExtensionObjectTypeID(v interface{}) *ExpandedNodeID {
 	case *ServerStatusDataType:
 		return NewFourByteExpandedNodeID(0, id.ServerStatusDataType_Encoding_DefaultBinary)
 	default:
-		return NewTwoByteExpandedNodeID(0)
+		t := eotypes.Lookup(v)
```
magiconair: We should refactor the `TypeRegistry` to use `NodeID`s as keys so that we don't have to ignore the error below. I'll add a patch for this.

uatest/method_test.go
```diff
@@ -11,7 +11,14 @@ import (
 	"github.com/pascaldekloe/goe/verify"
 )
 
+type Complex struct {
+	i int
```
magiconair: We should not use `int` and `uint` since they are platform dependent. 

```go
type Complex struct {
    i,j int64
}
```

magiconair (changes requested):
The method returns a `Boolean` and the `Complex` type using `int` make the test fail. 

Additionally, I think we should refactor the `TypeRegistry` to guarantee that all keys are valid `NodeID`s.

magiconair (approved):

magiconair:
Fixed the test and added the change to refactor the type registry first. pls have a look

siscia:
Wonderful, thanks!

Maybe it make sense to implement the test correctly, what you think?

Ref: https://github.com/FreeOpcUa/python-opcua/issues/1364

magiconair:
What is wrong with the test?

siscia:
The result is hardcoded and not computed.

https://github.com/gopcua/opcua/pull/446/files#diff-078637d9373b2277a5d4cfbe1b2efc5fcc67b76c96df99916d627f952d6233e0R18

We are not actually reading the value from the object.

The byte array looks correct though!

magiconair:
ok, then change the test to actually compute the result but that should be enough.

siscia:
@magiconair I tried again to fix this, but with no success whatsoever.

If you want to take a look yourself it would be great.

PR #454, opened 2021-09-13
go1.17

magiconair:


kung-foo:
We've also just started testing with 1.17. No issues so far.

PR #456, opened 2021-09-14
fix policyURI=="" when secure channel's policy=none

lcw2:
Signed-off-by: lcw2 <wmy_lcw@163.com>
when secure channel's policy = none and policyURI =="", it would happen errors.


kung-foo:
LGTM. Thanks for the PR!

PR #457, opened 2021-09-15
the lock seems to be unused

siscia:


kung-foo:
@magiconair this seems to be left over from a refactor here: https://github.com/gopcua/opcua/commit/0ec335774d01128f8ca2ef45ff3bc3a04702679d

PR #462, opened 2021-09-17
Fix secure channel races (attempt 1)

magiconair:
This is the first attempt to address the data races in the secure channel. I am not fully comfortable with the code yet since I don't fully understand it yet. 

What I've concluded so far is that the `reset()` method is the main reason for the data races since it does a lot more than just resetting the `closing` and `disconnected` channels. It also resets other data structures which are guarded by other locks like the `instances`, `chunks` and the `handlers`. 

`reset()` is only called when the secure channel is closed and since the client implementation never re-uses an existing secure channel instance I don't think this method is required at all. 

Removing `reset()` immediately fixed all the data races we got with `make test-race` but generated a `panic: send on closed channel`.  This is because outstanding requests try to bubble up errors to the client via the `sechanErr` channel which the client closes in its `Close()` method. Since we are closing the client it seems OK to leave the `sechanErr` channel open to avoid this panic. However, we should try to shutdown all outstanding requests properly in the secure channel before returning. So I consider not closing the `sechanErr` a bit of a hack. 

Update: I think I've found the culprit for the `panic` and fixed it. 

See #376
See #460 
See #459

Issue #460, opened 2021-09-16
Data race on security channel timeout


antlad:
One more data race when when adding option opcua.Lifetime(someDuration)  when it is time to renew. Happening with Prosys simulator server, but not with Kepware.

```
==================
WARNING: DATA RACE
Write at 0x00c0004dc180 by goroutine 47:
  github.com/gopcua/opcua/uasc.(*SecureChannel).reset()
      xxx/vendor/github.com/gopcua/opcua/uasc/secure_channel.go:174 +0x25c
  github.com/gopcua/opcua/uasc.(*SecureChannel).Close.func1()
      xxx/vendor/github.com/gopcua/opcua/uasc/secure_channel.go:839 +0x5e
  github.com/gopcua/opcua/uasc.(*SecureChannel).Close()
      xxx/vendor/github.com/gopcua/opcua/uasc/secure_channel.go:847 +0x2d4
  github.com/gopcua/opcua.(*Client).monitor()
      xxx/vendor/github.com/gopcua/opcua/client.go:323 +0x1364

Previous read at 0x00c0004dc180 by goroutine 48:
  github.com/gopcua/opcua/uasc.(*SecureChannel).popHandler()
      xxx/vendor/github.com/gopcua/opcua/uasc/secure_channel.go:732 +0xcc
  github.com/gopcua/opcua/uasc.(*SecureChannel).sendRequestWithTimeout()
      xxx/vendor/github.com/gopcua/opcua/uasc/secure_channel.go:712 +0x4e5
  github.com/gopcua/opcua/uasc.(*SecureChannel).SendRequestWithTimeout()
      xxx/vendor/github.com/gopcua/opcua/uasc/secure_channel.go:760 +0x135
  github.com/gopcua/opcua.(*Client).sendWithTimeout()
      xxx/vendor/github.com/gopcua/opcua/client.go:764 +0xf5
  github.com/gopcua/opcua.(*Client).sendPublishRequest()
      xxx/vendor/github.com/gopcua/opcua/client_sub.go:512 +0x316
  github.com/gopcua/opcua.(*Client).publish()
      xxx/vendor/github.com/gopcua/opcua/client_sub.go:373 +0x18b
  github.com/gopcua/opcua.(*Client).monitorSubscriptions()
      xxx/vendor/github.com/gopcua/opcua/client_sub.go:355 +0x2c9

Goroutine 47 (running) created at:
  github.com/gopcua/opcua.(*Client).Connect.func1()
[...]

PR #465, opened 2021-09-22
protect with locks all the changes that may happen

siscia:
I believe we are addressing #460 

This is the wrong way of doing it, though!

Can somebody test it?


magiconair:
Let's coordinate this with the race fix

magiconair:
The race fix #462 removes the `reset` function altogether since a secure channel instance is never re-used. So we don't need this IMHO.

siscia:
All the problem is due to the `reset`!

Issue #392, opened 2020-10-21
why scheduleRenewal not start when security mode is None ?


ahuret:
Hi all !

We're implementing a prometheus exporter that scrap metrics from an opcua server and we faced an interesting issue:

For test purpose, we're using `Anonymous` Authentication mode and `None` Security mode.
Every ~ 1 hour, we had issue with connection. We found that this 1 hour correlate with default SecureChannel `Lifetime` setting.
But we'd still to get how to refresh connection until it's expired. We tested multiple stuff, custom renew implementation first (doesn't works), then we decided to use https://github.com/gopcua/opcua/tree/298-reconnect-reactivate branch, but still no reconnection.
We finally found https://github.com/gopcua/opcua/blob/298-reconnect-reactivate/uasc/secure_channel.go#L600 and scheduleRenewal that schedule the SecurityToken renewal and it seems this is exactly what we're looking for.

My question here is:
```go
if s.cfg.SecurityMode != ua.MessageSecurityModeNone {
	go s.scheduleRenewal(instance)
        ...
```

specifies that renewal only occurs when security mode is not None, but seems in our case we need it, why that decision then ?

Thanks a lot

jbguerraz:
Maybe @kung-foo would you have a clue ? Thank you a lot!

jbguerraz:
Can't find evidence of such condition in https://reference.opcfoundation.org/v104/Core/docs/Part4/5.5.2/

vereecw:
Hi!
We're also having this issue. Using v0.1.13 this issue is the most visible in the form of a lost connection.
Later versions seem to hide the problem when using the AutoReconnect option.
Removing the if statement mentioned above seems to fix the problem for us, so I did some digging in the OPC-UA documentation and this was what I found regarding MessageSecurityModeNone.

https://reference.opcfoundation.org/v104/Core/docs/Part4/5.5.2/#Table11
> A SecureChannel may have to be created even if the securityMode is NONE. The exact behaviour depends on the mapping used and is described in the OPC 10000-6.

https://reference.opcfoundation.org/v104/Core/docs/Part6/6.7.4/#:~:text=See%20OPC%2010000,or%20expired%20TokenIds.
[...]

PR #370, opened 2020-08-25
rework secure channel logic

magiconair:
Rebased #326 to master. Not yet squashed. 

@kung-foo please review and if OK squash it with a sensible commit message. 

Should we merge this to master or a release-branch? I'm thinking `master` since this is latest-and-greatest ... but I don't know whether this patch will break anything.

Fixes #324 
Closes #326

PR #466, opened 2021-09-23
Fix always schedule renewal of security token

vereecw:
Fix for issue https://github.com/gopcua/opcua/issues/392

Link to reason in OPC-UA documentation in comment in the issue.

We've also taken a look at the [C library for OPC-UA](https://github.com/open62541/open62541) and found no check for MessageSecurityMode.

[renewSecureChannel](https://github.com/open62541/open62541/blob/1c364d24d0aa635d0def3fe46078bb5526fb8f0f/src/client/ua_client_connect.c#L478)
[function call 1](https://github.com/open62541/open62541/blob/1c364d24d0aa635d0def3fe46078bb5526fb8f0f/src/client/ua_client.c#L200)
[function call 2](https://github.com/open62541/open62541/blob/1c364d24d0aa635d0def3fe46078bb5526fb8f0f/src/client/ua_client.c#L701)

### DIFF
--- a/uasc/secure_channel.go
+++ b/uasc/secure_channel.go
@@ -597,10 +597,8 @@ func (s *SecureChannel) handleOpenSecureChannelResponse(resp *ua.OpenSecureChann
 
 	debug.Printf("received security token: channelID=%d tokenID=%d createdAt=%s lifetime=%s", instance.secureChannelID, instance.securityTokenID, instance.createdAt.Format(time.RFC3339), instance.revisedLifetime)
 
-	if s.cfg.SecurityMode != ua.MessageSecurityModeNone {
-		go s.scheduleRenewal(instance)
-		go s.scheduleExpiration(instance)
-	}
+	go s.scheduleRenewal(instance)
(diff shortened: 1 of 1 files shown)


magiconair:
Hi @VereeckenWannes, I assume you've tested that on your setup. @kung-foo added that in #370 Maybe there was a specific reason.

magiconair:
Lets merge this and see if this breaks things.

kung-foo:
@magiconair honestly, 99% of my testing is against "secure" servers. I never did any lifetime/session testing against servers with a security mode of `none`.

IIRC I did have a patched version of a `node-opcua`-based server where I forced the session lifetime to something extremely short (i.e. 15 seconds). Maybe it's worth resurrecting something like that for the integration tests.

vereecw:
@magiconair yes, we tested in on multiple setups and this change fixed our issue with losing our opcua connection after the lifetime expires when SecurityMode is set to none.

magiconair:
@kung-foo that sounds like a good idea. @VereeckenWannes thank you.

PR #467, opened 2021-09-23
Add support for ByteArray

### LINKED ISSUE #463: Error writing byte(uint8) arrays. Wrong array length.
Hello,
I read some Data from server and get a variant with this properties:
  ```
  mask: 131
  arrayLength: 54
  arrayDimensionsLength: 0
  arrayDimensions: nil
  value: interface{} ([]uint8) [......]
  ```
If I try to write this data back
  ```
  var buf []byte = make([]byte, 54)
  buf = (*(*[54]byte)(unsafe.Pointer(&wrVal)))[:]
  v, err := ua.NewVariant(buf)
```
At this point the new variant has this properties:
```
  mask: 15
  arrayLength: 0
  arrayDimensionsLength: 0
  arrayDimensions: nil
  value: interface{} ([]uint8) [......]
```
I get: "The value supplied for the attribute is not of the same type as the attribute's value. StatusBadTypeMismatch (0x80740000)"

This issue is caused by variang.go (row: 412-415)

magiconair:
This patch adds support for the ByteArray type in a Variant.
This allows the caller to send a []byte as an array of Byte instead of a
length encoded ByteString which is the default.

Fixes #463

/cc @SkyFox277

### DIFF
--- a/ua/variant.go
+++ b/ua/variant.go
@@ -25,6 +25,9 @@ const (
 	VariantArrayValues = 0x80
 )
 
+// ByteArray encodes a byte array as an OPC/UA array of Byte.
+type ByteArray []byte
+
 // Variant is a union of the built-in types.
 //
 // Specification: Part 6, 5.1.2 Table 1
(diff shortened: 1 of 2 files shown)


kung-foo (approved):
LGTM. I suppose _some_ of this can be made cleaner with generics. On the other hand, we could also just code-gen constructors for all the opc types...

PR #468, opened 2021-09-23
Make dial timeout configurable

### LINKED ISSUE #464: Allow Timeout when Listing Endpoints/Dial
Hi and thanks for the amazing library!

When my application tries to List all Endpoints of a specific server URL that server doesn't exists, it freezes for some time until I receive a 'connection timed out' error.

Is it possible to configure a Timeout so I won't have to wait for too long?

I Tried to create a new Client with `SessionTimeout` and `RequestTimeout` options and then calling Dial, but it ignored those values. 

How it Is:

```go
opcua.GetEndpoints(server.URL)
```

How I propose it could be:

```go
opts := []opcua.DialOption{
    opcua.Timeout(time.Seconds * 2),
}
opcua.GetEndpoints(server.URL, opts...)
```

magiconair:
This patch makes the `uacp.Dialer` configurable and allows to configure the dial timeout for establishing the uucp connection.

Fixes #464

### DIFF
--- a/config.go
+++ b/config.go
@@ -54,7 +54,7 @@ func DefaultSessionConfig() *uasc.SessionConfig {
 // Config contains all config options.
 type Config struct {
 	dialer      *uacp.Dialer
-	dialTimeout time.Duration
+	dialTimeout *time.Duration
 	sechan      *uasc.Config
 	session     *uasc.SessionConfig
 }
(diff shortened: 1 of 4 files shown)


magiconair:
I think my biggest concern with this change is that it enforces a dial timeout and that will be non-zero since it defaults to the request timeout. There is no way to have no dial timeout. This is different from before.

config.go
```diff
@@ -428,7 +455,22 @@ func AuthIssuedToken(tokenData []byte) Option {
...
+
+// Dialer sets the uacp.Dialer to establish the connection to the server.
+func Dialer(d *uacp.Dialer) Option {
+	return func(cfg *Config) {
+		cfg.dialer = d
+	}
+}
+
+// DialTimeout sets the timeout for establishing the UACP connection.
+// Defaults to DefaultDialTimeout. Set to zero for no timeout.
```
kung-foo, line 471: Does `DefaultDialTimeout` exist?

magiconair:
I'll merge this and then we start testing this on our systems.

Issue #395, opened 2020-10-26
3K nodes monitoring - timeouts/broken pipe issues


axlev:
Hi,

I have a setup with 3K nodes monitored every 1 sec. It is working well with open62541 based client (C language library). However, when I am trying to use gopcua, I am experiencing the below:
1) When trying to subscribe in single monitor,  then I see a timeout
2) splitting in different monitors each handling 50-500 nodes will be working, but from time to time experience a "broken pipe" error reported requiring to reestablish the connection.

Q1:
Do you have any best practices to apply in this case?    

Q2:
What could be probably wrong with my implementation?

Thanks
Alex

alexlev2004:
Just observed the Error: Bad Sequence returned by OPC server...
And indeed it can be seen from debug messages that 6392 comes after 6393:
 
Oct 28 16:07:47 halo-OptiPlex-7040 halodigital/opcua-client:0.45/opcua-client-0.45/86ecd1b465c7[2716]: debug: uasc 1/6387: recv MSGF with 7982 bytes
Oct 28 16:07:47 halo-OptiPlex-7040 halodigital/opcua-client:0.45/opcua-client-0.45/86ecd1b465c7[2716]: debug: uasc 1/6391: send *ua.PublishRequest with 74 bytes
Oct 28 16:07:47 halo-OptiPlex-7040 halodigital/opcua-client:0.45/opcua-client-0.45/86ecd1b465c7[2716]: debug: uasc 1/6387: recv *ua.PublishResponse
Oct 28 16:07:47 halo-OptiPlex-7040 halodigital/opcua-client:0.45/opcua-client-0.45/86ecd1b465c7[2716]: debug: sending *ua.PublishResponse to handler
Oct 28 16:07:47 halo-OptiPlex-7040 halodigital/opcua-client:0.45/opcua-client-0.45/86ecd1b465c7[2716]: debug: conn 1: recv MSGF with 7220 bytes
Oct 28 16:07:47 halo-OptiPlex-7040 halodigital/opcua-client:0.45/opcua-client-0.45/86ecd1b465c7[2716]: debug: uasc 1/6382: recv MSGF with 7220 bytes
Oct 28 16:07:47 halo-OptiPlex-7040 halodigital/opcua-client:0.45/opcua-client-0.45/86ecd1b465c7[2716]: debug: uasc 1/6382: recv *ua.PublishResponse
[...]

PR #473, opened 2021-09-29
Client chunking

diericd:
Hi,

This I my implementation of client to server chunking. It has been tested against multiple OPC servers in plaintext and encryption. This should fix issue #395

### DIFF
--- a/uasc/secure_channel_instance.go
+++ b/uasc/secure_channel_instance.go
@@ -33,6 +33,7 @@ type channelInstance struct {
 	securityTokenID uint32
 	sequenceNumber  uint32
 	algo            *uapolicy.EncryptionAlgorithm
+	maxBodySize     uint32
 
 	messagesSent uint32
 	// messagesReceived uint32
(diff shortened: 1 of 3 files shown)


uasc/secure_channel_instance.go
```diff
@@ -33,11 +33,12 @@ type channelInstance struct {
 	securityTokenID uint32
 	sequenceNumber  uint32
 	algo            *uapolicy.EncryptionAlgorithm
+	maxBodySize     uint32
 
-	messagesSent uint32
-	// messagesReceived uint32
 	bytesSent uint64
 	// bytesReceived    uint64
+	messagesSent uint32
```
magiconair: I assume that you've reordered these to align for `atomic.Load/Store`. We should mark that with a comment and maybe even move them to the top to make this clear. Not sure how well `go vet` catches these alignment issues these days. In that case we may want to make the `sync.Mutex` explicit, e.g. `mu sync.Mutex`.

magiconair:
@diericd Thank you for this. I'm sure this will make some people very happy!

uasc/secure_channel_instance.go
diericd: Yes, it was to fix issues when running on arm32. Running "GOARCH=arm GOARM=6 go vet" did not catch this issue.

magiconair: I usually add a `// atomic.Load/Store` as a comment and if the alignment is important then maybe point that out as well. Problem with putting it at the end is that every new field might break it.

PR #475, opened 2021-10-05
Introduce test for Panic

siscia:
If we create a client and we close it immediately after, without actually connect it, the program panics.


siscia:
This was in the process to debug another issue in the codebase.

PR #476, opened 2021-10-05
Use ERN formula instead of OPCUA formula

diericd:
Hi, 

I've encountered an edge case where the encrypted chunk size is a tad to big when the maximum receive buffer isn't aligned to the algorithm block size.

I've replaced the calculation for the maximum body size of a chunk to a more conservative one (in line with the current implementation of node-opcua)

### DIFF
--- a/uasc/message.go
+++ b/uasc/message.go
@@ -5,6 +5,8 @@
 package uasc
 
 import (
+	"math"
+
 	"github.com/gopcua/opcua/errors"
 	"github.com/gopcua/opcua/ua"
 )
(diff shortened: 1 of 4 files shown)


diericd:
Apparently the calculation for the bodysize was not the issue. 
The bug was against a prosys simulation server which sets the maximum chunksize to 8196. This results in a maximum body size of 8135. But the way padding was set up another full blocksize of padding was added to the data resulting in a final chunksize of 8208.  My fix was inspired by the node-opcua library.

magiconair:
@diericd thank you. Could you move the calculation to a function and provide a test for the edge cases?

magiconair:
@diericd Pinging again since I'd like to include this in the v0.2.0 release on Monday

PR #477, opened 2021-10-05
add test for reading an unknow type

siscia:
I was expecting this test to just fail.

It actually panics.


magiconair:
Superseded by #481

PR #481, opened 2021-10-11
Handle reads with unknown extension objects

### LINKED ISSUE #480: Read request fails with unknown extension object
When a ReadRequest returns an extension object unknown to the client the entire read request fails with an `invalid extension object with id nnn` error. Depending on the usage this causes the client to disconnect or crash. 

Since the extension object was OK from the server's point of view the status in the response is OK. 

The OPC/UA specification specifies some status codes for the results but none of them seems to fit. 

https://reference.opcfoundation.org/v104/Core/docs/Part4/5.10.2/#5.10.2.4

I suggest to set the status code to `StatusBadDataTypeIDUnknown` since that is what has happened.

magiconair:
This patch sets the status code for a `DataValue` in a `ReadResponse` to `StatusBadDataTypeIDUnknown` when the value is an extension object but could not be decoded because it is unknown to the client.

This patch also replaces the test in #477 since I could not overwrite it.

Fixes #480
Closes #477

### DIFF
--- a/client.go
+++ b/client.go
@@ -799,7 +799,23 @@ func (c *Client) Read(req *ua.ReadRequest) (*ua.ReadResponse, error) {
 
 	var res *ua.ReadResponse
 	err := c.Send(req, func(v interface{}) error {
-		return safeAssign(v, &res)
+		err := safeAssign(v, &res)
+
+		// If the client cannot decode an extension object then its
+		// value will be nil. However, since the EO was known to the
+		// server the StatusCode for that data value will be OK. We
+		// therefore check for extension objects with nil values and set
+		// the status code to StatusBadDataTypeIDUnknown.
+		if err == nil {
+			for _, dv := range res.Results {
+				val := dv.Value.Value()
+				if eo, ok := val.(*ua.ExtensionObject); ok && eo.Value == nil {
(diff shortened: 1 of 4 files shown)


client.go
```diff
@@ -799,7 +799,24 @@ func (c *Client) Read(req *ua.ReadRequest) (*ua.ReadResponse, error) {
...
+
+		// If the client cannot decode an extension object then its
+		// value will be nil. However, since the EO was known to the
+		// server the StatusCode for that data value will be OK. We
+		// therefore check for extension objects with nil values and set
+		// the status code to StatusBadDataTypeIDUnknown.
+		if err == nil {
+			for _, dv := range res.Results {
+				val := dv.Value.Value()
+				fmt.Printf("read resp val: %#v\n", val)
```
kung-foo: should this go out via `debug.Logger(...)`?

magiconair: no this shouldn't be there at all.

magiconair:
Tests are still hanging. It worked when I pushed the fix. Looking.

magiconair:
Tests fail because we use the `ComplexType` in more than one test and then it isn't unknown anymore.

PR #487, opened 2021-10-21
add tests for reading basic arrays

siscia:


kung-foo:
@siscia I fixed the build issue in #489. Can you please rebase? That should allow your PR to pass.

PR #493, opened 2021-11-05
fix: return nil on subscription.Delete

henripqt:
First of all thank you for creating/maintaining this library.
I currently work for a company with a lot of clients in the industry field and we're planning to use OPCUA for various projets using Go. Therefore we'll try to contribute as much as we can according to our use cases.

# ISSUE:

When calling `Unsubscribe` on a subscription we expect an error.
However at the moment if the subscription client returns no error (err == nil) the `Unsubscribe` method will return `OK 0x0` instead of nil. 

## STEPS TO REPRODUCE:

- Create a node monitor
- Create a subscription from the monitor
- Call `subscription.Unsubscribe()` and check for the error
- If `Unsubscribe` is successful then we should receive an error of `OK 0x0` instead of nil

## SOLUTION:

- Return nil from the `subscription.delete` method if `s.c.Send == nil ` || `s.c.Send == ua.StatusOK`

### DIFF
--- a/subscription.go
+++ b/subscription.go
@@ -88,18 +88,20 @@ func (s *Subscription) delete() error {
 	req := &ua.DeleteSubscriptionsRequest{
 		SubscriptionIDs: []uint32{s.SubscriptionID},
 	}
+
 	var res *ua.DeleteSubscriptionsResponse
 	err := s.c.Send(req, func(v interface{}) error {
 		return safeAssign(v, &res)
 	})
+
 	switch {
-	case err == ua.StatusOK:
-		s.items = make(map[uint32]*monitoredItem)
-		return nil
 	case err != nil:
 		return err
+	case res.Results[0] == ua.StatusOK:
+		s.items = make(map[uint32]*monitoredItem)
+		return nil
 	default:
-		return res.ResponseHeader.ServiceResult
+		return res.Results[0]
 	}
 }
 


magiconair:
Hey @henripqt, thank you and welcome to the team. Looking forward to your contributions. 

This looks good to me and I'll merge it on Monday when we release v0.2.0.

magiconair:
I think this fixes the issue in the wrong way since the real issue is that we don't return the result of the `DeleteSubscription` operation. Could you test this patch instead please?

```go
// delete removes the subscription from the server.
func (s *Subscription) delete() error {
	req := &ua.DeleteSubscriptionsRequest{
		SubscriptionIDs: []uint32{s.SubscriptionID},
	}
	var res *ua.DeleteSubscriptionsResponse
	err := s.c.Send(req, func(v interface{}) error {
		return safeAssign(v, &res)
	})

	switch {
	case err != nil: 
		return err
	case res.Results[0] != ua.StatusOK:
		return res.Results[0]
	default:
		s.items = make(map[uint32]*monitoredItem)
		return ua.StatusOK
	}
}
```

Depending on how quickly you respond we may be able to include this in the v0.2.0 release today. Otherwise, it will go into v0.2.1

magiconair:
https://reference.opcfoundation.org/Core/docs/Part4/5.13.8/

henripqt:
Hello @magiconair. Thanks for pointing this out, I'll give it a shot this afternoon. 

However with your solution in place we'll still get a `ua.StatusOK` returned from the `subscription.delete()` method and therefore from the `subscription.Cancel` method.

Returning an `ok` status as an error isn't really idiomatic golang as the package's user will most likely expect the returned error to be `nil` in case everything went well.

This could potentially lead to logging an `OK 0x0` message with an error level or force the end user to always check if the received error isn't of type `statusOK` before handling the error

As per the documentation you've provided

```
OPC UA Servers should return good/success StatusCodes if the operation completed normally and the result is always valid. Different StatusCode values can provide additional information to the Client.
```

I do understand that we therefore need to retain the status code and propagate it but could we consider returning the status along with the error ? That way we can still check if err == nil if we simply want to ensure the operation was successful without having to check if the error ==  `ua.StatusOK` but still retain the status if one requires to do additional checkup.

The method signature would therefore be something like the following :

```go
// Cancel stops the subscription and removes it
// from the client and the server.
func (s *Subscription) Cancel() (statusCode ua.StatusCode, err error) {
	...
}

// delete removes the subscription from the server.
func (s *Subscription) delete() (statusCode ua.StatusCode, err error) {
	...
}
```

and the end user could simply decide to :

```go
if _, err := s.Cancel(); err != nil {
   logger.Errorf("error unsubscribing : %v, err)
}
```

or 

```go
statusCode, err := s.Cancel();

if err != nil {
   logger.Errorf("error unsubscribing : %v - received status of %v", err, statusCode)
}

// do whatever with status code
```

If not then we can consider updating the documentation of the Cancel method to indicate that the error potentially signifies that the operation is successful and that one needs to check if it isn't equal to ua.StatusOK.
What do you think ?

magiconair:
Yes, you're right. Try this then:

```go
// delete removes the subscription from the server.
func (s *Subscription) delete() error {
	req := &ua.DeleteSubscriptionsRequest{
		SubscriptionIDs: []uint32{s.SubscriptionID},
	}
	var res *ua.DeleteSubscriptionsResponse
	err := s.c.Send(req, func(v interface{}) error {
		return safeAssign(v, &res)
	})

	switch {
	case err != nil: 
		return err
	case res.Results[0] == ua.StatusOK:
		s.items = make(map[uint32]*monitoredItem)
		return nil
	default:
		return res.Results[0]
	}
}
```

henripqt:
@magiconair Done.

I took the liberty to add the `err == nil` case to your code as a safe guard. 
Tracking down the flow of the `Send` method I do not see a case where that could happen (unless the send handler would be nil) but I believe it's a good indication for the reader that `ua.StatusOK` means there was indeed no errors and, as there is currently no unit test for subscription, it might prevent seeing this issue appearing again if some changes are made down the line in a way that the Send method would return a nil error. Let me know if that works for you

magiconair:
This is wrong since the comma means OR and not AND if I'm not mistaken and `err == nil` is guaranteed at this point. Otherwise, the previous condition would trigger. Can you fix and squash the commits? Then I'll merge that today.

henripqt:
@magiconair You're absolutely right ! I've updated the branch accordingly and squashed the commits. Should be good to go. Thanks for your availability :)

magiconair (approved):

PR #497, opened 2021-11-18
subscription: do not fail Monitor() on bad node (#491)

### LINKED ISSUE #491: Refactor Needed: Subscription.Monitor()
In [Subscription.Monitor()](https://github.com/gopcua/opcua/blob/6e89091fa1796dc72a5dcbab39e4082522cf43c7/subscription.go#L107), if one of the monitored items returns bad status code, the function just [returns there](https://github.com/gopcua/opcua/blob/6e89091fa1796dc72a5dcbab39e4082522cf43c7/subscription.go#L126), discarding all of the other items even if their status code is good. This creates conflict in the state of subscription present in server and subscription stored in client. Since server will store the good items discarding the bad ones, while client discards all the items if just one is bad.
The server will continue to send the notifications for the good items it stored, but the client is now receiving notifications for items that don't exist in the client's subscription.

My suggestion would be that the Subscription.Monitor() stores the good items while discarding the bad ones, and it should return the MonitoredItemsResponse if atleast one item is good.

If any error occurs which is not related to individual items like `serverDisconnected` etc, return `nil, error`.
In all other cases, return `res, nil`.

magiconair:
This patch restores the behavior of the Monitor() call to the v0.1
behavior of only returning an error if the Monitor call itself failed.

Fixes #491

### DIFF
--- a/subscription.go
+++ b/subscription.go
@@ -122,12 +122,6 @@ func (s *Subscription) Monitor(ts ua.TimestampsToReturn, items ...*ua.MonitoredI
 		return nil, err
 	}
 
-	for _, result := range res.Results {
-		if status := result.StatusCode; status != ua.StatusOK {
-			return nil, status
-		}
-	}
-
 	// store monitored items
 	// todo(fs): should we guard this with a lock?
 	for i, item := range items {


magiconair:
@salmanazmat666 can you have a look as well?

magiconair:
Our internal test shows that this fixes our own problems.

magiconair:
Merging

Issue #506, opened 2021-11-19
Panic on Browse with nil NodeID


henripqt:
### ISSUE:

When attempting to trigger a `client.Browse` call I'm getting a panic from the `ua.NodeID.Encode` method in cases where `NodeID == nil`

### STEPS TO REPRODUCE:

- Install & start [OPCUA Simulation Server](https://downloads.prosysopc.com/opc-ua-simulation-server-downloads.php)
- Run the following 

```go
package main

import (
	"context"

	"log"

	"github.com/gopcua/opcua"
	"github.com/gopcua/opcua/ua"
)

func main() {
	c, err := newClient(context.Background())
	if err != nil {
		panic(err)
	}

	defer c.Close()

	nsIDs := []string{"ns=3;s=Random_cfg", "ns=3;i=1003", "ns=3;i=1004", "ns=3;i=1005"}
	var browserDescs []*ua.BrowseDescription

	for _, nodeID := range nsIDs {
		nID, err := ua.ParseNodeID(nodeID)
		if err != nil {
			log.Println(err)
			continue
		}

		browserDescs = append(browserDescs, &ua.BrowseDescription{
			NodeID:          nID,
			BrowseDirection: ua.BrowseDirection(ua.BrowseDirectionBoth),
			IncludeSubtypes: false,
			NodeClassMask:   1,
			ResultMask:      1,
		})
	}

	_, err = c.Browse(&ua.BrowseRequest{NodesToBrowse: browserDescs})

	if err != nil {
		log.Println(err)
	}

	select {}
}

// newClient creates a new client and connects to the endpoint.
func newClient(ctx context.Context) (*opcua.Client, error) {
	client := opcua.NewClient("opc.tcp://127.0.0.1:4840/OPCUA/SimulationServer")

	if err := client.Connect(ctx); err != nil {
		return nil, err
	}

	return client, nil
}
```

The program will panic at the following line : 

```go
func (n *NodeID) Encode() ([]byte, error) {
	buf := NewBuffer(nil)
	buf.WriteByte(byte(n.mask)) // <---- Panic here. n == nil
// [...]
```
If I look over the `ExtensionObject.Encode` method we see that we have a safeguard for such cases

```go
func (e *ExtensionObject) Encode() ([]byte, error) {
	buf := NewBuffer(nil)
	if e == nil { // <---- Safeguard
[...]

PR #507, opened 2021-11-22
Encode nil node id as zero

magiconair:
This patch encodes a `nil` NodeID as a two byte zero node id.

See #506

### DIFF
--- a/ua/node_id.go
+++ b/ua/node_id.go
@@ -405,6 +405,12 @@ func (n *NodeID) Decode(b []byte) (int, error) {
 }
 
 func (n *NodeID) Encode() ([]byte, error) {
+	// https://github.com/gopcua/opcua/issues/506
+	// encode 'nil' node ids as two byte zero values
+	if n == nil {
+		return []byte{0, 0}, nil
+	}
+
 	buf := NewBuffer(nil)
 	buf.WriteByte(byte(n.mask))
 
--- a/ua/node_id_test.go
+++ b/ua/node_id_test.go
@@ -12,6 +12,7 @@ import (
 	"testing"
 
 	"github.com/gopcua/opcua/errors"
+	"github.com/pascaldekloe/goe/verify"
 )
 
 func TestNodeID(t *testing.T) {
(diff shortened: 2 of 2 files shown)


magiconair:
Pinging the other OPC/UA experts on the team about this change.

kung-foo:
not sure... seems like encoding a `nil` node id _should_ panic. In other words, I would be _surprised_ when it _didn't_ panic since a `nil` node seems like an error on my part.

However, I see this:

> A null NodeId has special meaning. For example, many services defined in OPC 10000-4 define special behaviour if a null NodeId is passed as a parameter. Each IdType has a set of identifier values that represent a null NodeId. These values are summarised in Table 24.

https://reference.opcfoundation.org/v104/Core/docs/Part3/8.2.4/

So maybe there is a valid reason. But in that case, I'd rather have a `NewNullNodeID()` func.

magiconair:
Having a function still means that you have to set it everywhere and `Encode` would still panic if you forget it. What would be the benefit of the function?

kung-foo:
It's the surprise factor. I'd be surprised if it didn't panic, and I might be surprised if it returned unexpected data because there was a special "null" node id generated (unexpectedly) behind the scenes. I just don't like surprises (birthday parties excepted). Where would we document that _any where_ and for _any reason_ a node id is nil, it will still (sorta) work? I would not expect a user to dig in and look at `NodeID.Encode`...

magiconair:
Since the concept of a `nil` NodeID doesn't exist and is sent as zero value I think this should be in the `Encode` method. This is how the protocol works and how we translate it to Go. TBH, I always considered it odd that I have to provide the zero value explicitly if all I want to do is to send the default and was usually surprised by the `panic`. Yes, setting the zero value is more explicit but it also feels like initialising a struct with `0` and `false` which you wouldn't do in Go.

What I'm struggling with is whether this should stop at the `NodeID`. The `BrowseRequest` for example also has an optional `ViewDescription`. Why do I have to provide a `&ua.ViewDescription{ViewID: ua.NewTwoByteNodeID(0)}`if all I want is to omit it? That's also exactly the thing the original author was struggling with and what the defaults helper functions provide. Maybe optional fields should be serialised to their zero value when set to `nil`.

magiconair:
But it is also this inconsistency which makes me a bit reluctant about this PR in particular since we should either address all optional fields or none of them but not just a select few.

magiconair:
Lets park this one.

magiconair:
I'll close this one since there is no agreement.

PR #508, opened 2021-11-22
client: set sensible defaults for Browse

magiconair:
This patch sets sensible defaults for the `Browse` method of the client.

See #506

### DIFF
--- a/client.go
+++ b/client.go
@@ -847,8 +847,36 @@ func (c *Client) Write(req *ua.WriteRequest) (*ua.WriteResponse, error) {
 	return res, err
 }
 
+func cloneBrowseRequest(req *ua.BrowseRequest) *ua.BrowseRequest {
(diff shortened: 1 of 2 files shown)


kung-foo:
These "sane defaults" are fine, however does this mean we should really do this for _all_ *Requests.

kung-foo:
LGTM once merge conflict resolved.

kung-foo (approved):

magiconair:
Well, we did this for `Read` and we have `client.Browse` and `node.Browse` which is confusing. I think other libraries solve this by providing defaults in arguments or helper methods. I'm OK with solving this in another way if you have a suggestion.

kung-foo:
I guess the issue is that "the struct is the api". Since it is unlikely that we want to make everything private (i.e. _only_ make a `BrowseRequest` instance via builders/helpers) then we need strong, public validators and/or implicit defaults.

PR #512, opened 2021-11-24
ua: add support for resolving namespace urls

magiconair:
This patch adds support for resolving node ids with namespace URLs (`nsu=`). Since I don't want to break the `ParseNodeID()` function I've created `ResolveNodeID()` function which takes a resolver for the lookup. 

If we are OK with breaking `ParseNodeID()` then we can roll this into one function.

### DIFF
--- a/client.go
+++ b/client.go
@@ -126,6 +126,9 @@ type Client struct {
 	// state of the client
 	state atomic.Value // ConnState
 
+	// list of cached namespaces on the server
+	namespaces atomic.Value // []string
+
 	// monitorOnce ensures only one connection monitor is running
 	monitorOnce sync.Once
 
@@ -158,6 +161,7 @@ func NewClient(endpoint string, opts ...Option) *Client {
 	c.publishTimeout.Store(uasc.MaxTimeout)
 	c.pauseSubscriptions(context.Background())
 	c.state.Store(Closed)
+	c.namespaces.Store([]string{})
 	return &c
 }
 
@@ -185,13 +189,15 @@ func (c *Client) Connect(ctx context.Context) (err error) {
 	if err := c.Dial(ctx); err != nil {
 		return err
 	}
+
 	s, err := c.CreateSession(c.cfg.session)
 	if err != nil {
-		_ = c.Close()
+		c.Close()
 		return err
 	}
+
 	if err := c.ActivateSession(s); err != nil {
-		_ = c.Close()
+		c.Close()
 		return err
 	}
 	c.state.Store(Connected)
@@ -202,6 +208,13 @@ func (c *Client) Connect(ctx context.Context) (err error) {
 		go c.monitor(mctx)
 		go c.monitorSubscriptions(mctx)
 	})
+
+	if !c.cfg.disableNamespaceUpdate {
+		if err := c.UpdateNamespaces(); err != nil {
+			c.Close()
+			return err
+		}
+	}
 
 	return nil
 }
@@ -489,11 +502,16 @@ func (c *Client) Dial(ctx context.Context) error {
 
 	c.sechan, err = uasc.NewSecureChannel(c.endpointURL, c.conn, c.cfg.sechan, c.sechanErr)
 	if err != nil {
-		_ = c.conn.Close()
+		c.conn.Close()
 		return err
 	}
 
-	return c.sechan.Open(ctx)
+	if err := c.sechan.Open(ctx); err != nil {
+		c.conn.Close()
+		return err
+	}
+
+	return nil
 }
 
 // Close closes the session and the secure channel.
@@ -527,6 +545,11 @@ func (c *Client) Close() error {
 
 func (c *Client) State() ConnState {
 	return c.state.Load().(ConnState)
+}
+
+// Namespaces returns the currently cached list of namespaces.
+func (c *Client) Namespaces() []string {
+	return c.namespaces.Load().([]string)
 }
 
 // Session returns the active session.
@@ -958,6 +981,31 @@ func (c *Client) HistoryReadRawModified(nodes []*ua.HistoryReadValueID, details 
 		return safeAssign(v, &res)
 	})
 	return res, err
+}
+
+// NamespaceArray returns the list of namespaces registered on the server.
+func (c *Client) NamespaceArray() ([]string, error) {
+	node := c.Node(ua.NewNumericNodeID(0, id.Server_NamespaceArray))
+	v, err := node.Value()
+	if err != nil {
+		return nil, err
+	}
+
+	ns, ok := v.Value().([]string)
+	if !ok {
+		return nil, errors.Errorf("error fetching namespace array. id=%d, type=%T", v.Type(), v.Value())
+	}
+	return ns, nil
+}
+
+// UpdateNamespaces updates the list of cached namespaces from the server.
+func (c *Client) UpdateNamespaces() error {
+	ns, err := c.NamespaceArray()
+	if err != nil {
+		return err
+	}
+	c.namespaces.Store(ns)
+	return nil
 }
 
 // safeAssign implements a type-safe assign from T to *T.
--- a/client.go
+++ b/client.go
@@ -209,11 +209,13 @@ func (c *Client) Connect(ctx context.Context) (err error) {
 		go c.monitorSubscriptions(mctx)
 	})
 
-	if !c.cfg.disableNamespaceUpdate {
-		if err := c.UpdateNamespaces(); err != nil {
-			c.Close()
-			return err
-		}
+	// todo(fs): we might need to guard this with an option in case of a broken
+	// todo(fs): server. For the sake of simplicity we left the option out but
+	// todo(fs): see the discussion in https://github.com/gopcua/opcua/pull/512
+	// todo(fs): and you should find a commit that implements this option.
+	if err := c.UpdateNamespaces(); err != nil {
+		c.Close()
+		return err
 	}
 
 	return nil
--- a/ua/node_id.go
+++ b/ua/node_id.go
@@ -9,8 +9,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
-	"strconv"
-	"strings"
 
 	"github.com/gopcua/opcua/errors"
 )
@@ -94,90 +92,25 @@ func MustParseNodeID(s string) *NodeID {
 // ParseNodeID returns a node id from a string definition of the format
 // 'ns=<namespace>;{s,i,b,g}=<identifier>'.
 //
-// For string node ids the 's=' prefix can be omitted.
+// The 's=' prefix can be omitted for string node ids in namespace 0.
 //
 // For numeric ids the smallest possible type which can store the namespace
 // and id value is returned.
 //
-// Namespace URLs 'nsu=' are not supported since they require a lookup.
+// Namespace URIs are not supported since NodeID cannot store them. If you need
+// to support namespace URIs use ParseExpandedNodeID.
 func ParseNodeID(s string) (*NodeID, error) {
-	if s == "" {
-		return NewTwoByteNodeID(0), nil
+	id, err := ParseExpandedNodeID(s, nil)
+	if err != nil {
+		return nil, err
 	}
-
-	var nsval, idval string
-
-	p := strings.SplitN(s, ";", 2)
-	switch len(p) {
-	case 1:
-		nsval, idval = "ns=0", p[0]
-	case 2:
-		nsval, idval = p[0], p[1]
-	default:
-		return nil, errors.Errorf("invalid node id: %s", s)
+	if id.HasNamespaceURI() {
+		return nil, errors.Errorf("namespace uris are not supported. use `ua.ParseExpandedNodeID`")
 	}
-
-	// parse namespace
-	var ns uint16
-	switch {
-	case strings.HasPrefix(nsval, "nsu="):
-		return nil, errors.Errorf("namespace urls are not supported: %s", s)
-
-	case strings.HasPrefix(nsval, "ns="):
-		n, err := strconv.Atoi(nsval[3:])
-		if err != nil {
-			return nil, errors.Errorf("invalid namespace id: %s", s)
-		}
-		if n < 0 || n > math.MaxUint16 {
-			return nil, errors.Errorf("namespace id out of range (0..65535): %s", s)
-		}
-		ns = uint16(n)
-
-	default:
-		return nil, errors.Errorf("invalid node id: %s", s)
+	if id.HasServerIndex() {
+		return nil, errors.Errorf("server index is not supported. use `ua.ParseExpandedNodeID`")
 	}
-
-	// parse identifier
-	switch {
-	case strings.HasPrefix(idval, "i="):
-		id, err := strconv.ParseUint(idval[2:], 10, 64)
-		if err != nil {
-			return nil, errors.Errorf("invalid numeric id: %s", s)
-		}
-		switch {
-		case ns == 0 && id < 256:
-			return NewTwoByteNodeID(byte(id)), nil
-		case ns < 256 && id < math.MaxUint16:
-			return NewFourByteNodeID(byte(ns), uint16(id)), nil
-		case id <= math.MaxUint32:
-			return NewNumericNodeID(ns, uint32(id)), nil
-		default:
-			return nil, errors.Errorf("numeric id out of range (0..2^32-1): %s", s)
-		}
-
-	case strings.HasPrefix(idval, "s="):
-		return NewStringNodeID(ns, idval[2:]), nil
-
-	case strings.HasPrefix(idval, "g="):
-		n := NewGUIDNodeID(ns, idval[2:])
-		if n == nil || n.StringID() == "" {
-			return nil, errors.Errorf("invalid guid node id: %s", s)
-		}
-		return n, nil
-
-	case strings.HasPrefix(idval, "b="):
-		b, err := base64.StdEncoding.DecodeString(idval[2:])
-		if err != nil {
-			return nil, errors.Errorf("invalid opaque node id: %s", s)
-		}
-		return NewByteStringNodeID(ns, b), nil
-
-	case strings.HasPrefix(idval, "ns="):
-		return nil, errors.Errorf("invalid node id: %s", s)
-
-	default:
-		return NewStringNodeID(ns, idval), nil
-	}
+	return id.NodeID, nil
 }
 
 // EncodingMask returns the encoding mask field including the
--- a/config.go
+++ b/config.go
@@ -53,9 +53,10 @@ func DefaultSessionConfig() *uasc.SessionConfig {
 
 // Config contains all config options.
 type Config struct {
-	dialer  *uacp.Dialer
-	sechan  *uasc.Config
-	session *uasc.SessionConfig
+	dialer                 *uacp.Dialer
+	sechan                 *uasc.Config
+	session                *uasc.SessionConfig
+	disableNamespaceUpdate bool
 }
 
 // NewDialer creates a uacp.Dialer from the config options
@@ -477,6 +478,14 @@ func DialTimeout(d time.Duration) Option {
 	return func(cfg *Config) {
 		initDialer(cfg)
 		cfg.dialer.Dialer.Timeout = d
+	}
+}
+
+// DisableUpdateNamespaceArrayOnConnect disables fetching the NamespaceArray
+// from the server on connect.
+func DisableUpdateNamespaceArrayOnConnect() Option {
+	return func(cfg *Config) {
+		cfg.disableNamespaceUpdate = true
 	}
 }
 
--- a/config.go
+++ b/config.go
@@ -53,10 +53,9 @@ func DefaultSessionConfig() *uasc.SessionConfig {
 
 // Config contains all config options.
 type Config struct {
-	dialer                 *uacp.Dialer
-	sechan                 *uasc.Config
-	session                *uasc.SessionConfig
-	disableNamespaceUpdate bool
+	dialer  *uacp.Dialer
+	sechan  *uasc.Config
+	session *uasc.SessionConfig
 }
 
 // NewDialer creates a uacp.Dialer from the config options
@@ -478,14 +477,6 @@ func DialTimeout(d time.Duration) Option {
 	return func(cfg *Config) {
 		initDialer(cfg)
 		cfg.dialer.Dialer.Timeout = d
-	}
-}
-
-// DisableUpdateNamespaceArrayOnConnect disables fetching the NamespaceArray
-// from the server on connect.
-func DisableUpdateNamespaceArrayOnConnect() Option {
-	return func(cfg *Config) {
-		cfg.disableNamespaceUpdate = true
 	}
 }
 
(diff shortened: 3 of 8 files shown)


magiconair:
Still need to add tests.

ua/node_id.go
```diff
@@ -99,8 +112,22 @@ func MustParseNodeID(s string) *NodeID {
...
 // and id value is returned.
 //
-// Namespace URLs 'nsu=' are not supported since they require a lookup.
+// Use ResolveNodeID for ids with a namespace URL.
 func ParseNodeID(s string) (*NodeID, error) {
+	return ResolveNodeID(s, nil)
+}
+
+// ResolveNodeID returns a node id from a string definition of the format
+// 'ns=<namespace>;{s,i,b,g}=<identifier>'.
```
kung-foo: `{ns,nsu}=...`

kung-foo:
Should we provide a "default" resolver? Or at least a helper? I have this bit of code in one of our util packages:

```go
func ReadNamespaces(client *opcua.Client) ([]string, error) {
	node := client.Node(ua.NewNumericNodeID(0, id.Server_NamespaceArray))
	v, err := node.Value()
	if err != nil {
		return nil, err
	}

	ns, ok := v.Value().([]string)
	if !ok {
		return nil, errors.Errorf("invalid namespace value type. id=%d, type=%T", v.Type(), v.Value())
	}

	return ns, nil
}
```

magiconair:
Yes, absolutely. I'll add it.

magiconair:
I guess we want to cache the namespace array

siscia:
Sorry, maybe I missunderstand the whole point of the ParseNodeId function.

But if we try to 

```go
node, err := ua.ParseNodeId("ns=12;s=foo")
```
where the namespace = 12 is not defined, does the `ParseNodeId` function returns an error?

Why should it be different for `nsu`? 

I am missing some OPCUA specs?

kung-foo:
@siscia The idea is that instead of `ns=5;s=Counter1` you could use `nsu=http://www.prosysopc.com/OPCUA/SimulationNodes;s=Counter1`. To resolve the url to 5, you need to query the `Server_NamespaceArray`.

`ParseNodeId` does not _validate_ any of the values, just the structure. So it will send whatever is in there to the server.

~~@magiconair what happens if we just send an `nsu=...` node it to the server?~~ Nevermind, forgot that `ns` must be an `int16`

kung-foo (approved):

magiconair:
Thx. I'll add some tests and then merge it.

ua/resolve/resolve.go
```diff
@@ -0,0 +1,67 @@
...
+// Lookup returns the id from a list of namespace URLs or an error.
+// The values are not parsed and are compared case-sensitive.
+func Lookup(s string, ns []string) (uint16, error) {
+	for id, nsu := range ns {
+		if nsu == s {
+			return uint16(id), nil
+		}
+	}
+	return 0, errors.Errorf("unknown namespace url %s", s)
+}
```
siscia: This will be called as `resolve.Lookup` where we are trying to get numeric id of a namespace.

Shouldn't the string "namespace" into the invocation?

Something like "namespace.Lookup()"?

magiconair: good point. Maybe we can call it `nsu`

ua/resolve/resolve.go
```diff
@@ -0,0 +1,67 @@
...
+	return nil
+}
+
+// Lookup returns the id of the namespace url or an error.
+// The values are not parsed and are compared case-sensitive.
+func (r *ClientResolver) Lookup(s string) (uint16, error) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	return Lookup(s, r.ns)
+}
```
siscia: If we call this us:

```
c := ClientResolver{Client: c}
c.Lookup() 
```
will return an error as `c.ns` is nil.

So we should first invoke `.Refresh` which is odd.

magiconair: Good point. I could refactor it like this:

```go
type ClientResolver struct {
	c  *opcua.Client
	mu sync.Mutex
	ns []string
}

func NewClientResolver(c *opcua.Client) (*ClientResolver, error) {
	r := &ClientResolver{c: c}
	if err := r.Refresh(); err != nil {
		return nil, err
	}
	return r, nil
}
```

but that would require that the connection has been established before you can use the resolver. It might also make sense to move this into the `Client` itself, e.g. `c.Resolver() ua.NSUResolver`. Maybe `NamespaceResolver` is a better name as well.

magiconair:
After thinking about this I think this approach is wrong. I would like to change `ParseNodeID(s string)` to `ParseNodeID(s string, ns []string)` and make the resolution entirely static. The client should then maintain the namespace array by fetching it on connection and with an optional refresh timer as an option. Opinions?

magiconair:
This is what I should have done right from the start but hindsight is always 20-20. Allow for parsing all node ids in `ParseNodeID` and move the responsibility of fetching and caching the namespace array to the caller. This is also why I would feel comfortable in breaking `ParseNodeID` at this point since it feels like the right change and it is easy to fix.

magiconair:
Both approaches would not make it possible to have `nsu=` in JSON node ids though. :/

magiconair:
Hmm, going down this route also means to pass the `ns` array into every lookup function everywhere. this doesn't look promising. But only passing it into the `ParseNodeID` function means that we only support `nsu=` ids for a subset of cases. I think to make this work we probably need to hide some magic like a global namespace list.

siscia:
What about adding `ParseNodeIDWithNamespaces(s string, ns []string)`

However, I would like to point out that `NodeID`s and `ExtendedNodeID`s (the one with `nsu=...`) are different things.

I am not sure hiding them under the same object is the best choice.

magiconair:
I think I have a better approach. We basically leave the `ParseNodeID` function as is but hide it behind a global node id parser which can be swapped out. 

```go
func init() {
	parser.Store(&NamespaceParser{})
}

// parser stores the global node id parser.
// It is guaranteed that there will always be a node id parser.
var parser atomic.Value // NodeIDParser

// RegisterNodeIDParser configures a new global node id parser.
func RegisterNodeIDParser(p NodeIDParser) {
	if p == nil {
		panic("NodeIDParser must not be nil")
	}
	parser.Store(p)
}

// NodeIDParser defines the interface for a node id parser.
type NodeIDParser interface {
	Parse(s string) (*NodeID, error)
}

// NamespaceParser provides a node id parser which is aware of
// server namespaces.
type NamespaceParser struct {
	Namespaces []string
}

func (p *NamespaceParser) Parse(s string) (*NodeID, error) {
	return parseNodeID(s, p.Namespaces)
}

// MustParseNodeID returns a node id from a string definition
// if it is parseable by ParseNodeID. Otherwise, the function
// panics.
func MustParseNodeID(s string) *NodeID {
	id, err := ParseNodeID(s)
	if err != nil {
		panic(err.Error())
	}
	return id
}

// ParseNodeID returns a node id from a string definition of the format
// '{ns,nsu}=<namespace>;{s,i,b,g}=<identifier>'.
//
// The 's=' prefix can be omitted for string node ids in namespace 0.
//
// For numeric ids the smallest possible type which can store the namespace
// and id value is returned.
//
// It depends on the implementation of the registered node id parser whether
// the function is capable of resolving namespace uris.
func ParseNodeID(s string) (*NodeID, error) {
	return parser.Load().(NodeIDParser).Parse(s)
}

func parseNodeID(s string, nsu []string) (*NodeID, error) {
...
}
```

magiconair:
Then we change the `client.Dial` as follows and boom :

```patch
diff --git a/client.go b/client.go
index eb79f54..a7b9588 100644
--- a/client.go
+++ b/client.go
@@ -480,11 +480,23 @@ func (c *Client) Dial(ctx context.Context) error {
 
 	c.sechan, err = uasc.NewSecureChannel(c.endpointURL, c.conn, c.cfg.sechan, c.sechanErr)
 	if err != nil {
-		_ = c.conn.Close()
+		c.conn.Close()
 		return err
 	}
 
-	return c.sechan.Open(ctx)
+	if err := c.sechan.Open(ctx); err != nil {
+		c.conn.Close()
+		return err
+	}
+
+	ns, err := c.NamespaceArray()
+	if err != nil {
+		c.conn.Close()
+		return err
+	}
+
+	ua.RegisterNodeIDParser(&ua.NamespaceParser{Namespaces: ns})
+	return nil
 }
 
 // Close closes the session and the secure channel.
```

magiconair:
For auto-refresh of the namespaces we add an option which kicks off a go routine in the background to fetch the namespace array on a regular basis.

@kung-foo @siscia what do you think?

magiconair:
So where the smart client approach fails if you have a program with multiple OPC/UA connections since they would clobber each others' namespaces. So we should make the auto-initialising of namespaces optional but enable it by default.

magiconair:
pushing in a minute

kung-foo:
A couple of thoughts (in no particular order):

* I don't like globals like the registry. In this case, it fails to allow for multiple client connections in the same process.
* Since the namespace index (according to the spec) is _only_ valid during the lifetime of the current connection, then it sorta implies that funcs like `ParseNodeID` should be on the client class.
* we (my company) have had a few discussions internally about how we store node ids. right now, we just do something like browse -> store node ids in db. But these node ids always have a ns index and not the actual namespace. if we were to update our flow, we would store _only_ the nsu= version, and never "see" the indexed version as it would be handled transparently in all calls.

kung-foo:
I checked our (my company) codebase to see where we use `ua.ParseNodeID`. There are two places where we use it, and there is a client connection "nearby" in both cases.

kung-foo:
In the main opcua repo, we allow raw node ids (i.e. just strings) in the `Subscription.AddNodes(...)` API. There we call `us.ParseNodeID` as a convenience. But, even in that case, there is client conn available.

This has me leaning towards moving `ParseNodeID` to the client, and optionally maintaining a static helper function that explicitly takes a ns slice.

magiconair:
Yeah, that might make even more sense.

magiconair:
But how would that work for JSON decoding?

magiconair:
However, this also means that you cannot use `nsu=` for services and extension object types unless we move them into the client as well.

kung-foo:
I've never touched any JSON stuff (with regards to OPC-UA) so I'm not familiar with the use case.

Should the `NodeID` struct contain the namespace uri?

magiconair:
I guess that would be a reasonable solution.

magiconair:
Should we then change `ParseNodeID` to contain the `ns` list?

kung-foo:
virtual whiteboard:

```
// tag with numeric id, may or may not work downstream in api calls
ua.ParseNodeID("ns=2;s=Tag1") ->  NodeID(ns=2, s=Tag1)

// guarantees (or returns an error) that the numeric ns at least exists, though "Tag1" may or may not 
conn.ParseNodeID("ns=2;s=Tag1") ->  NodeID(ns=2, nsu=urn:my-server, s=Tag1)

// guarantees (or returns an error) that the ns exists. "Tag1" _probably_ exists assuming nothing changed with regards to the tag itself
conn.ParseNodeID("nsu=urn:my-server;s=Tag1") -> NodeID(ns=1337, nsu=urn:my-server, s=Tag1)

// current api
ua.ParseNodeID("nsu=urn:my-server;s=Tag1") -> Err

// new api
ua.ParseNodeID("nsu=urn:my-server;s=Tag1", client.NamespaceArray()) -> NodeID(ns=1337, nsu=urn:my-server, s=Tag1)

// current approach
conn.Browse() -> NodeID(ns=2, s=Tag1).String() -> "ns=2;s=Tag1"

// better approach? essentially we always fillout the nsu field when possible
conn.Browse() -> NodeID(ns=2, nsu=urn:my-server, s=Tag1).String() -> "nsu=urn:my-server;s=Tag1"

// current api (returns ns index is which is not guaranteed the moment the connection closes)
NodeID(ns=2, s=Tag1).String() -> "ns=2;s=Tag1"

// updated string api (returns nsu _iff_ it is set)
NodeID(ns=1337, nsu=urn:my-server, s=Tag1).String() -> "nsu=urn:my-server;s=Tag1"
```

kung-foo:
@siscia 
> What about adding ParseNodeIDWithNamespaces(s string, ns []string)
> However, I would like to point out that NodeIDs and ExtendedNodeIDs (the one with nsu=...) are different things.
> I am not sure hiding them under the same object is the best choice.

I guess from my perspective, the `NodeID` (with only a ns index) is an internal, not-valid-outside-this-conn object. Whereas the `ExtendedNodeID` is the one I end reading and writing with.

So I'd _almost_ be ok with saying the `ExtendedNodeID` _is_ the `NodeID`, and that the ns-index-only is just a private bit that no client sees.

But that's probably too opinionated for this lib...

siscia:
So, the real important details that was contributed in the discussion in the fact that the namespace array is valid only as long as the OPCUA session.

This means that the client should not worry about cache invalidation, as long as it uses a single session.

It is also telling us that we should treat the `nsu` as first class citizens, since each time we store, a nodeid with a namespace encoded, is a possible source of problem in the future.

But something like 

`ParseExtendedNodeId(s) *ExtendedNodeID` and then provide a method:
`(e *ExtendedNodeID) NodeID(ns []string) *NodeID`

Does not work?

I really don't like to force the parsing of node IDs to go through a connection.

It forces us to have a connection to validate that a string is a valid NodeID, that I find wrong as we are forcing a string to be a valid nodeid in a specific amount of time and when there is an OPCUA server close by.

magiconair:
Namespaces can change during the session

siscia:
From the spec:


> NamespaceArray defines an array of namespace URIs. This Variable is also referred as namespace table. The indexes into the namespace table are referred to as NamespaceIndexes. NamespaceIndexes are used in NodeIds in OPC UA Services, rather than the longer namespace URI. Index 0 is reserved for the OPC UA namespace, and index 1 is reserved for the local Server. Clients may read the entire namespace table or they may read individual entries in the namespace table. The Server shall not modify or delete entries of the namespace table while any client has an open session to the Server, because clients may cache the namespace table. A Server may add entries to the namespace table even if clients are connected to the Server. It is recommended that Servers not change the indexes of the namespace table but only add entries, because the client may cache NodeIds using the indexes. Nevertheless, it might not always be possible for Servers to avoid changing indexes in the namespace table. Clients that cache NamespaceIndexes of NodeIds should always check when starting a session to verify that the cached NamespaceIndexes have not changed.


https://reference.opcfoundation.org/v104/Core/docs/Part5/6.3.1/

So yes it can change during a session, but it should not delete entries.

kung-foo:
> It forces us to have a connection to validate that a string is a valid NodeID, that I find wrong as we are forcing a string to be a valid nodeid in a specific amount of time and when there is an OPCUA server close by.

Is it possible to validate a NodeID without a session? We can validate that it (assuming a string) syntactically represents a node. I guess it is the difference between parsing a node id and validating/resolving.

magiconair:
I think what we are struggling with is in the fact that a NodeID does not support namespace URIs. It only supports numeric namespace indexes. So from that perspective adding support for namespaces to the `ParseNodeID` function seems wrong. And maybe that is why we are struggling with this. 

https://reference.opcfoundation.org/v104/Core/docs/Part6/5.2.2/#5.2.2.9

Namespace URIs only make sense in the context of a session since they have to be resolved to a numeric id before they can be used. So maybe we should leave `ParseNodeID` the way it is and don't support `nsu=` but instead add a `ParseExpandedNodeID` function for the static parsing and a `client.Namespaces()` function which returns the list of cached namespaces.

magiconair:
Something like this:

```go
// ParseNodeID returns a node id from a string definition of the format
// 'ns=<namespace>;{s,i,b,g}=<identifier>'.
//
// The 's=' prefix can be omitted for string node ids in namespace 0.
//
// For numeric ids the smallest possible type which can store the namespace
// and id value is returned.
//
// Namespace URIs are not supported since NodeID cannot store them. If you need
// to support namespace URIs use ParseExpandedNodeID.
func ParseNodeID(s string) (*NodeID, error) {
	id, err := ParseExpandedNodeID(s, nil)
	if err != nil {
		return nil, err
	}
	if id.NamespaceURI != "" {
		return nil, errors.Errorf("namespace uris are not supported")
	}
	return id.NodeID, nil
}

// ParseExpandedNodeID returns a node id from a string definition of the format
// '{ns,nsu}=<namespace>;{s,i,b,g}=<identifier>'.
//
// The 's=' prefix can be omitted for string node ids in namespace 0.
//
// For numeric ids the smallest possible type which can store the namespace
// and id value is returned.
//
// Namespace URIs are resolved to ids from the provided list of namespaces.
func ParseExpandedNodeID(s string, ns []string) (*ExpandedNodeID, error) {
...
}
```

magiconair:
@kung-foo and @siscia could you please have another look?

magiconair:
Make sure you look at the `Files` section in the review tab to see the current list of changes to `main` and not just the delta to the previous iteration.

kung-foo:
ugh. :facepalm: I never noticed there was an _actual_ `ExpandedNodeID` struct in the lib. I thought this was all hypothetical... Ignore most of my comments today...

siscia:
I think it looks good!

ua/node_id.go
```diff
@@ -118,10 +142,22 @@ func ParseNodeID(s string) (*NodeID, error) {
...
 	// parse namespace
-	var ns uint16
+	var nsid uint16
+	var nsu string
 	switch {
 	case strings.HasPrefix(nsval, "nsu="):
-		return nil, errors.Errorf("namespace urls are not supported: %s", s)
+		nsuval := strings.TrimPrefix(nsval, "nsu=")
+		ok := false
+		for id, uri := range ns {
```
kung-foo: I think a separate error for `ns == nil` will help.

```go
if ns == nil {
    return nil, errors.Error("namespace urls require a server NamespaceArray")
}
```

ua/node_id.go
```diff
@@ -118,10 +142,22 @@ func ParseNodeID(s string) (*NodeID, error) {
...
+			if uri == nsuval {
+				nsid, nsu = uint16(id), uri
+				ok = true
+				break
+			}
+		}
+		if !ok {
+			return nil, errors.Errorf("invalid namespace uri %s", nsval)
```
kung-foo:
```go
return nil, errors.Errorf("namespace uri %q not found in the server NamespaceArray %q", nsval, ns)
```

ua/node_id.go
```diff
@@ -94,15 +94,39 @@ func MustParseNodeID(s string) *NodeID {
...
-// Namespace URLs 'nsu=' are not supported since they require a lookup.
+// Namespace URIs are not supported since NodeID cannot store them. If you need
+// to support namespace URIs use ParseExpandedNodeID.
 func ParseNodeID(s string) (*NodeID, error) {
+	id, err := ParseExpandedNodeID(s, nil)
+	if err != nil {
+		return nil, err
+	}
+	if id.HasNamespaceURI() {
+		return nil, errors.Errorf("namespace uris are not supported")
```
kung-foo:
```go
errors.Errorf("namespace uris are not supported. use `ua.ParseExpandedNodeID`")
```

ua/node_id.go
```diff
@@ -94,15 +94,39 @@ func MustParseNodeID(s string) *NodeID {
...
+	}
+	if id.HasServerIndex() {
+		return nil, errors.Errorf("server index is not supported")
```
kung-foo:
```go
errors.Errorf("server index is not supported. use `ua.ParseExpandedNodeID`")
```

client.go
```diff
@@ -480,11 +484,16 @@ func (c *Client) Dial(ctx context.Context) error {
...
 		return err
 	}
 
-	return c.sechan.Open(ctx)
+	if err := c.sechan.Open(ctx); err != nil {
+		c.conn.Close()
+		return err
+	}
+
+	return c.UpdateNamespaces()
```
kung-foo: Is there ever a case where someone is dialing a server, but does _not_ have permission to read `Server_NamespaceArray`? Or maybe is connecting to a broken server? Maybe this should be gated by an option?

magiconair:
I'll flatten the changes and work on your comments @kung-foo. I'll add the option to update the namespaces back again. Also need to find out why the tests are failing all of a sudden.

ua/node_id.go
siscia: Can we use the %q instead of the %s ?

I find that avoid issues when the value is empty or when it is something that may follow the flow of the text!

client.go
siscia: This is one of those cases where I will suggest to wait for somebody to actually complain about it or find a broken server somewhere.

Adding an option will add complexity in the interface and we are covering ourselves from a case that it is quite unlikely to happen in my opinion.

siscia:
Yesterday I though more about this.

And I am not able to see why the interface I proposed above was not liked.

Unfortunately, `ExtendedNodeID` where introduces as an after though to the OPC/UA implementation (or at least it seems like) and I don't think that it makes any benefit in trying to hide the fact that those are different.

I would really like to understand why the interface:

```go
ParseExtendedNodeId(s) (*ExtendedNodeID, err)
(e *ExtendedNodeID) NodeID(ns []string) (*NodeID, err)
```

was not considered.

In case of defaults nodeids, the ParseExtendedNodeId function will return a zero value ExtendedNodeID with only the NodeID field set.

In the case of an ExtendedNodeId `nsu=...` the function will return the ExtendedNodeID without verifying that the namespace exists.

Finally the method will allow to jump back to the standard `NodeID` if the namespace match, otherwise will raise an error.

ua/node_id.go
kung-foo: comment updated

magiconair:
Afterthought or not, `ExpandedNodeID` is part of the standard and it exists to support node ids with namespace URIs. By using two different parsing functions we make it clear that they are different. 

I think the main issue with what you're proposing is that it might break existing code since you could no longer rely on the public `NodeID` field to be valid. Also, using a two step process to resolve a string to a final node id seems too complex. I think users know whether they deal with `nsu` or not. And if they are unsure they have to use `ExpandedNodeID` from the start. 

It also feels wrong to me to make the `ExpandedNodeID` responsible for resolving a node id uri. It should be just a data container and all logic is usually handled in helper functions. @kung-foo what do you think?

ua/node_id.go
```diff
@@ -118,10 +142,26 @@ func ParseNodeID(s string) (*NodeID, error) {
...
+		ok := false
+		for id, uri := range ns {
+			if uri == nsuval {
+				nsid, nsu = uint16(id), uri
+				ok = true
+				break
+			}
+		}
+		if !ok {
+			return nil, errors.Errorf("namespace uri %s not found in the server NamespaceArray %s", nsval, ns)
```
kung-foo: nit: @siscia suggested using `%q`. makes sense, though we aren't particularly consistent about this.

magiconair: No, we're not

magiconair: When I look at the `ParseNodeID` function we're always using `%s` - right or not. I prefer consistency here. If we want to change this we should change it everywhere or at least in enough places.

magiconair:
Rebasing to `main`

magiconair:
Found the bug which made the tests fail.

magiconair:
I'll squash the changes into a single commit. We don't need the path on how we got here.

client.go
magiconair: Oh, missed this comment. Then I'll remove the option again. I'm in favor of keeping it simple.

magiconair: Kept this in a separate commit so reverting it will implement the option :)

PR #522, opened 2021-12-06
subscription: add support to modify monitored items

### LINKED ISSUE #492: New Feature: Implement modifyMonitoredItem
we have `subscription.Monitor()` and `subscription.Unmonitor()`, but modify functionality is missing.

magiconair:
Fixes #492

### DIFF
--- a/subscription.go
+++ b/subscription.go
@@ -164,7 +164,16 @@ func (s *Subscription) Unmonitor(monitoredItemIDs ...uint32) (*ua.DeleteMonitore
 	return res, err
 }
 
-func (s *Subscription) Modify(ts ua.TimestampsToReturn, items ...*ua.MonitoredItemModifyRequest) (*ua.ModifyMonitoredItemsResponse, error) {
+func (s *Subscription) ModifyMonitoredItems(ts ua.TimestampsToReturn, items ...*ua.MonitoredItemModifyRequest) (*ua.ModifyMonitoredItemsResponse, error) {
+	s.itemsMu.Lock()
+	for _, item := range items {
+		id := item.MonitoredItemID
+		if _, exists := s.items[id]; !exists {
+			return nil, fmt.Errorf("sub %d: cannot modify unknown monitored item id: %d", s.SubscriptionID, id)
+		}
+	}
+	s.itemsMu.Unlock()
+
 	req := &ua.ModifyMonitoredItemsRequest{
 		SubscriptionID:     s.SubscriptionID,
 		TimestampsToReturn: ts,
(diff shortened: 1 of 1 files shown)


magiconair:
@salmanazmat666 can you have a look please?

subscription.go
```diff
@@ -164,6 +164,40 @@ func (s *Subscription) Unmonitor(monitoredItemIDs ...uint32) (*ua.DeleteMonitore
...
+	})
+
+	if err == nil {
+		// update monitored items
+		s.itemsMu.Lock()
+		for i, res := range res.Results {
+			id := req.ItemsToModify[i].MonitoredItemID
+			item := s.items[id]
+			if item == nil {
+				return nil, fmt.Errorf("sub %d: cannot modify unknown monitored item id: %d", s.SubscriptionID, id)
```
salmanazmat666: @magiconair 
wouldn't it be better that we check for unknown monItem id before sending the request ? since this is a check on Request struct and not on the Response struct. If the request is not valid, we don't process it at all.
Currently, if one monItem is unknown, you don't process the rest of items, even if they are valid.

So, my suggestion would be to
1. Verify the Request before it is sent.
2. And in the for loop, we only check and update those monItems whose response has status code OK. If it's not Ok, then they are not modified on the server, and we shouldn't modify on client side as well, so we just ignore them in the for loop.
3. Also, if the function returns here, then mutex will stay Locked, which will cause deadlock. so we should use defer to Unlock the mutex.

```go
func (s *Subscription) Modify(ts ua.TimestampsToReturn, items ...*ua.MonitoredItemModifyRequest) (*ua.ModifyMonitoredItemsResponse, error) {
	for i, item := range items {
		id := item.MonitoredItemID
		item := s.items[id]
		if _, exists := s.items[id]; !exists {
			return nil, fmt.Errorf("sub %d: cannot modify unknown monitored item id: %d", s.SubscriptionID, id)
		}
	}

	req := &ua.ModifyMonitoredItemsRequest{
		SubscriptionID:     s.SubscriptionID,
		TimestampsToReturn: ts,
		ItemsToModify:      items,
	}
	var res *ua.ModifyMonitoredItemsResponse
	err := s.c.Send(req, func(v interface{}) error {
		return safeAssign(v, &res)
	})

	if err == nil {
		// update monitored items
		s.itemsMu.Lock()
		defer s.itemsMu.Unlock()
		for i, res := range res.Results {
			if res.StatusCode == ua.StatusCodeOK {
				id := req.ItemsToModify[i].MonitoredItemID
				item := s.items[id]
				item.ts = req.TimestampsToReturn
				item.req.RequestedParameters = req.ItemsToModify[i].RequestedParameters
				item.res.StatusCode = res.StatusCode
				item.res.RevisedSamplingInterval = res.RevisedSamplingInterval
				item.res.RevisedQueueSize = res.RevisedQueueSize
				item.res.FilterResult = res.FilterResult
			}
		}
	}

	return res, err
}
```

subscription.go
```diff
@@ -150,11 +154,46 @@ func (s *Subscription) Unmonitor(monitoredItemIDs ...uint32) (*ua.DeleteMonitore
...
 		for _, id := range monitoredItemIDs {
 			delete(s.items, id)
 		}
+		s.itemsMu.Unlock()
+	}
+
+	return res, err
+}
+
+func (s *Subscription) Modify(ts ua.TimestampsToReturn, items ...*ua.MonitoredItemModifyRequest) (*ua.ModifyMonitoredItemsResponse, error) {
```
salmanazmat666: @magiconair 
`Modify()` feels like if we are modifying the subscription iteself. A better name might be `ModifyMonitoredItems()`.
And we can use Modify() for sending the `ModifySubscriptionRequest`, which is also yet to be implemented.

magiconair:
All good points. Thank you.

subscription.go
magiconair: You have to lock the first loop as well and there is no benefit in using `defer` in the second loop.

magiconair:
@salmanazmat666 can you have another look, please?

salmanazmat666:
> @salmanazmat666 can you have another look, please?

@magiconair 
Hey, thanks for the changes, looks good to me,

PR #527, opened 2021-12-09
client: exit client on timeout when auto-reconnect is disabled

### LINKED ISSUE #504: Don't start monitor when AutoReconnect is disabled
Should the monitor be started if AutoReconnect is disabled?

For example:
If we receive a STATUS_BAD_TIMEOUT from the server the monitor loop will pause the subscriptions and set state to disconnected.

magiconair:
Fixes #504

### DIFF
--- a/client.go
+++ b/client.go
@@ -270,6 +270,14 @@ func (c *Client) monitor(ctx context.Context) {
 				switch x := err.(type) {
 				case *uacp.Error:
 					switch ua.StatusCode(x.ErrorCode) {
+					case ua.StatusBadTimeout:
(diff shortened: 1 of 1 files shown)


magiconair:
@VereeckenWannes could you have a look whether this addresses your issue in #504 ?

vereecw:
This doesn't fix our issue.
The state is still set to disconnected and the subscriptions are still paused.
Plus it won't ever reach this code since the if on line 252 in this file does the same thing.

magiconair:
OK, thank you.

Issue #502, opened 2021-11-19
Client functions that accept a context to be cancellable.


CrowdHailer:
Calls to client functions (such as Read/Browse) may be slow and may want to be cancelled.
A standard pattern for handling this is for functions to take contexts. Can we add the following.

- `ReadWithContext(ctx, request) (response, error)`
- ... etc

`ReadWithContext` is a potentially verbose API, the other option would be to add context as an argument to existing functions. But this would be a breaking change.

PR #528, opened 2021-12-16
add contexts to all functions that make a remote call

CrowdHailer:
Proposal to addresses issue #502. This PR is exploration of how much the API would need to break for all methods that call an OPC UA server to accept a context. Some notes below

- [ ] TODO finish updating examples. (checking if this approach is acceptable first)
- Myself and a college are of the opinion that making breaking changes to the API rather than adding `client.FooWithContext` is the correct way forward. because there are so many functions that make calls to the server. i.e. ActivateSession, AttributeID that it would be extremely cumbersome to have them all duplicated.
- also we noted that the timeout value passed to `sendWithTimeout` was sent as a hint to the server. This has subtly different meaning to cancelling a context and so we have left functions to take both timeout and context.

### DIFF
--- a/client.go
+++ b/client.go
@@ -33,7 +33,7 @@ func GetEndpoints(ctx context.Context, endpoint string, opts ...Option) ([]*ua.E
 		return nil, err
 	}
 	defer c.Close()
-	res, err := c.GetEndpoints()
+	res, err := c.GetEndpoints(ctx)
 	if err != nil {
 		return nil, err
 	}
@@ -190,13 +190,13 @@ func (c *Client) Connect(ctx context.Context) (err error) {
 		return err
 	}
 
-	s, err := c.CreateSession(c.cfg.session)
+	s, err := c.CreateSession(ctx, c.cfg.session)
 	if err != nil {
 		c.Close()
 		return err
 	}
 
-	if err := c.ActivateSession(s); err != nil {
+	if err := c.ActivateSession(ctx, s); err != nil {
 		c.Close()
 		return err
 	}
@@ -213,7 +213,7 @@ func (c *Client) Connect(ctx context.Context) (err error) {
 	// todo(fs): server. For the sake of simplicity we left the option out but
 	// todo(fs): see the discussion in https://github.com/gopcua/opcua/pull/512
 	// todo(fs): and you should find a commit that implements this option.
-	if err := c.UpdateNamespaces(); err != nil {
+	if err := c.UpdateNamespaces(ctx); err != nil {
 		c.Close()
 		return err
 	}
@@ -366,7 +366,7 @@ func (c *Client) monitor(ctx context.Context) {
 							action = createSecureChannel
 							continue
 						}
-						if err := c.ActivateSession(s); err != nil {
+						if err := c.ActivateSession(ctx, s); err != nil {
 							dlog.Printf("restore session failed")
 							action = recreateSession
 							continue
@@ -380,13 +380,13 @@ func (c *Client) monitor(ctx context.Context) {
 						// create a new session to replace the previous one
 
 						dlog.Printf("trying to recreate session")
-						s, err := c.CreateSession(c.cfg.session)
+						s, err := c.CreateSession(ctx, c.cfg.session)
 						if err != nil {
 							dlog.Printf("recreate session failed: %v", err)
 							action = createSecureChannel
 							continue
 						}
-						if err := c.ActivateSession(s); err != nil {
+						if err := c.ActivateSession(ctx, s); err != nil {
 							dlog.Printf("reactivate session failed: %v", err)
 							action = createSecureChannel
 							continue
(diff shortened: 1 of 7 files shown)


magiconair:
This will break a loooot of code so I'm quite reluctant to merge this into the v0.2.x versions. v0.3.0 at least. 

Can you split the PR into the change that adds the timeout handling to the `sendWithTimeout` method, then one which updates the methods and a third one which updates the examples? I know that they won't compile but it will make it somewhat easier to review than this big blob.

@kung-foo I'd like some input here on how to proceed. We need to warn the Influx team at least.

siscia:
If you don't want to create too much trouble, consider to provide a `FuncNameNoCtx()` that act just like `FuncName` now (without a context) and then highlight how to use the gofmt to change all the code automatically.

Another alternative would it be to provide `FuncNameWithCtx` and leave `FuncName` as it is, but :/

magiconair:
`context` became part of the standard library with go1.7 and granted I should have added this from the start but alas here we are. 

@siscia `FooNoCtx` would serve no purpose since we want to add `context` and it would still break the API. Also, I haven't seen anybody else doing this since using context is not controversial anymore. At least to my knowledge. 

[`database/sql`](https://pkg.go.dev/database/sql) adds a `FooContext` and AWS adds a  [`FooWithContext`](https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.42.23/service/dynamodb#DynamoDB.BatchExecuteStatementWithContext) method so that users can migrate to using `context` at their own pace. 

These posts on the Go blog talk about how to do that:

https://go.dev/blog/context-and-structs
https://go.dev/blog/module-compatibility

The question is how much we care about backwards compatibility. In the README we've stated that users [should not expect API stability](https://github.com/gopcua/opcua#disclaimer) but that we would try hard not to break users programs without a good reason. And breaking the `Client` because we don't want to add another method doesn't feel like we're trying very hard. 

Since we are no longer using this library in our own little bubble I'm leaning towards the `FooContext` approach since it creates the least amount of friction for the users and the amount of extra code is negligible. 

```go
func Foo() error {
    return FooContext(context.Background())
}

func FooContext(ctx context.Context) error {
    ...
}
```

I'd like to get some outside NV opinion on this: @kung-foo @alexbrdn what is your opinion?

kung-foo:
I took a look at our code base to see how difficult it would be to require a context. Using @CrowdHailer's branch, there were 17 calls in our app that needed to be updated. Of the 17, 13 calls had a context var available in the surrounding code. And 4 required a `context.TODO()`.

So overall, not a big deal (though of course, a bit more work is required to ensure the caller's context makes sense).

I'm always a fan of doing it the right way, even if it entails a bit of refactoring. So saying that in `0.5.0`, context is required, is fine by me.

I am more concerned with the different API this exposes. For example, in `Client.Subscribe`, is the context used for just the setup? Does canceling a context here implicitly call `Unsibscribe`? It should be crystal clear what canceling does for each API we expose with a context.

magiconair:
I've asked @CrowdHailer to rebase the PR to `main` and change the method names to `XXXContext(ctx)` so that we can get this merged and tested. We then deprecate the old functions and switch from `XXXContext(ctx)` to just `XXX(ctx)` in v0.4.0. I'd like to test this first in our environment. 

@CrowdHailer could you also address @kung-foo:s comment on documenting what the context does where it isn't obvious?

CrowdHailer:
One other nice thing about keeping backwards compatibility is that we can break down the PR for the changes and have each of them compile. Therefore here are the parts I have broken it down too,.

 First #541 To add a context to the lowest level, where the requirement is to listen to `ctx.Done` **This needs to be merged first**
 Then a PR each for 
- public client methods #542
- methods for subscriptions (coming soon)

CrowdHailer:
@kung-foo 

Sorry I didn't mean to skip over your comment.

> I am more concerned with the different API this exposes. For example, in Client.Subscribe, is the context used for just the setup? Does canceling a context here implicitly call Unsibscribe? It should be crystal clear what canceling does for each API we expose with a context.

So I suppose to some degree the API is different by definition, it would never cancel before. In the example you give. No the context passed to subscribe can only be used to cancel the subscription process. Once subscribed an explicit call to Unsubscribe is needed. This seemed the most intuitive to me. Also in a usecase I had, I started a subscription in a web request. If the process of subscribing failed then I would like to inform the client of that. However once subscription process was completed I returned ok and the web request context was cancelled. I would be very confused if that had then called an Unsubscribe, because once subscription was running the process of monitoring continued.

Potentially it's not obvious, and I agree it should be super clear in documentation. At least with a backwards compatible approach it can be tried this way with minimal impact on other users. I will certainly add documentation on the subscription PR.

@magiconair the two PR's I linked above don't rely on the subscription so is there anything else stopping them being merged.

magiconair:
Superseded by #541 and #542 as a non-breaking change

PR #533, opened 2021-12-20
client: expose statistics

### LINKED ISSUE #532: Expose statistics
We should expose statistics about number of operations, monitored nodes, status changes and so forth. I need this for an internal troubleshooting project and can expose an experimental API for this.

magiconair:
This patch collects statistics for all client operations and some of the
errors and exposes them via the Stats() method.

This API is experimental and should not be considered stable at this
point.

Fixes #532

### DIFF
--- a/client.go
+++ b/client.go
@@ -7,6 +7,7 @@ package opcua
 import (
 	"context"
 	"crypto/rand"
+	"expvar"
 	"fmt"
 	"io"
 	"log"
@@ -20,6 +21,7 @@ import (
 	"github.com/gopcua/opcua/debug"
 	"github.com/gopcua/opcua/errors"
 	"github.com/gopcua/opcua/id"
+	"github.com/gopcua/opcua/stats"
 	"github.com/gopcua/opcua/ua"
 	"github.com/gopcua/opcua/uacp"
 	"github.com/gopcua/opcua/uasc"
@@ -186,17 +188,23 @@ func (c *Client) Connect(ctx context.Context) (err error) {
 
 	c.setState(Connecting)
 	if err := c.Dial(ctx); err != nil {
+		stats.RecordError(err)
+
 		return err
 	}
 
 	s, err := c.CreateSession(c.cfg.session)
 	if err != nil {
 		c.Close()
+		stats.RecordError(err)
+
 		return err
 	}
 
 	if err := c.ActivateSession(s); err != nil {
 		c.Close()
+		stats.RecordError(err)
+
 		return err
 	}
 	c.setState(Connected)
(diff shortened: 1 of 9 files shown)


kung-foo:
Maybe we could use the stdlib's `expvar` package?

magiconair:
Yes, I didn't think of that. I'll have a look.

kung-foo:
I've never looked into how a _lib_ should expose metrics via `expvar`, but I assume something as simple as prefixing each var with `gopcua.` would work.

magiconair:
I think `expvar` is a more generic approach for which ready-made integrations with other systems exist. Prometheus for example as an `expvar` collector. So that seems a much better choice.

magiconair:
Refactored using `expvar`

magiconair:
```
"gopcua": {"Client": {"Dial": 1, "NodesToRead": 2, "Read": 2, "Send": 2, "State": 0}, "Error": {"ua.StatusBadTimeout": 1}, "Subscription": {}},
```

kung-foo:
error name reflection doesn't seem to work.

```
stats.RecordError(errors.New("hello"))
spew.Dump(stats.Error)
```

```
(*expvar.Map)(0xc0000a4600)({"": 1})
```

kung-foo:
here's a bit of test code I was using:

```go
//go:build integration
// +build integration

package uatest

import (
	"context"
	"expvar"
	"testing"

	"github.com/gopcua/opcua"
	"github.com/gopcua/opcua/stats"
	"github.com/pascaldekloe/goe/verify"
)

func newExpVarInt(i int64) *expvar.Int {
	v := &expvar.Int{}
	v.Set(i)
	return v
}

func TestStats(t *testing.T) {
	srv := NewServer("rw_server.py")
	defer srv.Close()

	c := opcua.NewClient(srv.Endpoint, srv.Opts...)
	if err := c.Connect(context.Background()); err != nil {
		t.Fatal(err)
	}

	_ = c.Close()

	expected := map[string]*expvar.Int{
		"Dial":             newExpVarInt(1),
		"ActivateSession":  newExpVarInt(1),
		"NamespaceArray":   newExpVarInt(1),
		"UpdateNamespaces": newExpVarInt(1),
		"NodesToRead":      newExpVarInt(1),
		"Read":             newExpVarInt(1),
		"Send":             newExpVarInt(2),
		"Close":            newExpVarInt(1),
		"CloseSession":     newExpVarInt(2),
	}

	for k, ev := range expected {
		v := stats.Client.Get(k)
		verify.Values(t, k, v, ev)
	}
}
```

magiconair:
Unit tests might go a long way (cough) ... I'll add some.

magiconair:
I've refactored the code to use only one global variable to make it easier to reset the stats for the tests. Fixed the error type reflection and added tests.

magiconair:
@kung-foo can you have another look?

PR #534, opened 2022-01-03
client: do not clear session during reconnect

### LINKED ISSUE #503: The GOPCUA client crashed when the OPCUA server was restarted?
Hi, experts.
 I am running GOPCUA client (gopcua@v0.2.1/example/monitor/monitor.go) with OPCUA server(Siemens simatic WinccOA3.17 OPCUA server). When I restart(shutdown and re-open) OPCUA server, the GOPCUA client crashed, and printed out the following logs:

```
INFO: 2021/11/10 20:09:14 monitor_zyh.go:272: [channel 0] counter=166000
[signal 0xc0000005 code=0x0 addr=0x18 pc=0x5aa5f6]
 
goroutine 13 [running]:
github.com/gopcua/opcua.(*Client).ActivateSession(0xc0001c81a0, 0x0, 0x0, 0x0)
        D:/GoWorks/pkg/mod/github.com/gopcua/opcua@v0.2.1/client.go:649 +0x56
github.com/gopcua/opcua.(*Client).monitor(0xc0001c81a0, 0x711200, 0xc00003d800)
        D:/GoWorks/pkg/mod/github.com/gopcua/opcua@v0.2.1/client.go:345 +0xc66
created by github.com/gopcua/opcua.(*Client).Connect.func1
        D:/GoWorks/pkg/mod/github.com/gopcua/opcua@v0.2.1/client.go:193 +0x6a
 ```
 
My question:
1. What can I do to solve this problem?
 
 
 
2. There is no exception(OPCUA server shutdown) raised in gopcua@v0.2.1, although I already have set "SetErrorHandler" to client monitor handle.
But, I can  catch this exception in gopcua@v0.1.13 by "SetErrorHandler", and reconnect OPCUA server, monitor node.
What should I do to catch this situation in the client side with gopcua@v0.2.1?

### LINKED ISSUE #538: v0.2.4 and v0.2.5 fail with StatusBadSessionIDInvalid
It looks like #534 created a regression since our code fails with `StatusBadSessionIDInvalid`.

v0.2.3 works. Need to investigate

magiconair:
This patch checks if the session to be restored exists before trying to
restore it. It also prevents the internal session variable from being
cleared during a reconnect since the client would then try to restore
the session only once.

Fixes #503

### DIFF
--- a/client.go
+++ b/client.go
@@ -360,12 +360,14 @@ func (c *Client) monitor(ctx context.Context) {
 						// This only works if the session is still open on the server
 						// otherwise recreate it
 
-						dlog.Printf("trying to restore session")
-						s, err := c.DetachSession()
-						if err != nil {
-							action = createSecureChannel
+						s := c.Session()
+						if s == nil {
+							dlog.Printf("no session to restore")
+							action = recreateSession
 							continue
 						}
+
+						dlog.Printf("trying to restore session")
 						if err := c.ActivateSession(s); err != nil {
 							dlog.Printf("restore session failed")
 							action = recreateSession


magiconair:
this seems to have a regression. Our code fails with `StatusBadSessionIDinvalid`. Need to investigate.

magiconair:
Fixed in v0.2.6

Issue #550, opened 2022-01-24
Running examples/browse.go returns EOF error


Opsi:
I was trying to run the [browse.go example](https://github.com/gopcua/opcua/blob/main/examples/browse/browse.go) against my Prosys OPC UA Simulation Server as well as a running production server and every time I get an EOF error. I usually start the browsing at the root node `"i=84"`, but using the id of a FolderType node resulted in the same outcome. I also tried to run on a linux machine as well as my mac m1. The `io.EOF err` is returned whenever the code reaches [this line of the browse code](https://github.com/gopcua/opcua/blob/main/examples/browse/browse.go#L54) for the **second** time. After using a debugger it seems like the EOF error is created by [this line in the secure channel code](https://github.com/gopcua/opcua/blob/main/uasc/secure_channel.go#L689).

This is the corresponding debug output:

```
debug: uacp: connecting to opc.tcp://<SERVER-ADDRESS>
debug: uacp 1: start HEL/ACK handshake
debug: uacp 1: sent HELF with 84 bytes
debug: uacp 1: recv ACKF with 28 bytes
debug: uacp 1: server has no chunk limit. Using 512
debug: uacp 1: server has no message size limit. Using 2097152
debug: uacp 1: recv &uacp.Acknowledge{Version:0x0, ReceiveBufSize:0x2004, SendBufSize:0x2004, MaxMessageSize:0x200000, MaxChunkCount:0x200}
debug: uasc 1/1: send *ua.OpenSecureChannelRequest with 132 bytes
debug: uacp 1: recv OPNF with 135 bytes
debug: uasc 1/1: recv OPNF with 135 bytes
debug: uasc 1/1: recv *ua.OpenSecureChannelResponse
debug: uasc 1/1: sending *ua.OpenSecureChannelResponse to handler
debug: uasc 1: received security token. channelID=398608 tokenID=1 createdAt=2022-01-24T12:54:13Z lifetime=1h0m0s
debug: uasc 1/2: send *ua.CreateSessionRequest with 291 bytes
debug: uasc 1: security token expires at 2022-01-24T14:09:13Z. channelID=398608 tokenID=1
debug: uasc 1: security token is refreshed at 2022-01-24T13:39:11Z (45m0s). channelID=398608 tokenID=1
debug: uacp 1: recv MSGC with 7075 bytes
debug: uasc 1/2: recv MSGC with 7075 bytes
debug: uacp 1: recv MSGC with 7075 bytes
[...]

PR #551, opened 2022-01-25
[WIP] Fix hiding of errors in SecureChannel.readChunk

Opsi:
Related to #550 

`readChunk` of `secure_channel.go` was hiding the errors of `Conn.Receive`, because of a wrong comparison. `len(b) == 0` returns true whenever `b=nil`.

### DIFF
--- a/uasc/secure_channel.go
+++ b/uasc/secure_channel.go
@@ -341,8 +341,7 @@ func (s *SecureChannel) readChunk() (*MessageChunk, error) {
 	// read a full message from the underlying conn.
 	b, err := s.c.Receive()
 
-	sliceIsEmpty := b != nil && len(b) == 0
-	if err == io.EOF || sliceIsEmpty {
+	if err == io.EOF || err == nil && len(b) == 0 {
 		return nil, io.EOF
 	}
 	// do not wrap this error since it hides conn error
--- a/uasc/secure_channel.go
+++ b/uasc/secure_channel.go
@@ -340,7 +340,9 @@ func (s *SecureChannel) receive(ctx context.Context) *response {
 func (s *SecureChannel) readChunk() (*MessageChunk, error) {
 	// read a full message from the underlying conn.
 	b, err := s.c.Receive()
-	if err == io.EOF || len(b) == 0 {
+
+	sliceIsEmpty := b != nil && len(b) == 0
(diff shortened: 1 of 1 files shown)


magiconair:
Good catch. I would suggest to fix it slightly differently since `b != nil` is a side effect of `err == nil` in this case.

```go
if err == io.EOF || err == nil && len(b) == 0 {
    return nil, io.EOF
}
```

magiconair:
Hmm, on my local branch the `TestAutoReconnection/subscription_failure` hangs all of a sudden.

magiconair:
Hmm, I think this is surfacing a different issue

magiconair:
have the same issue when I'm applying that patch on v0.3.0

magiconair:
I can't see it. I'll push this to v0.3.2 since this needs more investigation.

Opsi:
Sorry I totally overlooked the integration tests. The problem probably has to do with [`SecureChannel.dispatcher`](https://github.com/gopcua/opcua/blob/main/uasc/secure_channel.go#L187)

```go
func (s *SecureChannel) dispatcher() {
    ctx, cancel := context.WithCancel(context.Background())
    defer cancel()

    defer func() {
        close(s.disconnected)
    }()

    for {
        select {
        case <-s.closing:
            return
        default:
            resp := s.receive(ctx) // this is now returning non-io.EOF errors...
                
            if resp.Err != nil {
                select {
                case <-s.closing:
                    return
                default:
                    select {
                    case s.errCh <- resp.Err:
                    default:
                    }
                }
            }
                
            if resp.Err == io.EOF { // therefore this is NOT returning from the dispatcher loop
                return
            }
            ...
```

magiconair:
Hmm, when I run the test in isolation then it works every time. But not when I run the entire suite.

magiconair:
@Opsi can you rebase your branch and fix the integration tests?

Opsi:
Rebasing is no problem, but I don't if I can fix the integration tests.

1) There seems to be some kind of race condition in `uatest/reconnection_test.go TestAutoReconnection/subscription_failure` that's (most of the time) causing an endless loop for me on the fourth run (where a "subscription_failure" is simulated)

2) I wasn't able to look into the second error in `uatest/timeout_test.go TestClientTimeoutViaContext`

PR #564, opened 2022-03-08
Add support for arrays

siscia:
Any reason why this was not implemented?

We have a use case with static size array, and it was returning an error.

### DIFF
--- a/ua/decode.go
+++ b/ua/decode.go
@@ -128,7 +128,7 @@ func decodeSlice(b []byte, val reflect.Value, name string) (int, error) {
 	}
 
 	if n > math.MaxInt32 {
-		return buf.Pos(), errors.Errorf("array too large: %d", n)
+		return buf.Pos(), errors.Errorf("array too large: %d > %d", n, math.MaxInt32)
 	}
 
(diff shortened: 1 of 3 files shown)


magiconair:
Don't recall. Can you add a test?

siscia:
Do we have tests for the encoding part of the library?

I can't find them.

magiconair:
`ua/decode_test.go` it does both

siscia:
Found it!

I also implement the decoding!

magiconair (changes requested):
I think the encode and decode functions can be the same for both slice and array and the type switch can be merged to `case reflect.Slice, reflect.Array:` to make it obvious. Maybe rename the functions to `writeSliceOrArray` and similar. The tests will ensure that it works.

magiconair (approved):
After looking at #566 which merges the functions it became clear that this approach doesn't increase readability. There are too many subtle differences. So we go with separate functions for encoding slices and arrays.

I've found a fast path for byte arrays, updated the tests and made the error messages consistent. 

LGTM now

PR #576, opened 2022-04-29
ua: log unknown extension object type id

magiconair:
Print the type id of an unknown extension object to the debug log.


magiconair:
Ideally this should be in the error message of `StatusBadDataTypeIDUnknown` but that requires to change the `StatusCode` type from a `uint32` to a struct. That's a different change.

PR #577, opened 2022-05-02
Avoid nil pointer error

dza1:
when encoding a FieldMetaData, a nil pointer error occurred,
if DataSetFieldID and Description are uninitialized

### DIFF
--- a/ua/datatypes.go
+++ b/ua/datatypes.go
@@ -153,6 +153,9 @@ func (g *GUID) Decode(b []byte) (int, error) {
 }
 
 func (g *GUID) Encode() ([]byte, error) {
+	if g == nil {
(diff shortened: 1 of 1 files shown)


magiconair:
Hi, thank you for your first submission to the project. 

However, the way we normally solve this problem is to provide empty objects as for example here: https://github.com/gopcua/opcua/blob/main/ua/read_request_test.go#L32

So you need to init your `FieldMetaData` as follows:

```go
d := &FieldMetaData{
    // ... other fields
    Description: &ua.LocalizedText{},
    DataSetFieldID: &ua.GUID{},
}
```

I think we've discussed this before but if we would do this then we would have to do this for all types and not just the two you have in the PR.

PR #585, opened 2022-06-07
Change encryption URI for aes128Sha256RsaOaep to w3.org

### LINKED ISSUE #584: UserIdentityToken EncryptionAlgorithm RSA-OAEP-SHA1 uses non-existent URL
Hello,

I use the Telegraf plugin for OPC UA and want to connect to a OPC UA of a supplier of ours. Their server only support `http://opcfoundation.org/UA/SecurityPolicy#Aes128_Sha256_RsaOaep` as the SecurityPolicyUri for the UserTokenPolicy.  The supplier has used the [open62541](https://github.com/open62541/open62541) library.

When the gopcua library encrypts the ActivateSessionRequest user password, it indicates it uses the `http://opcfoundation.org/ua/security/rsa-oaep-sha1` algorithm to encrypt the password, whereas uaExpert claims to use `http://www.w3.org/2001/04/xmlenc#rsa-oaep`. 

It looks like this may be the cause of the problem, as we get a `0x80200000 [BadIdentityTokenInvalid]` response from the server. 

Sometimes the `w3.org` url is referenced in the encyrption algorithms, and sometimes the `opcfoundation.org`. The opcfoundation.org url, does not resolve to anything.

I have compiled a version of Telegraf with the fix below, and the connection worked properly, including the password encryption. 

Will you accept a PR to with this change?

```diff
github.com\gopcua\opcua@v0.3.1\uapolicy\policyAes128Sha256RsaOaep.go
- encryptionURI:         "http://opcfoundation.org/ua/security/rsa-oaep-sha1",
+ encryptionURI:         "http://www.w3.org/2001/04/xmlenc#rsa-oaep",
```

LarsStegman:
Fix #584

### DIFF
--- a/uapolicy/policyAes128Sha256RsaOaep.go
+++ b/uapolicy/policyAes128Sha256RsaOaep.go
@@ -115,7 +115,7 @@ func newAes128Sha256RsaOaepAsymmetric(localKey *rsa.PrivateKey, remoteKey *rsa.P
 		nonceLength:           nonceLength,
 		signatureLength:       localKeySize,
 		remoteSignatureLength: remoteKeySize,
-		encryptionURI:         "http://opcfoundation.org/ua/security/rsa-oaep-sha1",
+		encryptionURI:         "http://www.w3.org/2001/04/xmlenc#rsa-oaep",
 		signatureURI:          "http://www.w3.org/2001/04/xmldsig-more#rsa-sha256",
 	}, nil
 }


LarsStegman:
@magiconair @dwhutchison Have you had time to review this?

magiconair:
I'll try to do that today. In the middle of a move.

dwhutchison:
I was able to reproduce the error on the unpatched version with open62541 and this patch fixed it. 
I also tested against the Ignition (Eclipse Milo) and Prosys servers and they accepted both versions.
If I get some more time I’ll try test against the .Net reference server but I think this patch is good.

LarsStegman:
That's great! Thank you.

magiconair:
Released as v0.3.5

PR #589, opened 2022-06-29
Update the schema to v1.05.02-2022-11-01 and regenerate code

gstvg:
- stringer: use go install instead of go get
- skip enums with no values, e.g. the abstract Enumeration
- remove Bad/Uncertain, already included on status.csv and remove quote
- import new package location
- remove unexistent stats cmd call and stringer, switch stringer to go install
- sh raises error with set -o pipefail, switch to bash
- run update-schema.sh
- run make gen
- go mod tidy for new crypto example lib import location


magiconair:
I've rebased it to current `main`

magiconair:
@gstvg I think your changes are good as is. Nice cleanup and it's time for us to update the schema. I'll use this as an excuse to bump to 0.4

magiconair (approved):

PR #596, opened 2022-08-11
set SecureChannel nil in Close() method

hotafrika:
I was trying to implement this flow `Connect() -> Read data -> Disconnect network from host -> Close() -> Connect() in loop` but I was always receiving `opcua: already connected`.
Client has option `opcua.AutoReconnect(false)`.
So I noticed that SecureChannel exists even after Close() method call.

### DIFF
--- a/client.go
+++ b/client.go
@@ -563,6 +563,7 @@ func (c *Client) CloseWithContext(ctx context.Context) error {
 	}
 	if c.SecureChannel() != nil {
 		c.SecureChannel().Close()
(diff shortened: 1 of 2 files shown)


hotafrika:
@kung-foo Hi! Please rerun workflows. I forgot to check it in my fork repo. Now it is OK (I checked)!

pavel-fokin:
@kung-foo Hi! This PR should solve an issue with reconnection. Please, review when you have time.

magiconair (approved):

PR #600, opened 2022-08-17
Fix small typo in ua/variant

EJTJ3:


kung-foo:
Thanks for the patch!

PR #608, opened 2022-10-05
uasc: stop expiration timer

magiconair:
We have a memory leak with some code and it points to lots of timers. @kung-foo did we forget to stop a timer here or is there a good reason not to stop the timer?


magiconair:
image

magiconair:
I'll merge this since I don't see any reason why this timer shouldn't be stopped. @kung-foo pls review in any case. We can always revert.

kung-foo:
Seems like we just missed it. However I'm curious how this would leak. What were your symptoms?

kung-foo:
Ah, now I see your symptoms. Inflight Wi-Fi is flakey :)

PR #617, opened 2022-11-14
fix decode NodeID

huskar-t:
When I use `prosys OPC UA Simulation Server` as server, I have the same problem as #550, I borrowed from [python-opcua](https://github.com/FreeOpcUa/python-opcua/blob/master/opcua/ua/ua_binary.py#L333-L361) and fix the Decode method of NodeID, which seems to solve my problem.

### DIFF
--- a/ua/node_id.go
+++ b/ua/node_id.go
@@ -302,39 +302,42 @@ func (n *NodeID) String() string {
 
 func (n *NodeID) Decode(b []byte) (int, error) {
 	buf := NewBuffer(b)
-
-	n.mask = NodeIDType(buf.ReadByte())
-	typ := n.mask & 0xf
+	encoding := buf.ReadByte()
+	typ := NodeIDType(encoding & 0b00111111)
 
 	switch typ {
 	case NodeIDTypeTwoByte:
 		n.nid = uint32(buf.ReadByte())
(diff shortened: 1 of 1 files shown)


huskar-t:
@kung-foo @magiconair Please review, I really need this fix

magiconair:
Sure. Can you please remove the `_ =` since we don't do that. Can you also look up the relevant part of the spec and add a URL to the code.

magiconair:
Also, please add some tests. No merge without. Thank you.

huskar-t:
> Sure. Can you please remove the `_ =` since we don't do that. Can you also look up the relevant part of the spec and add a URL to the code.

I can't find the client protocol specification on the web, can you give me some specifications or websites that you use?

magiconair:
@huskar-t You should find the `NodeID` specification somewhere in here: This is a link to the `ExpandedNodeId` but you should find the TCP/IP binary representation model somewhere in here as well.

https://reference.opcfoundation.org/v104/Core/docs/Part4/7.11/

huskar-t:
@magiconair Thank you very much, but after reading the code it seems that I made a mistake, I will solve this problem as soon as possible. Maybe close this pr first and create a new one after I solve it

magiconair:
Sounds good.

PR #628, opened 2023-01-15
Prevent hanging on TCP connections that won't return [any/full] OPC UA data

jb0n:
Great lib! I was using it to look at some interesting OPC UA servers I found on the internet (scanning for them truth be told) and ran into situations where io.ReadFull hangs forever. I made some changes and allowed the Dialer to have a read timeout it can pass to connections.

### DIFF
--- a/config.go
+++ b/config.go
@@ -508,6 +508,12 @@ func ReadTimeout(d time.Duration) Option {
 		cfg.dialer.ReadTimeout = d
 	}
 }
+
+// WriteTimeout sets the timeout for every write operation.
+func WriteTimeout(d time.Duration) Option {
+	return func(cfg *Config) {
+		initDialer(cfg)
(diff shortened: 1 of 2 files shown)


magiconair (dismissed):
Thank you. Can you just use a `time.Duration` and check if it is greater than zero instead of using a pointer? 

Also, can you squash your commits please?

Then I'll merge it.

magiconair:
I've rebased your changes onto `main`, squashed them and also added support for write timeouts. I've also cleared the deadline after the read/write. Not sure if this is necessary. Could you please review and test this change?

@kung-foo it would be great if you could have a look at this as well.

I've tagged this for v0.4 since this may have some wider implications.

jb0n:
Hey thanks for the reply. I had a few things I was going to take out of
that CR, I'll do that as well as implement your suggestion.

PR #629, opened 2023-01-19
uacp: honor the context deadline during the handshake

magiconair:
This patch honors the `context` deadline during the `uacp` handshake so that when the server accepts the connection but does not respond to the `HEL` request in time the connection gets cancelled.

### DIFF
--- a/uacp/conn.go
+++ b/uacp/conn.go
@@ -11,6 +11,7 @@ import (
 	"net"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


magiconair:
I am wondering if we can write a test for that ...

magiconair:
We tested that internally. Merging

kung-foo:
FWIW, we tested this internally using `$ nc -l 50000`.

PR #634, opened 2023-02-01
ua: decode types that convert to time.Time

### LINKED ISSUE #633: Decoder fails to decode type which converts to time.Time
The decoder will fail to decode `time Timestamp time.Time` as `time.Time`. The problem is in the code that detects a time type which is too strict. We need to use `reflect.CanConvert`

magiconair:
The codec does not handle types that can be converted to time.Time correctly since it is too strict in determining what a time.Time type is. This patch relaxes this by checking if a type can be converted to time.Time.

Fixes #633

### DIFF
--- a/ua/decode.go
+++ b/ua/decode.go
@@ -23,7 +23,7 @@ func isBinaryDecoder(val reflect.Value) bool {
 }
 
 func isTime(val reflect.Value) bool {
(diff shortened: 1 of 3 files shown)


magiconair:
Note that I tested this on our codebase. Will merge since it is blocking.

PR #637, opened 2023-02-22
Add ability to access dynamic datatypes

### LINKED ISSUE #588: add functionality for registering maps for Extension objects
Add function to register extension objects with server stored data types.
Registering a node using new method will decode data type into a map[string]interface{}
When reading or subscribing the returned value will be written into the registered map[string]interface{} structure

joeACV:
Some OPC values are data types defined in the server.  This will allow you to register the data type value and read or subscribe to it and be encoding into a map[string] interface{} allowing it to be a dynamic data structure
closes #588


magiconair:
I have a similar requirement for this and wrote some internal code which uses `reflect` to create a type on the fly. So we can check which approach works better. 

A couple of comments on the PR:

* Could you please clean up the commits? You have provided us with 20 commits which are rebases and internal changes that don't really belong here. 
* Please clean up the code and the comments. Line comments start with `//<space>...` for example
* Please remove unused code.
* Try to stay away from `if ... else` and see if `switch` is a better pattern. Using `else` is somewhat of an anti-pattern in Go. I'm sure there is more.

In general, please provide more rationale and context for the change. Explain the approach. Think about how you would like to see contributions to your code from someone you haven't seen before. I appreciate your effort and want to review it. However, this library is in critical active production use. It runs in a lot of places. Let your code quality reflect that please.

Thank you
Frank

joeACV:
Thanks Frank for the comments.
I will close the pull request until I get all of your recommendations completed.

PR #638, opened 2023-02-23
Create a subscription to prevent reconnection infinite loop

DVasselli:
##Issue
The main problem is that if we have a long running connection, with little operations, when the next request comes, the session will have been expired. The autoreconnection logic assumes there are subscription to be restarted, but when this is attempted, the restart errors out since there is nothing to be restarted. This triggers the reconection loop again, ending in a forever running loop.

##Fix 
1. Add a default subscription to prevent this loop
2. Add a monitor to the time node for this subscription. The reason why we need it is that without it the session will timeout once every minute or so and this would trigger the reconnection logic, thus creating unneeded load on the server.

### DIFF
--- a/client.go
+++ b/client.go
@@ -244,6 +244,28 @@ func (c *Client) Connect(ctx context.Context) (err error) {
 		return err
 	}
 
+	if c.cfg.sechan.AutoReconnect {
+		notifCh := make(chan *PublishNotificationData)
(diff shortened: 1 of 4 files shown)


magiconair:
Thank you @DVasselli. While this may address the problem I think this is the wrong solution. We should fix the reconnection logic. It should be possible to detect that there are no subscriptions to prevent the loop.

DVasselli:
I am closing this since this solution, although fixed the issue, is a hacky approach and should not be used

PR #648, opened 2023-04-20
fix: handle extra padding if key length > 2048

Maddin-619:
Connecting to a opc ua server with a 4096 bit key length gave me an error. The server could not decode the `OpenSecureChannelRequest` because the `ExtraPaddingSize` byte was missing.

If the key length is greater than 2048 bits there should be an `ExtraPaddingSize` byte in the message footer:
[UA Part 6: Mappings - 6.7.2.5 Message Footer](https://reference.opcfoundation.org/Core/Part6/v105/docs/6.7.2.5)

The PR should fix that problem while encoding and decoding messages.

I implemented the fix in accordance with the [c reference implmentation (open62541)](https://github.com/open62541/open62541/blob/master/src/ua_securechannel_crypto.c#L403)


uasc/secure_channel_instance.go
```diff
@@ -170,11 +170,21 @@ func (c *channelInstance) signAndEncrypt(m *Message, b []byte) ([]byte, error) {
 	var encryptedLength int
 	if c.sc.cfg.SecurityMode == ua.MessageSecurityModeSignAndEncrypt || isAsymmetric {
 		plaintextBlockSize := c.algo.PlaintextBlockSize()
-		paddingLength := plaintextBlockSize - ((len(b[headerLength:]) + c.algo.SignatureLength() + 1) % plaintextBlockSize)
+		extraPadding := c.algo.RemoteSignatureLength() > 256
+		var paddingBytes int
```
magiconair: Could you change this to avoid `else` please?

```go
paddingBytes := 1
if extraPadding {
    paddingBytes = 2
}
```

magiconair (changes requested):
Thank you! Could you also please add a test for this? Then I'll merge it.

Maddin-619:
It's not clear for me how and where to write tests for the secure_channel_instance. There are only a few tests for the `newRequestMessage` method. To test `signAndEncrypt` and `verifyAndDecrypt` we need test fixtures for the message and the channelInstance with an `EncryptionAlgorithm` (with mocked encrypt, decrypt, signature and verify functions). That is quite a lot to set up before writing the test that I need a little help with.

magiconair:
Hmm, fair enough. Let me have a look

magiconair:
I don't think this is so bad since `signAndEncrypt` is mostly self-contained. You only need to configure a channel instance with the right algorithm and security mode and a bare-bones message for this to work. Sometimes I factor this kind of code out into a standlone function to make it more testable. 

Maybe start with something like this (code does NOT compile :))

```go
func TestSignAndEncrypt(t *testing.T) {
    tests := []struct{
        name string
        c *channelInstance
        m *Message
        b []byte
        want []byte
        err error
    }{
        {"none", 
            &channelInstance{
                sc: SecureChannel{cfg: Config{SecurityMode: None}}
            },
            ...
        }
     }


    for _, tt := range tests {
        t.Run(tt.name, func(t *testing.T) {
            signed, err := tt.c.signAndEncrypt(tt.m, tt.b)
            if got, want := err.String(), tt.err.String(); got != want {
                t.Fatalf("got error %v want %v", got, want)
            }

            if got, want := signed, tt.want; !bytes.Equal(got, want) {
                ...   
            }
        }
    }
}

// create some helper funcs to create ciphertext from plaintext
// to make testing easier
```

magiconair:
Also, `EncryptionAlgorithm` already has `Encrypt` and `Decrypt` methods so there is no mocking IMHO. Just test with some real algorithms.

magiconair:
If it makes it easier you could put the golden plaintext data into `testdata` folder and use `embed` to pull it in.

magiconair:
@Maddin-619 let me know if that works for you or if you need more help.

Maddin-619:
I have problems using the real algorithems, because the encryption always leads to a different ciphertext. So we can't assert with static bytes.
I see two possible solutions:

1. using the `verifyAndDecrypt` for the assertion getting back the plaintext
2. mock the encrypt method s.t. plaintext = ciphertext

In general it should be better to isolate the testee without the dependencies. So I would prefer solution 2.

magiconair (approved):

PR #652, opened 2023-05-22
Remove 'if err == nil' anti-pattern

magiconair:
Using 'if err == nil' is an anti-pattern in Go and should be avoided since most if not all Go developers expect 'if err != nil {...}'

This patch flips/removes the checks in the main client code but leaves the checks in the test cases were we require an error.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1030,24 +1030,25 @@ func (c *Client) ReadWithContext(ctx context.Context, req *ua.ReadRequest) (*ua.
 	err := c.SendWithContext(ctx, req, func(v interface{}) error {
 		err := safeAssign(v, &res)
 
+		if err != nil {
+			return err
+		}
+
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -1030,24 +1030,25 @@ func (c *Client) ReadWithContext(ctx context.Context, req *ua.ReadRequest) (*ua.
 	err := c.SendWithContext(ctx, req, func(v interface{}) error {
 		err := safeAssign(v, &res)
 
+		if err != nil {
+			return err
+		}
```
wmnsk:
```suggestion
		err := safeAssign(v, &res)
		if err != nil {
			return err
		}
```

I don't think Go devs usually put a blank line here (or even do like `if err := safeAssign(v, &res); err != nil { ... }`).

wmnsk (approved):
Nice work!

PR #653, opened 2023-05-22
Improve error handling

bitomaxsp:
- Handle mode errors while debugging to give better logging
- Set connection status when recreating sessions
- Expose RevisedSessionTimeout on Session
- Fix subscriotion.go example to set endpoint to connect to
- Add read retry logic based on what error do we  (Northvolt) see on production PLC


magiconair (approved):
LGTM.

PR #657, opened 2023-06-01
Bugfix: Close session properly if activation failed

Maddin-619:
At the moment if the session activation failed the session dosn't get closed. Depending on the server session timeout there can accumulate lots of danggling sessions.

To close the session after a activation failure the session must be set before calling the `Close` function.

### DIFF
--- a/client.go
+++ b/client.go
@@ -222,6 +222,7 @@ func (c *Client) Connect(ctx context.Context) error {
 	}
 
 	if err := c.ActivateSessionWithContext(ctx, s); err != nil {
+		c.setSession(s)
 		c.CloseWithContext(ctx)
 		stats.RecordError(err)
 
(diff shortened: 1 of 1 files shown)


magiconair:
Thank you! I've refactored the change to make closing the session explicit and not just a side effect of setting the session. However, we still call `setSession` in the `ActivateSession` handler and there is a side-effect here. Question is at which point should the `session` variable contain a valid value. After creation or after activation? I'll merge it and release a v0.3.16 shortly together with your other fix.

magiconair (approved):

magiconair:
This will go into v0.4.0

Maddin-619:
@magiconair In the current version the bug occurs again. A good PR should have include some tests, damn it. Sorry about that.
It's basically the same problem as the last time: If `ActivateSession` fails, the session is not stored in the atomic variable and therefore cannot be closed. Are you good if i submit another PR? If yes do you prefer the explicit way again?

magiconair:
@Maddin-619 don't beat yourself up. A PR is always welcome. Thank you! 🥰

PR #662, opened 2023-06-13
v0.4.0

magiconair:
* Clarify release policy (support current and previous Go version)
* Use golang.org/exp/x/slices where appropriate
* Update dependencies
* Mark `v0.2.4` and `v0.2.5` as `retracted`

Since this version requires generics I've cleaned up the README to remove references to specific Go versions and instead just kept the statement that we support the current and the previous Go version.

I would also like to use this as a starting point to make the breaking API changes for v0.5.0 soon-ish.


bitomaxsp (approved):
Nice

PR #664, opened 2023-06-14
uasc: return an error for invalid uri/mode combinations with None

magiconair:
The previous code forced security mode None if the URI was None changing
the incoming Config object. While it makes the code more robust it also
has a side effect and we were already checking for other invalid
combinations. This patch puts all checks in one place and returns an
error if the combination is invalid.

### DIFF
--- a/uasc/secure_channel.go
+++ b/uasc/secure_channel.go
@@ -129,19 +129,17 @@ func NewSecureChannel(endpoint string, c *uacp.Conn, cfg *Config, errCh chan<- e
 		return nil, errors.Errorf("no secure channel config")
 	}
 
-	if cfg.SecurityPolicyURI != ua.SecurityPolicyURINone {
-		if cfg.SecurityMode == ua.MessageSecurityModeNone {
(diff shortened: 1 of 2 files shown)


magiconair:
I've added https://github.com/stretchr/testify as a dependency for testing. I'm open for using something else like https://github.com/gotestyourself/gotest.tools if there are strong opinions. What do others think?

Update: just in the same file we already use the `verify` package and `if err != nil { t.Fatal(err) }` pattern. Adding `require` would add another pattern for testing the same things which probably adds more confusion than it is worth. 

I've implemented the `errorContains` function which I've needed for the test and removed the stretchr dependency again.

kung-foo:
FWIW, I really like the testify (require) package and we use it in 100% of our internal go projects.

kung-foo (approved):
LGTM

magiconair:
We use testify as well. For this one I was more concerned with having three different testing strategies in the codebase. Didn't seem worth the effort.

PR #674, opened 2023-08-21
refactor: make NewClient return an error

### LINKED ISSUE #621: Return error on NewClient
See #616 for details. Planned for v0.5.0

sruehl:
This PR addresses the TODO `Note: Starting with v0.5 this function will return an error.` and let NewClient return an error.

Note on the PR: I use `github.com/stretchr/testify` for error assertion with seems IMHO pretty standard in the go projects I worked with. If this is unwanted I could rework this.

resolves https://github.com/gopcua/opcua/issues/621

### DIFF
--- a/client.go
+++ b/client.go
@@ -30,7 +30,10 @@ import (
 // GetEndpoints returns the available endpoint descriptions for the server.
 func GetEndpoints(ctx context.Context, endpoint string, opts ...Option) ([]*ua.EndpointDescription, error) {
 	opts = append(opts, AutoReconnect(false))
-	c := NewClient(endpoint, opts...)
+	c, err := NewClient(endpoint, opts...)
+	if err != nil {
+		return nil, err
+	}
 	if err := c.Dial(ctx); err != nil {
 		return nil, err
 	}
@@ -133,15 +136,6 @@ type Client struct {
 
 	// monitorOnce ensures only one connection monitor is running
 	monitorOnce sync.Once
-
-	// cfgerr contains an error that was captured in ApplyConfig.
-	// Since the API does not allow to bubble the error up in NewClient
-	// and we don't want to break existing code right away we carry the
-	// error here and bubble it up during Dial and Connect.
-	//
-	// Note: Starting with v0.5 NewClient will return the error and this
-	// variable needs to be removed.
-	cfgerr error
 }
 
 // NewClient creates a new Client.
@@ -155,10 +149,11 @@ type Client struct {
 // #Option for details.
 //
 // https://godoc.org/github.com/gopcua/opcua#Option
-//
-// Note: Starting with v0.5 this function will return an error.
-func NewClient(endpoint string, opts ...Option) *Client {
-	cfg := ApplyConfig(opts...)
+func NewClient(endpoint string, opts ...Option) (*Client, error) {
+	cfg, err := ApplyConfig(opts...)
+	if err != nil {
+		return nil, err
+	}
 	c := Client{
 		endpointURL: endpoint,
 		cfg:         cfg,
@@ -167,7 +162,6 @@ func NewClient(endpoint string, opts ...Option) *Client {
 		pendingAcks: make([]*ua.SubscriptionAcknowledgement, 0),
 		pausech:     make(chan struct{}, 2),
 		resumech:    make(chan struct{}, 2),
-		cfgerr:      cfg.Error(), // todo(fs): remove with v0.5.0 and return the error
 	}
 	c.pauseSubscriptions(context.Background())
 	c.setPublishTimeout(uasc.MaxTimeout)
@@ -175,7 +169,7 @@ func NewClient(endpoint string, opts ...Option) *Client {
 	c.setSecureChannel(nil)
 	c.setSession(nil)
 	c.setNamespaces([]string{})
-	return &c
+	return &c, nil
 }
 
 // reconnectAction is a list of actions for the client reconnection logic.
@@ -194,11 +188,6 @@ const (
 
 // Connect establishes a secure channel and creates a new session.
 func (c *Client) Connect(ctx context.Context) error {
-	// todo(fs): remove with v0.5.0
-	if c.cfgerr != nil {
-		return c.cfgerr
-	}
-
 	// todo(fs): the secure channel is 'nil' during a re-connect
 	// todo(fs): but we expect this method to be called once during startup
 	// todo(fs): so this is probably safe
@@ -503,16 +492,16 @@ func (c *Client) monitor(ctx context.Context) {
 						// populated in the previous step.
 
 						activeSubs = 0
-						for _, id := range subsToRepublish {
-							if err := c.republishSubscription(ctx, id, availableSeqs[id]); err != nil {
-								dlog.Printf("republish of subscription %d failed", id)
-								subsToRecreate = append(subsToRecreate, id)
+						for _, subID := range subsToRepublish {
+							if err := c.republishSubscription(ctx, subID, availableSeqs[subID]); err != nil {
+								dlog.Printf("republish of subscription %d failed", subID)
+								subsToRecreate = append(subsToRecreate, subID)
 							}
 							activeSubs++
 						}
 
-						for _, id := range subsToRecreate {
-							if err := c.recreateSubscription(ctx, id); err != nil {
+						for _, subID := range subsToRecreate {
+							if err := c.recreateSubscription(ctx, subID); err != nil {
 								dlog.Printf("recreate subscripitions failed: %v", err)
 								action = recreateSession
 								continue
@@ -555,11 +544,6 @@ func (c *Client) monitor(ctx context.Context) {
 
 // Dial establishes a secure channel.
 func (c *Client) Dial(ctx context.Context) error {
-	// todo(fs): remove with v0.5.0
-	if c.cfgerr != nil {
-		return c.cfgerr
-	}
-
 	stats.Client().Add("Dial", 1)
 
 	if c.SecureChannel() != nil {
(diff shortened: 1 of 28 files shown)


sruehl:
@magiconair I worked on a TODO you defined in source. ~Not sure if there is a ticket for that. If I find it I link it here.~ (Edit: found it and linked)

client.go
```diff
@@ -30,11 +30,18 @@ import (
...
-	c := NewClient(endpoint, opts...)
+	c, err := NewClient(endpoint, opts...)
+	if err != nil {
+		return nil, err
+	}
 	if err := c.Dial(ctx); err != nil {
 		return nil, err
 	}
-	defer c.Close(ctx)
+	defer func(c *Client, ctx context.Context) {
```
magiconair: Let's not do this since there isn't anything we can do with that error anyway. If we really want to log it we should add something to the `Close` method.

client.go
```diff
@@ -215,14 +208,18 @@ func (c *Client) Connect(ctx context.Context) error {
 
 	s, err := c.CreateSession(ctx, c.cfg.session)
 	if err != nil {
-		c.Close(ctx)
+		if err := c.Close(ctx); err != nil {
```
magiconair: same

client.go
```diff
@@ -215,14 +208,18 @@ func (c *Client) Connect(ctx context.Context) error {
...
+			// TODO(sruehl): log error
+		}
 		stats.RecordError(err)
 
 		return err
 	}
 
 	if err := c.ActivateSession(ctx, s); err != nil {
-		c.Close(ctx)
+		if err := c.Close(ctx); err != nil {
```
magiconair: same

client.go
```diff
@@ -896,7 +910,7 @@ func (c *Client) closeSession(ctx context.Context, s *Session) error {
 // DetachSession removes the session from the client without closing it. The
 // caller is responsible to close or re-activate the session. If the client
 // does not have an active session the function returns no error.
-func (c *Client) DetachSession(ctx context.Context) (*Session, error) {
+func (c *Client) DetachSession(_ context.Context) (*Session, error) {
```
magiconair: Please leave `ctx` in there since it is part of the documentation.

config_test.go
```diff
@@ -167,9 +168,10 @@ func TestOptions(t *testing.T) {
 	defer os.Remove(keyPEMFile)
 
 	tests := []struct {
-		name string
-		opt  Option
-		cfg  *Config
+		name    string
+		opt     Option
+		cfg     *Config
+		wantErr assert.ErrorAssertionFunc
```
magiconair: `s/wantErr/err/` and also just make it the error you want.

client.go
```diff
@@ -503,16 +506,16 @@ func (c *Client) monitor(ctx context.Context) {
 						// populated in the previous step.
 
 						activeSubs = 0
-						for _, id := range subsToRepublish {
-							if err := c.republishSubscription(ctx, id, availableSeqs[id]); err != nil {
-								dlog.Printf("republish of subscription %d failed", id)
-								subsToRecreate = append(subsToRecreate, id)
+						for _, subId := range subsToRepublish {
```
magiconair: It should be `subID` but I don't see how this makes it more readable?

magiconair (changes requested):
Thank you! I've missed that part with the move to v0.5. Please do not check the error for the `c.Close()` method everywhere since we cannot do anything with that error anyway. There is no way to recover. We just have this in the signature to be consistent. 

As for `testify`. I went down that route before and decided against this since it would introduce the third testing framework into the code base. Yes, it is more convenient but it doesn't justify the cost of having different code styles in different parts of the code base. So please refactor this to look like the rest (i.e. `got`,`want` style and maybe use `verify.Values`)

client.go
sruehl: ok got a linter warning for that. Should I don't touch it at all or replace it with `_ = c.Close(ctx)`?

client.go
sruehl: it does overshadow the id import at top of the file. Again linter error, but I can leave it as is.

client.go
sruehl: got a unused warning as it is not used, therefore removed it. Reverting it...

client.go
sruehl: image

magiconair: Please no. Fix your linter. This is a cargo cult being pushed on the Go community by people who want to change the Go code style through the back door. There is zero readability advantage and that's the main thing we are trying to achieve. It just adds unnecessary clutter.

client.go
magiconair: Same comment about over-aggressive linter settings from people coming from other languages.

client.go
magiconair: Ah, then this is a good catch. Please change to `subID` then.

sruehl:
Oh and before I forget, take a closer look (@magiconair) what I did with `errors`. So I collect the errors there and return them with join. Could also be done via early exit. Also I wrap errors with `%w`. Maybe this is unwanted too.

magiconair:
I'll have a look. As a guiding principle I prefer consistency when we make changes like this. I'm OK with changing direction but then let's do it all over the place. And in a separate PR.

client.go
magiconair: This is still `subId`

client.go
```diff
@@ -754,11 +738,15 @@ func (c *Client) CreateSession(ctx context.Context, cfg *uasc.SessionConfig) (*S
 		// Ensure we have a valid identity token that the server will accept before trying to activate a session
 		if c.cfg.session.UserIdentityToken == nil {
 			opt := AuthAnonymous()
-			opt(c.cfg)
+			if err := opt(c.cfg); err != nil {
```
magiconair: revert

client.go
```diff
@@ -754,11 +738,15 @@ func (c *Client) CreateSession(ctx context.Context, cfg *uasc.SessionConfig) (*S
...
+				// TODO(sruehl): log error
+			}
 
 			p := anonymousPolicyID(res.ServerEndpoints)
 			opt = AuthPolicyID(p)
-			opt(c.cfg)
+			if err := opt(c.cfg); err != nil {
+				// TODO(sruehl): log error
```
magiconair: revert

client.go
sruehl: FYI: opt returning an error now, so here in that case that was me not the linter, but will do

client_test.go
```diff
@@ -11,8 +11,9 @@ import (
 )
 
 func TestClient_Send_DoesNotPanicWhenDisconnected(t *testing.T) {
-	c := NewClient("opc.tcp://example.com:4840")
-	err := c.Send(context.Background(), &ua.ReadRequest{}, func(i interface{}) error {
+	c, err := NewClient("opc.tcp://example.com:4840")
+	verify.Values(t, "no error", err, nil)
```
magiconair: We use `verify.Values` mostly as a replacement for `reflect.DeepEqual`

```go
if err != nil { 
    t.Fatal(err) 
}
```

client.go
magiconair: OK, can we actually error here? Will this fail to create the session? If yes, then we should return the error.

client.go
sruehl: Whoops, sorry, could have sworn I changed it, somehow got lost duing my switches...

client.go
magiconair: If you want to add a `todo` then I suggest the following:

```go
// todo(sr): opt returns an error but we concluded that this call cannot 
// todo(sr): fail and that we do not want to stop creating the session
// todo(sr): hence we ignore it.
```

Don't just write `log.Error` since the next person don't know whether that is critical or not and if we do we spam other people's log files with stuff they cannot change.

sruehl:
> I'll have a look. As a guiding principle I prefer consistency when we make changes like this. I'm OK with changing direction but then let's do it all over the place. And in a separate PR.

Clarifying question on that: I assumed that comment was related to wrapping with %w. Should I switch from errrors join to early exit too?

config.go
```diff
@@ -80,68 +68,76 @@ func NewDialer(cfg *Config) *uacp.Dialer {
...
 // todo(fs): Can we find a better name?
-//
-// Note: Starting with v0.5 this function will return an error.
-func ApplyConfig(opts ...Option) *Config {
+func ApplyConfig(opts ...Option) (*Config, error) {
 	cfg := &Config{
 		sechan:  DefaultClientConfig(),
 		session: DefaultSessionConfig(),
 	}
+	var foundErrors []error
```
magiconair: s/foundErrors/errs/

magiconair (changes requested):
Final nitpick about a variable name but then this LGTM.

config.go
magiconair: Tried to fix this via codespaces (which is actually quite nice) but I couldn't push to your PR. Can you fix this so that we can merge?

sruehl: hmm, ok don't know how that is supposed to work with codespaces. I set it to be allowed to be edited by maintainers so not sure what else I need to set. Usually I also use the suggestion feature right besides the heading selector: 
image
Fixed that in the latest push.

magiconair (approved):

client.go
sruehl: I would suggest adding your suggested TODO. For the first call it might fail eventually once this TODO here gets fixed 
```
if !ok {
	// todo(fs): should we Fatal here?
	log.Printf("non-anonymous authentication already configured, ignoring")
	return nil
}
```
In case you decide to return the error there you must remember to handle the error in the call we talk about.

magiconair:
As an aside: I've done the force-push update as well earlier in my career and while that keeps the commit cleaner it makes reviewing harder since I don't see what you have changed and always have to review the full change set. I've come to settle on individual commits and leave it to the maintainer to squash the commits if they want.

sruehl:
Usually I don't force push either but I read somewhere that it was requested to squash commits before merge so I started doing that from the start. But yes I agree: that makes following changes hard

magiconair:
On this project? I'll have a look or can you point me to it...

sruehl:
Maybe my mind tricked me... I'll stick to normal commits for future PRs (or look if I can find that comment) 👍🏼 
Btw just in case you didn't notice the little compare button on the right shows the diffs on force pushes too which is quite helpful 
image

magiconair:
You'll learn something new every day :)

PR #675, opened 2023-08-21
feat: add support for FindServers and FindServersOnNetwork

sruehl:
This PR adds support for FindServers and FindServersOnNetwork

### DIFF
--- a/client.go
+++ b/client.go
@@ -27,6 +27,36 @@ import (
 	"github.com/gopcua/opcua/uasc"
 )
 
+// FindServers returns the servers known to a server or discovery server.
+func FindServers(ctx context.Context, endpoint string, opts ...Option) ([]*ua.ApplicationDescription, error) {
+	opts = append(opts, AutoReconnect(false))
+	c := NewClient(endpoint, opts...)
+	if err := c.Dial(ctx); err != nil {
+		return nil, err
+	}
+	c.Close(ctx)
+	res, err := c.FindServers(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return res.Servers, nil
+}
+
+// FindServersOnNetwork returns the servers known to a server or discovery server. Unlike FindServers, this service is only implemented by discovery servers.
+func FindServersOnNetwork(ctx context.Context, endpoint string, opts ...Option) ([]*ua.ServerOnNetwork, error) {
+	opts = append(opts, AutoReconnect(false))
+	c := NewClient(endpoint, opts...)
+	if err := c.Dial(ctx); err != nil {
+		return nil, err
+	}
+	c.Close(ctx)
+	res, err := c.FindServersOnNetwork(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return res.Servers, nil
+}
+
 // GetEndpoints returns the available endpoint descriptions for the server.
 func GetEndpoints(ctx context.Context, endpoint string, opts ...Option) ([]*ua.EndpointDescription, error) {
 	opts = append(opts, AutoReconnect(false))
@@ -924,6 +954,32 @@ func (c *Client) sendWithTimeout(ctx context.Context, req ua.Request, timeout ti
 // through this client connection.
 func (c *Client) Node(id *ua.NodeID) *Node {
 	return &Node{ID: id, c: c}
+}
+
+// FindServers finds the servers available at an endpoint
+func (c *Client) FindServers(ctx context.Context) (*ua.FindServersResponse, error) {
+	stats.Client().Add("FindServers", 1)
+
+	req := &ua.FindServersRequest{
+		EndpointURL: c.endpointURL,
+	}
+	var res *ua.FindServersResponse
+	err := c.Send(ctx, req, func(v interface{}) error {
+		return safeAssign(v, &res)
+	})
+	return res, err
+}
+
+// FindServersOnNetwork finds the servers available at an endpoint
+func (c *Client) FindServersOnNetwork(ctx context.Context) (*ua.FindServersOnNetworkResponse, error) {
+	stats.Client().Add("FindServersOnNetwork", 1)
+
+	req := &ua.FindServersOnNetworkRequest{}
+	var res *ua.FindServersOnNetworkResponse
+	err := c.Send(ctx, req, func(v interface{}) error {
+		return safeAssign(v, &res)
+	})
+	return res, err
 }
 
 // GetEndpoints returns the list of available endpoints of the server.
--- a/client.go
+++ b/client.go
@@ -37,7 +37,7 @@ func FindServers(ctx context.Context, endpoint string, opts ...Option) ([]*ua.Ap
 	if err := c.Dial(ctx); err != nil {
 		return nil, err
 	}
-	c.Close(ctx)
+	defer c.Close(ctx)
 	res, err := c.FindServers(ctx)
 	if err != nil {
 		return nil, err
@@ -55,7 +55,7 @@ func FindServersOnNetwork(ctx context.Context, endpoint string, opts ...Option) 
 	if err := c.Dial(ctx); err != nil {
 		return nil, err
 	}
-	c.Close(ctx)
+	defer c.Close(ctx)
 	res, err := c.FindServersOnNetwork(ctx)
 	if err != nil {
 		return nil, err
--- a/client.go
+++ b/client.go
@@ -30,7 +30,10 @@ import (
 // FindServers returns the servers known to a server or discovery server.
 func FindServers(ctx context.Context, endpoint string, opts ...Option) ([]*ua.ApplicationDescription, error) {
 	opts = append(opts, AutoReconnect(false))
-	c := NewClient(endpoint, opts...)
+	c, err := NewClient(endpoint, opts...)
+	if err != nil {
+		return nil, err
+	}
 	if err := c.Dial(ctx); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


sruehl:
A note on that: Support was there before by using `c.Send`, so in a nutshell this PR exposes this functionality top level and adjusting the README

sruehl:
if this PR is unwanted because this functionality is unwanted (and independent of that) I would suggest adjusting the README to add the highlevel/lowlevel support as columns because at first clance without reading the small disclaimer at the top it seems like gopcua doesn't have discovery services support implemented.

client.go
```diff
@@ -936,6 +974,32 @@ func (c *Client) Node(id *ua.NodeID) *Node {
...
+	var res *ua.FindServersResponse
+	err := c.Send(ctx, req, func(v interface{}) error {
+		return safeAssign(v, &res)
+	})
+	return res, err
+}
+
+// FindServersOnNetwork finds the servers available at an endpoint
+func (c *Client) FindServersOnNetwork(ctx context.Context) (*ua.FindServersOnNetworkResponse, error) {
+	stats.Client().Add("FindServers", 1)
```
magiconair: Shouldn't that be `FindServersOnNetwork` ?

client.go
```diff
@@ -27,6 +27,44 @@ import (
...
+}
+
+// FindServersOnNetwork returns the Servers known to a Server or Discovery Server. Unlike FindServers, this Service is only implemented by Discovery Servers.
+func FindServersOnNetwork(ctx context.Context, endpoint string, opts ...Option) ([]*ua.ServerOnNetwork, error) {
+	opts = append(opts, AutoReconnect(false))
+	c := NewClient(endpoint, opts...)
+	if err := c.Dial(ctx); err != nil {
+		return nil, err
+	}
+	defer func(c *Client, ctx context.Context) {
```
magiconair: Same comment as on #674 . Just call `c.Close()` and ignore the error since it cannot be recovered anyway. The caller is interested in the root cause of the close.

client.go
```diff
@@ -27,6 +27,44 @@ import (
...
 )
 
+// FindServers returns the Servers known to a Server or Discovery Server.
+func FindServers(ctx context.Context, endpoint string, opts ...Option) ([]*ua.ApplicationDescription, error) {
+	opts = append(opts, AutoReconnect(false))
+	c := NewClient(endpoint, opts...)
+	if err := c.Dial(ctx); err != nil {
+		return nil, err
+	}
+	defer func(c *Client, ctx context.Context) {
```
magiconair: `c.Close()` see #674

examples/discovery/discovery.go
```diff
@@ -0,0 +1,53 @@
...
+func main() {
+	endpoint := flag.String("endpoint", "opc.tcp://localhost:4840", "OPC UA Endpoint URL")
+	flag.BoolVar(&debug.Enable, "debug", false, "enable debug logging")
+	flag.Parse()
+	log.SetFlags(0)
+
+	ctx := context.Background()
+
+	log.Println("Finding servers on network")
+	serversOnNetwork, err := opcua.FindServersOnNetwork(ctx, *endpoint)
```
magiconair: Please no IntelliJ style auto-generated variable names. Since you are calling `FindServersOnNetwork` you'll get a list of `servers` or maybe even only endpoints back.

Also, please stick to checking `if err != nil {}` first since this is what almost all Go code does. Doing it the other way around makes reading harder since you have to use `else` instead of a `return` or `log.Fatal` and people have to actually read the code instead of scanning it.

```go
servers, err := opcua.FindServersOnNetwork(ctx, *endpoint)
if err != nil {
    log.Fatal(err)
}

for _, srv := range servers { ... }
```

client.go
```diff
@@ -27,6 +27,44 @@ import (
 	"github.com/gopcua/opcua/uasc"
 )
 
+// FindServers returns the Servers known to a Server or Discovery Server.
```
magiconair: the servers known to a server or discovery server.

magiconair (changes requested):

client.go
sruehl: yes, thanks, copy paste error :)

magiconair:
> if this PR is unwanted because this functionality is unwanted (and independent of that) I would suggest adjusting the README to add the highlevel/lowlevel support as columns because at first clance without reading the small disclaimer at the top it seems like gopcua doesn't have discovery services support implemented.

Let's add this. Thank you.

examples/discovery/discovery.go
sruehl: regardings the variable name `serversOnNetwork` was chosen on purpose as I already use the `servers` below on the `FindServers` call. So I could use `servers` for both, but then I have to put them in curly braces. The error check is inverse as in my case the `https://github.com/digitalpetri/opc-ua-demo-server` server doesn't support the `FindServersOnNetwork` call and only the `FindServers` call. So I could just invert the if but not sure if it makes it more readable then.

sruehl:
> > if this PR is unwanted because this functionality is unwanted (and independent of that) I would suggest adjusting the README to add the highlevel/lowlevel support as columns because at first clance without reading the small disclaimer at the top it seems like gopcua doesn't have discovery services support implemented.
> 
> Let's add this. Thank you.

Ok, should I add it to this PR or a new PR?

client.go
sruehl: Fixed. FYI: that came from copying that from the spec: https://reference.opcfoundation.org/Core/Part4/v104/docs/5.4.2

examples/discovery/discovery.go
sruehl: Fixed by naming vars `servers` and putting those calls in curly brackets

sruehl:
I added the README adjustments to a new PR #676

client.go
```diff
@@ -27,6 +27,36 @@ import (
 	"github.com/gopcua/opcua/uasc"
 )
 
+// FindServers returns the servers known to a server or discovery server.
+func FindServers(ctx context.Context, endpoint string, opts ...Option) ([]*ua.ApplicationDescription, error) {
+	opts = append(opts, AutoReconnect(false))
+	c := NewClient(endpoint, opts...)
```
magiconair: Can you rebase this PR on top of #674 so that we can update this line and the one in `FindServersOnNetwork` ?

examples/discovery/discovery.go
```diff
@@ -0,0 +1,57 @@
...
+
+func main() {
+	endpoint := flag.String("endpoint", "opc.tcp://localhost:4840", "OPC UA Endpoint URL")
+	flag.BoolVar(&debug.Enable, "debug", false, "enable debug logging")
+	flag.Parse()
+	log.SetFlags(0)
+
+	ctx := context.Background()
+
+	{
```
magiconair: yes, the `{}` scope trick works and there are places to do this but this isn't one of them IMHO. Just create two functions and call them from `main()`

examples/discovery/discovery.go
```diff
@@ -0,0 +1,57 @@
...
+		log.Println("Finding servers on network")
+		servers, err := opcua.FindServersOnNetwork(ctx, *endpoint)
+		if err != nil {
+			log.Printf("Error calling find servers on network: %v", err)
```
magiconair: Isn't this a `log.Fatal()`? Using `else` in Go is usually a sign that the code flow is wrong.

magiconair (changes requested):

magiconair:
merged #674 so you can rebase to `main`

client.go
sruehl: rebased and pushed. Also noticed something that slipped through. The close call should have been defers. Fixed that too :)

magiconair:
I'm making some changes to the GitHub actions. I'll ask you to rebase one more time when I'm done.

magiconair:
See #677

examples/discovery/discovery.go
sruehl: done

examples/discovery/discovery.go
sruehl: Reason I didn't it that way was because from my understanding `FindServers` should be implemented by all servers and `FindServersOnNetwork` only by some. As I do both calls in the example I don't want to Fatal here as I don't see the second call then. One way would also be to reorder the calls.

sruehl: For reference

FindServers: https://reference.opcfoundation.org/Core/Part4/v104/docs/5.4.2
FindServersOnNetwork: https://reference.opcfoundation.org/Core/Part4/v104/docs/5.4.3
Quote: "Unlike [FindServers](https://reference.opcfoundation.org/search/17?t=FindServers), this [Service](https://reference.opcfoundation.org/search/17?t=Service) is only implemented by [Discovery](https://reference.opcfoundation.org/search/17?t=Discovery) [Servers](https://reference.opcfoundation.org/search/17?t=Servers)."

sruehl:
rebased on main after #677 was merged

examples/discovery/discovery.go
```diff
@@ -0,0 +1,61 @@
...
+
+func main() {
+	endpoint := flag.String("endpoint", "opc.tcp://localhost:4840", "OPC UA Endpoint URL")
+	flag.BoolVar(&debug.Enable, "debug", false, "enable debug logging")
+	flag.Parse()
+	log.SetFlags(0)
+
+	ctx := context.Background()
+
+	findServersOnNetwork(ctx, *endpoint)
```
magiconair: I'd make the functions return an error in both cases and `log.Print(err)` in the one case and `log.Fatal(err)` in the other.

magiconair (changes requested):

magiconair (approved):

magiconair:
All done and released `v0.5.1`. Thank you @sruehl !!! 😊

PR #676, opened 2023-08-21
Readme: adjust  Services section

sruehl:
as noted in #675 this PR adjusts the README to clarify usage of services

PR #677, opened 2023-08-22
Update GitHub actions

magiconair:
* update the GitHub actions to disable the warnings
* build for push on `main` OR on PR but not on both
* take advantage of the Go module cache


sruehl:
FYI: independent of what the actions are doing I put a `[skip ci]` into my commit message to not dispatch actions if I know changes don't need a build. (https://github.com/gopcua/opcua/pull/676 for reference)

PR #680, opened 2023-08-23
refactor(id-cmd): split up id files into multiple files

### LINKED ISSUE #679: id_gen.go too large
at the moment id_gen.go is a pretty big file. Github has trouble rendering it and my IDE (Goland) just bails because the file is too big (it even can't resolve symbols so you get red errors everywhere).

image
image

Looking into it I had the idea to split the file up into 1+n files where n is the number of `NodeClass`es ([ref](https://reference.opcfoundation.org/Core/Part6/v104/docs/A.3)). So we would have on `common.go` for the `Name` function feeding from n maps and a `id_${NodeClass}_gen.go` file for nodes grouped by class.

so codewise nothing would change as the name map is private anyway at the moment, we would just split up the load between multiple files.

sruehl:
This PR spilts up id_gen.go into multiple files

resolves https://github.com/gopcua/opcua/issues/679


sruehl:
On note on this PR. The variables enum is still to big. In plc4x I solved it with this regex `^([A-Z][a-z]*)(.*),Variable$` -> `$1$2,Variable$1` https://github.com/apache/plc4x/commit/b2f3acb97b3b209b0434c2441bdb09a9eee641b6. I could extend this PR to split them even more up like I did in the plc4x commit.

magiconair:
Hmm, I could argue that VSCode and vim do not have a problem with this at all. Isn't this a limit that you can tune in Goland? Seems like an arbitrary limit. 

I could live with splitting the enums and the maps into two different files but making a random split because of a Goland limit seems pretty weird. Lets try to fix the IDE first.

magiconair:
@sruehl could you rebase your PR and then I'll merge it to v0.6.1.

sruehl:
@magiconair rebased

magiconair (approved):

PR #692, opened 2023-10-16
Fixed panic if h.MessageSize < hdrlen bytes.

SailorStu:
This fix resolves a Panic seen in Receive() where if a MessageSize in Packet was less than expected 8 bytes, then a panic would result from an out of bounds slice. Fix is to check MessageSize against hdrlen(8) bytes and if smaller, return in error. https://github.com/SailorStu/Pcaps/blob/main/gopcua/opcua-crash.pcap

### DIFF
--- a/uacp/conn.go
+++ b/uacp/conn.go
@@ -367,7 +367,10 @@ func (c *Conn) Receive() ([]byte, error) {
 	}
 
 	if h.MessageSize > c.ack.ReceiveBufSize {
-		return nil, errors.Errorf("uacp: message too large: %d > %d bytes", h.MessageSize, c.ack.ReceiveBufSize)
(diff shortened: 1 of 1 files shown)


uacp/conn.go
```diff
@@ -367,7 +367,10 @@ func (c *Conn) Receive() ([]byte, error) {
 	}
 
 	if h.MessageSize > c.ack.ReceiveBufSize {
-		return nil, errors.Errorf("uacp: message too large: %d > %d bytes", h.MessageSize, c.ack.ReceiveBufSize)
+		return nil, errors.Errorf("uacp: message too large: %d > %d bytes. MsgType=%s, ChunkType=%c", h.MessageSize, c.ack.ReceiveBufSize)
```
kung-foo: Seems like there are some missing fmt params here.

kung-foo (changes requested):

uacp/conn.go
SailorStu: Sorry about that - fixed and squashed.

kung-foo (approved):

PR #693, opened 2023-10-19
Fix unchecked type assertion in Subscription Stats

Jarrah-libremfg:
Fixes the following panic:

```
github.com/gopcua/opcua.(*Subscription).StatsWithContext(0xc0000a48c0, {0xf66c28, 0x1545c20})
            github.com/gopcua/opcua@v0.3.11/subscription.go:315 +0x207
 github.com/gopcua/opcua.(*Subscription).Stats(...)
            github.com/gopcua/opcua@v0.3.11/subscription.go:296
```

### DIFF
--- a/subscription.go
+++ b/subscription.go
@@ -276,8 +276,16 @@ func (s *Subscription) Stats(ctx context.Context) (*ua.SubscriptionDiagnosticsDa
 		return nil, errors.Errorf("empty SubscriptionDiagnostics for sub=%d", s.SubscriptionID)
 	}
 
-	for _, eo := range v.Value().([]*ua.ExtensionObject) {
-		stat := eo.Value.(*ua.SubscriptionDiagnosticsDataType)
+	eos, ok := v.Value().([]*ua.ExtensionObject)
+	if !ok {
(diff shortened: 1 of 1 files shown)


subscription.go
```diff
@@ -276,8 +276,15 @@ func (s *Subscription) Stats(ctx context.Context) (*ua.SubscriptionDiagnosticsDa
 		return nil, errors.Errorf("empty SubscriptionDiagnostics for sub=%d", s.SubscriptionID)
 	}
 
-	for _, eo := range v.Value().([]*ua.ExtensionObject) {
-		stat := eo.Value.(*ua.SubscriptionDiagnosticsDataType)
+	extensions, ok := v.Value().([]*ua.ExtensionObject)
```
magiconair: * Please use `eos` instead of `extensions`
* `return nil, errors.Errorf("invalid type for SubscriptionDiagnosticsArray. Want []*ua.ExtensionObject. subID=%d nodeID=%s type=%T", s.SubscriptionID, node.String(), v.Value())`
* Please add a blank line between the `if` block and the `for` block

magiconair (changes requested):

subscription.go
Jarrah-libremfg: Thanks for the review! Those should be fixed now.

magiconair (approved):

PR #699, opened 2023-12-06
Register encode/decode funcs instead of types

deosjr:
This change would allow users to register a function to encode or decode extensionobjects they read with a custom function, bypassing the default behaviour that comes from registering a type.
This would enable for example to decode an extensionobject into a type generated at runtime.


ua/extension_object.go
```diff
@@ -38,9 +65,9 @@ type ExtensionObject struct {
 	Value        interface{}
 }
 
-func NewExtensionObject(value interface{}) *ExtensionObject {
+func NewExtensionObject(value interface{}, typeID *ExpandedNodeID) *ExtensionObject {
```
bitomaxsp, line 56: It worth making a comment why and what typeID need to be provided i think

ua/decode.go
```diff
@@ -43,6 +43,7 @@ func decode(b []byte, val reflect.Value, name string) (n int, err error) {
 		}()
 	}
 
+	// fmt.Printf("decode: %s is a %s\n", name, val.Kind())
```
bitomaxsp, line 46: Do you really need that commented or it's a leftover?

ua/extension_object.go
```diff
@@ -5,17 +5,41 @@
...
-	if err := eotypes.Register(typeID, v); err != nil {
+	ef := func(vv interface{}) ([]byte, error) {
+		// TODO check/ensure vv is of type v ?
+		// TODO
+		return Encode(vv)
+	}
+	df := func(b []byte, vv interface{}) error {
+		rv := reflect.ValueOf(vv)
+		if rv.Kind() != reflect.Pointer || rv.IsNil() {
+			return fmt.Errorf("incorrect type to decode into")
```
bitomaxsp: I think we need to return what type is expected to avoid people going to the source code to figure that out.

ua/extension_object.go
```diff
@@ -5,17 +5,41 @@
...
+		buf.ReadStruct(r)
+		reflect.Indirect(rv).Set(reflect.ValueOf(r))
+		return nil
+	}
+	RegisterExtensionObjectFunc(typeID, ef, df)
+}
+
+// RegisterExtensionObjectFunc registers a new extension object type using encode and decode functions
+// It panics if the type or the id is already registered.
+func RegisterExtensionObjectFunc(typeID *NodeID, ef encodefunc, df decodefunc) {
```
bitomaxsp, line 25: If this method is public the `encodefunc` and `decodefunc` should also be public imo

ua/extension_object.go
```diff
@@ -28,6 +52,9 @@ const (
 	ExtensionObjectXML    = 2
 )
 
+type encodefunc func(v interface{}) ([]byte, error)
```
bitomaxsp, line 43: encodefunc and decodefunc should also be public imo

ua/extension_object.go
deosjr: > 	err := decode(body.Bytes(), &e.Value)
so currently always true; I took this from example in encoding/json golib. 
Perhaps simply removing the check altogether is better

ua/extension_object.go
```diff
@@ -74,13 +101,17 @@ func (e *ExtensionObject) Decode(b []byte) (int, error) {
 	}
 
 	typeID := e.TypeID.NodeID
-	e.Value = eotypes.New(typeID)
-	if e.Value == nil {
+	decode := eotypes.DecodeFunc(typeID)
+	if decode == nil {
 		debug.Printf("ua: unknown extension object %s", typeID)
```
bitomaxsp, line 95: I think this message is a bit misleading after refactoring. What we now register is funcs iiuc

ua/decode.go
deosjr: I moved it, it was there but commented out. I could move it back but its more useful here.

ua/extension_object.go
deosjr: No one should be using this function imo

ua/extension_object.go
deosjr: Why? Its essentially an interface that you can still adhere to from outside of the package (see read_unknow_node_id_test.go for example). I don't think anyone should even know about the type name, just the function handle

ua/typereg.go
```diff
@@ -5,92 +5,87 @@
...
-//
-// Types can be registered multiple times under different
-// identifiers.
-//
-// The implementation is safe for concurrent use.
-type TypeRegistry struct {
-	mu    sync.Mutex
-	types map[string]reflect.Type
-	ids   map[reflect.Type]string
+type FuncRegistry struct {
```
bitomaxsp, line 13: I think it deserves a comment but wrt to functions

deosjr: How do you feel about renaming the file from typereg.go -> funcreg.go ?

ua/decode.go
bitomaxsp: Nah, leave it.

deosjr:
Closing after discussion; this should not be part of this library.

Issue #673, opened 2023-08-17
StatusBadSessionNotActivated in updateNamespaces call during recreateSession action while reconnecting


law-dwg:
We are experiencing an issue with the autoreconnect functionality when trying to reestablish a connection to an OPC UA server that has been reset. In my current working example, I have a test server running out of a docker container, in which an initial connection can be established, but is unable to be reestablished when the container is restarted (docker pause seems to work fine, which makes sense since the connection is still active, just frozen)

Here is my small example:
```go
package main

import (
	"context"
	"log"
	"time"

	"github.com/gopcua/opcua"
	"github.com/gopcua/opcua/debug"
	"golang.org/x/exp/slog"
)

func main() {
	debug.Enable = true
	log.SetFlags(log.LstdFlags | log.Lshortfile)
	endpoint := "opc.tcp://localhost:4840"
	ctx := context.Background()
	c := opcua.NewClient(endpoint, opcua.AutoReconnect(true))

	if err := c.Connect(ctx); err != nil {
		log.Fatal(err)
	}
	defer c.Close(ctx)

	t := time.NewTicker(1 * time.Second)
	stop := time.After(60 * time.Second)

loop:
	for {
		select {
		case v := <-t.C:
			slog.Info("TICK", "v", v)
			slog.Info(c.State().String())

		case <-stop:
			break loop
		}
	}
	slog.Info("TICKS ARE DONE", "STATE", c.State().String())
}

```
___
Steps to reproduce are:

1.  start a docker container in the background with opcua server forwarding port 4840:

 ```shell
docker run -p 4840:4840 opcuadummyserver:v0.0.1
```

2.  run above code

 ```shell
go run main.go
```

3.  kill container
4.  run step 1 again
5.  observe ```main.go``` logs

steps 2-4 should be completed within 60 seconds otherwise the stop duration should be changed.

___

The debug logs from the package seem to indicate that its stuck in a loop delivering the following messages repeatedly:

```text
debug: client: monitor: action: createSecureChannel
debug: uacp 10858: close
debug: uasc 10858: readChunk EOF
debug: uasc 10858: Close()
debug: client: monitor: trying to recreate secure channel
debug: uacp: connecting to opc.tcp://localhost:4840
debug: uacp 10859: start HEL/ACK handshake
debug: uacp 10859: sent HELF with 56 bytes
debug: uacp 10859: recv ACKF with 28 bytes
[...]

PR #700, opened 2023-12-06
setSession to nil in recreateSession action to avoid unnecessary CloseSession 

mvandergrift:
In the client.go monitor loop, the recreateSession action builds a new session on the assumption that the server is no longer holding the previous. It does not clear the value of the Client's atomicSession, which leads to CloseSession being called in ActivateSession even if there are no old sessions on the server to be removed.

This behavior is referenced in issue #673.

### DIFF
--- a/client.go
+++ b/client.go
@@ -447,6 +447,10 @@ func (c *Client) monitor(ctx context.Context) {
 						c.setState(Reconnecting)
 						// create a new session to replace the previous one
 
+						// clear any previous session as we know the server has closed it
+						// this also prevents any unnecessary calls to CloseSession
+						c.setSession(nil)
+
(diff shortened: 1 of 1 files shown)


magiconair (approved):
I think this is a good catch and I agree. I hope this fixes the issue in #673

magiconair:
I'm going to cut a new release once this is merged. @law-dwg could you maybe test if this fixes your problem as well?

magiconair:
Released v0.5.3

law-dwg:
Works from my side, thanks!

magiconair:
Awesome!!! Thank you very much both @mvandergrift and @law-dwg

PR #702, opened 2023-12-15
RegisterExtensionObjectFromServer WIP

deosjr:


ua/schema.go
```diff
@@ -0,0 +1,137 @@
...
+
+import (
+	"fmt"
+	"reflect"
+	"time"
+
+	"github.com/pascaldekloe/name"
+)
+
+// Serializable checks if the Go struct t can be serialized into the UA type name.
```
bitomaxsp: remove `t` before `can`

ua/schema.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// we don't support structs with optional fields or unions (yet or maybe never)
+	if t.StructureType != StructureTypeStructure {
+		return nil, fmt.Errorf("schema: ua type %q has unsupported structure type %s", id, t.StructureType)
+	}
+
+	var fields []reflect.StructField
+	for _, f := range t.Fields {
+		if len(f.ArrayDimensions) > 1 {
+			return nil, fmt.Errorf("schema: ua type %s.%s is a multi-dimensional array: %v", id, f.Name, f.ArrayDimensions)
```
bitomaxsp, line 75: This error is unclear. It just states the fact without saying what is wrong imo  :)
I am not sure i understand why is it an error. This is a valid use case for PackML for example.

magiconair:
It looks like FreeOpcUa does not support the `DataTypeDefinition` attribute on the server on which this whole feature depends on... Need to find a different way for testing.

PR #705, opened 2024-01-17
Description is not necessary and handling BadNodeIdUnknown node in browse

womeikong:
browse.go in examples

1. Handle the exception of nil to string when description is nil
2. Browse next node when current nodeid is BadNodeIdUnknown

### DIFF
--- a/examples/browse/browse.go
+++ b/examples/browse/browse.go
@@ -11,6 +11,7 @@ import (
 	"log"
 	"os"
 	"strconv"
+	"strings"
 
 	"github.com/gopcua/opcua"
 	"github.com/gopcua/opcua/debug"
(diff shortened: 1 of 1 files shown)


examples/browse/browse.go
```diff
@@ -146,7 +150,20 @@ func browse(ctx context.Context, n *opcua.Node, path string, level int) ([]NodeD
 		for _, rn := range refs {
```
magiconair, line 150: I'd replace the nested `if` statements with this:

```go
		for _, rn := range refs {
			children, err := browse(ctx, rn, def.Path, level+1)
			switch {
			case err == ua.StatusBadNodeIDUnknown:
				browseName := strings.Split(rn.String(), def.BrowseName+".")[1]
				nodes = append(nodes, NodeDef{
					BrowseName:  browseName,
					Path:        join(def.Path, browseName),
					NodeID:      ua.NewStringNodeID(rn.ID.Namespace(), "BadNodeIdUnknown (0x80340000)"),
					Description: "BadNodeIdUnknown (0x80340000)",
					Writable:    false,
				})

			case err != nil:
				return errors.Errorf("browse children: %s", err)
			}
			nodes = append(nodes, children...)
		}
```

magiconair (changes requested):

PR #709, opened 2024-01-31
fix: ignore response with too many operations error and fix extension object decoding

huskar-t:
`StatusBadTooManyOperations` error does not need to close the connection

according to https://reference.opcfoundation.org/Core/Part6/v105/docs/5.2.2
`Body` `This field contains the raw bytes for [ByteString] bodies.`
So there is no need to parse length again when parsing XMLElement.Theoretically speaking, the encode method is also wrong.

### DIFF
--- a/client.go
+++ b/client.go
@@ -306,6 +306,10 @@ func (c *Client) monitor(ctx context.Context) {
 				continue
 			}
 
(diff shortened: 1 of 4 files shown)


huskar-t:
@magiconair could you please review my pull request?

PR #712, opened 2024-02-06
subscription: add SetMonitoringMode functionality

### LINKED ISSUE #711: subscription: add SetMonitoringMode functionality
Add `SetMonitoringMode` functionality to complete the `MonitoredItems Service Set`

jackchenjc:
closes #711

Add SetMonitoringMode functionality and fix the potential issue that leaves the mutex locked.

Simply add the following code to `examples/subscribe/subscribe.go` can test this new feature
```
resp, err := sub.SetMonitoringMode(ctx, ua.MonitoringModeSampling, res.Results[0].MonitoredItemID)
if err != nil || resp.Results[0] != ua.StatusOK {
	log.Fatal(err)
}
```

### DIFF
--- a/subscription.go
+++ b/subscription.go
@@ -177,14 +177,20 @@ func (s *Subscription) ModifyMonitoredItems(ctx context.Context, ts ua.Timestamp
 	stats.Subscription().Add("ModifyMonitoredItems", 1)
 	stats.Subscription().Add("ModifiedMonitoredItems", int64(len(items)))
 
+	var err error
 	s.itemsMu.Lock()
 	for _, item := range items {
 		id := item.MonitoredItemID
 		if _, exists := s.items[id]; !exists {
-			return nil, fmt.Errorf("sub %d: cannot modify unknown monitored item id: %d", s.SubscriptionID, id)
+			err = fmt.Errorf("sub %d: cannot modify unknown monitored item id: %d", s.SubscriptionID, id)
+			break
 		}
 	}
 	s.itemsMu.Unlock()
+
+	if err != nil {
+		return nil, err
+	}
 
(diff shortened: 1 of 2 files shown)


jackchenjc:
Not sure if we should update `item.req.MonitoringMode` while getting `ua.StatusOK`.

subscription.go
```diff
@@ -220,6 +220,34 @@ func (s *Subscription) ModifyMonitoredItems(ctx context.Context, ts ua.Timestamp
...
 }
 
+func (s *Subscription) SetMonitoringMode(ctx context.Context, monitoringMode ua.MonitoringMode, monitoredItemIDs ...uint32) (*ua.SetMonitoringModeResponse, error) {
+	stats.Subscription().Add("SetMonitoringMode", 1)
+	stats.Subscription().Add("SetMonitoringModeMonitoredItems", int64(len(monitoredItemIDs)))
+
+	s.itemsMu.Lock()
+	for _, id := range monitoredItemIDs {
+		if _, exists := s.items[id]; !exists {
+			return nil, fmt.Errorf("sub %d: cannot set monitoring mode for unknown monitored item id: %d", s.SubscriptionID, id)
```
kung-foo: returning here leaves the mutex locked

kung-foo (changes requested):

subscription.go
jackchenjc: Done and also fixed the one in `ModifyMonitoredItems`.

kung-foo (changes requested):
Ah, https://github.com/gopcua/opcua/blob/v0.5.3/README.md?plain=1#L206 needs an update as well. Otherwise LGTM

kung-foo (approved):

PR #714, opened 2024-02-19
subscription: add ModifySubscription functionality

### LINKED ISSUE #713: subscription: add ModifySubscription functionality
Add `ModifySubscription` functionality to fulfill the `Subscription Service Set`.

jackchenjc:
closes #713

### DIFF
--- a/subscription.go
+++ b/subscription.go
@@ -22,8 +22,6 @@ const (
 	DefaultSubscriptionPriority                   = 0
 )
 
-const terminatedSubscriptionID uint32 = 0xC0CAC01B
-
 type Subscription struct {
 	SubscriptionID            uint32
 	RevisedPublishingInterval time.Duration
--- a/subscription.go
+++ b/subscription.go
@@ -31,6 +31,7 @@ type Subscription struct {
 	RevisedMaxKeepAliveCount  uint32
 	Notifs                    chan<- *PublishNotificationData
 	params                    *SubscriptionParameters
+	paramsMu                  sync.Mutex
 	items                     map[uint32]*monitoredItem
 	itemsMu                   sync.Mutex
 	lastSeq                   uint32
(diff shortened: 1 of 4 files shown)


subscription.go
```diff
@@ -111,6 +111,38 @@ func (s *Subscription) delete(ctx context.Context) error {
...
+	err := s.c.Send(ctx, req, func(v interface{}) error {
+		return safeAssign(v, &res)
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	// update subscription parameters
+	s.params = &params
```
kung-foo, line 137: These might need a lock. But it doesn't look like we have one in Subscription. There's this func (with a truly awful name): https://github.com/gopcua/opcua/blob/5aed8b4915ff215537c42fddf8d12d8c5bb3ba10/subscription.go#L353

jackchenjc: Do you suggest to create a new **Subscription** based on the current subscription and then invoke its `recreate_NeedsSubMuxLock` func to safely update these value?

Someting like:

```suggestion
	s.itemsMu.Lock()
	sub := &Subscription{
		SubscriptionID:            s.SubscriptionID,
		RevisedPublishingInterval: time.Duration(res.RevisedPublishingInterval) * time.Millisecond,
		RevisedLifetimeCount:      res.RevisedLifetimeCount,
		RevisedMaxKeepAliveCount:  res.RevisedMaxKeepAliveCount,
		Notifs:                    s.Notifs,
		items:                     s.items,
		params:                    &params,
		nextSeq:                   s.nextSeq,
		c:                         s.c,
	}
	s.itemsMu.Unlock()

	err = sub.recreate_NeedsSubMuxLock(ctx)
	if err != nil {
		return nil, err
	}
```

jackchenjc:
Hi @kung-foo  @magiconair,
This is ready for review, thanks.

subscription.go
magiconair: Yeah, that needs to be refactored. Back then I was not able to see how to solve the tight coupling but I know how to do it now.

magiconair: @jackchenjc accessing `s.params` also needs to use the `paramsMu` lock in the `recreate_NeedSubMuxLock` function. I'll add it when I refactor this.

jackchenjc: @magiconair Good to hear that and thanks for the review.
Shoud I add some `TODO` comments or is there anything else I can do in this PR?

magiconair: @jackchenjc no, thank you. I'll take it from here.

magiconair: I've refactored the control flow so that it only goes from `client_sub` to `subscription` and not back and forth. As for the awful name @kung-foo I'm open for suggestions on how to reflect that an internal method should only be called when a certain lock is held. The reason I did this was because I ran into this issue. Is there a best-practice?

magiconair:
Rebased to `main`

magiconair:
Hi @kung-foo it'd be great if you could review this one.

PR #746, opened 2024-10-16
 Implement Event Subscription handling for channels

JonathanWhiteUK:
Hi, I have added event channel handling. This is a breaking change as it changes the signatures of some of the functions to allow for handling of both event messages and data change messages. Thanks :)


SCZwilling:
Hello, I was trying to use your implementation of event subscription but i am getting error- "The attribute is not supported for the specified Node. StatusBadAttributeIDInvalid (0x80350000)". I am subscribing to an object node which has below configuration-

```
Attribute | Value
Nodeld | ins=2;s=Sinumerik
NamespaceIndex 2 |  
IdentifierType | String
Identifier | Sinumerik
NodeClass | Object
BrowseName | 2, "Sinumerik"
DisplayName | "en-US", "Sinumerik"
Description | "en-US", "An object with several features of Sinumerik"
EventNotifier | SubscribeToEvents
```


I have tested this node on UA Expert and there i am able to see the events and alarms. I believe i need to change some configuration to subscribe events on this node, could you help with it @Jwcode-uk ?

magiconair:
Thank you @Jwcode-uk. I can see that one function in `subscription.go` gets an additional parameter. I think we could use a different pattern for functions like this. Not sure if we have more than just this one. Might be worthwhile to look. Would like to keep this consistent.

Not sure if this is complex enough to use the `Option` pattern for this constructor. What do you think?

```go
// Deprectated: Use NewDefaultMonitoredItemCreateRequest instead. Will be removed with 0.8.0
func NewMonitoredItemCreateRequestWithDefaults(nodeID *ua.NodeID, attributeID ua.AttributeID, clientHandle uint32) *ua.MonitoredItemCreateRequest {
	return NewDefaultMonitoredItemCreateRequest(MICreateRequestArgs{
		NodeID:       nodeID,
		AttributeID:  attributeID,
		ClientHandle: clientHandle,
	})
}

type MonitoredItemCreateRequestArgs struct {
	NodeID       *ua.NodeID
	AttributeID  ua.AttributeID
	ClientHandle uint32
	Filter       *ua.ExtensionObject
}

func NewDefaultMonitoredItemCreateRequest(args MonitoredItemCreateRequestArgs) *ua.MonitoredItemCreateRequest {
	if args.AttributeID == 0 {
		args.AttributeID = ua.AttributeIDValue
	}
	return &ua.MonitoredItemCreateRequest{
		ItemToMonitor: &ua.ReadValueID{
			NodeID:       args.NodeID,
			AttributeID:  args.AttributeID,
			DataEncoding: &ua.QualifiedName{},
		},
		MonitoringMode: ua.MonitoringModeReporting,
		RequestedParameters: &ua.MonitoringParameters{
			ClientHandle:     args.ClientHandle,
			DiscardOldest:    true,
			Filter:           args.Filter,
			QueueSize:        10,
			SamplingInterval: 0.0,
		},
	}
}
```

monitor/subscription.go
```diff
@@ -216,6 +229,49 @@ func (s *Subscription) pump(ctx context.Context, notifyCh chan<- *DataChangeMess
...
+					} else {
+						// Initialize the EventFields slice with the correct size
+						out.EventFields = make([]*ua.DataValue, len(item.EventFields))
+
+						for i, field := range item.EventFields {
+							// Create a new DataValue
+							dataValue := &ua.DataValue{
+								Value:           field,
+								Status:          ua.StatusOK,
+								SourceTimestamp: time.Now(),
```
magiconair: Shouldn't all `dataValue`s have the same timestamp? Isn't there a timestamp in the `item` (e.g. `ServerTimestamp`)?

monitor/subscription.go
```diff
@@ -155,7 +168,7 @@ func (s *Subscription) sendError(err error) {
 }
 
 // internal func to read from internal channel and write to client provided channel
-func (s *Subscription) pump(ctx context.Context, notifyCh chan<- *DataChangeMessage, cb MsgHandler) {
+func (s *Subscription) pump(ctx context.Context, notifyCh chan<- Message, cb MsgHandler, sub bool) {
```
magiconair: `sub` seems unused here and the name also collides with other usages of `sub` in this file where it usually denotes a `*Subscription`

monitor/subscription.go
```diff
@@ -254,16 +310,16 @@ func (s *Subscription) Dropped() uint64 {
 }
 
 // AddNodes adds nodes defined by their string representation
-func (s *Subscription) AddNodes(ctx context.Context, nodes ...string) error {
+func (s *Subscription) AddNodes(ctx context.Context, eventSub bool, filter *ua.ExtensionObject, nodes ...string) error {
```
magiconair: What does `eventSub` do here? Also, see comment on `ChanSubscribe` for a way to add this functionality without breaking the signatures.

monitor/subscription.go
```diff
@@ -281,6 +342,67 @@ func (s *Subscription) AddMonitorItems(ctx context.Context, nodes ...Request) ([
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
+	if len(nodes) == 0 {
+		// some server implementations allow an empty monitoreditemrequest, some don't.
+		// better to just return
+		return nil, nil
+	}
+
+	toAdd := make([]*ua.MonitoredItemCreateRequest, 0)
```
magiconair, line 481:
```go
var toAdd []*ua.MonitoredItemCreateRequest
```

monitor/subscription.go
```diff
@@ -216,6 +229,49 @@ func (s *Subscription) pump(ctx context.Context, notifyCh chan<- *DataChangeMess
...
+					out := &EventMessage{}
+					if !ok {
+						out.Error = errors.Errorf("handle %d not found", item.ClientHandle)
+						// TODO: should the error also propagate via the monitor callback?
+					} else {
+						// Initialize the EventFields slice with the correct size
+						out.EventFields = make([]*ua.DataValue, len(item.EventFields))
+
+						for i, field := range item.EventFields {
+							// Create a new DataValue
```
magiconair: pls remove this comment as it does not have value

monitor/subscription.go
```diff
@@ -216,6 +229,49 @@ func (s *Subscription) pump(ctx context.Context, notifyCh chan<- *DataChangeMess
...
+								Value:           field,
+								Status:          ua.StatusOK,
+								SourceTimestamp: time.Now(),
+								ServerTimestamp: time.Now(),
+							}
+							out.EventFields[i] = dataValue
+						}
+					}
+
+					if notifyCh != nil {
```
magiconair: I prefer `switch` statements over `if/else`, e.g.

```go
switch {
    case notifyCh != nil:
        select { ... }

    case cb != nil:
        cb(sb, out)
        atomic.AddUint64(&s.delivered, 1)

    default:
       panic(...)
}
```

subscription.go
```diff
@@ -45,27 +45,43 @@ type SubscriptionParameters struct {
...
-			NodeID:       nodeID,
-			AttributeID:  attributeID,
+			NodeID:       args.NodeID,
+			AttributeID:  args.AttributeID,
 			DataEncoding: &ua.QualifiedName{},
 		},
 		MonitoringMode: ua.MonitoringModeReporting,
 		RequestedParameters: &ua.MonitoringParameters{
-			ClientHandle:     clientHandle,
+			ClientHandle:     args.ClientHandle,
```
magiconair, line 107: Shouldn't the other parameters be configurable as well while providing sensible defaults?

monitor/subscription.go
```diff
@@ -137,13 +150,13 @@ func (m *NodeMonitor) Subscribe(ctx context.Context, params *opcua.SubscriptionP
 // via the monitor's `ErrHandler`.
 // The caller must call `Unsubscribe` to stop and clean up resources. Canceling the context
 // will also cause the subscription to stop, but `Unsubscribe` must still be called.
-func (m *NodeMonitor) ChanSubscribe(ctx context.Context, params *opcua.SubscriptionParameters, ch chan<- *DataChangeMessage, nodes ...string) (*Subscription, error) {
-	sub, err := newSubscription(ctx, m, params, 16, nodes...)
+func (m *NodeMonitor) ChanSubscribe(ctx context.Context, params *opcua.SubscriptionParameters, ch chan<- Message, eventSub bool, filter *ua.ExtensionObject, nodes ...string) (*Subscription, error) {
```
magiconair: I am starting to think that an `args` parameter might be useful here. We did something similar with the `WithContext` versions of all functions where we dropped the `WithContext` suffix in a later version. So maybe we can refactor this change so that we add `ChanSubscribeWithArgs(ctx context.Context, args ChanSubscribeArgs) (*Subscription, error)` and drop the `WithArgs` in 0.8.0. Add a deprecation note to the other version. What do you think?

magiconair (changes requested):
Did a first pass through the code and would like to see a change so that this implementation does not break existing code. See the comment on `ChanSubscribe` for a suggestion. I've also left some smaller style comments. 

Some more comments would be useful since it isn't immediately obvious to me what the code does. 

Lets do the refactor first - unless that isn't possible - and then I'll do another pass over the implementation and probably have a couple more questions.

FelixTing:
@magiconair Thank you for your review. I've synced the branch with the main branch and updated the code based on your suggestions. Could you please take another look when you have a moment? Thanks!

cloudxxx8:
this PR is ready for review again

cloudxxx8:
may we review this PR for the next release?

cloudxxx8:
@magiconair may we put this PR for the next release?

cloudxxx8:
@magiconair may we put this PR to the next release target?  we can rebase and resolve the conflict.

PR #753, opened 2024-12-04
Added Error return to SelectEndpoint function

NovaFell:
Updated the examples as well to call the function correctly.


magiconair:
This is a breaking change but something we could justify for v0.6.1

magiconair (approved):

PR #755, opened 2024-12-09
Issue #678: Fix Variant to handle nil slices

### LINKED ISSUE #678: StatusBadEncodingLimitsExceeded on single read request
After connecting to an OPC UA Server by "Softing dataFEED" I did a simple ReadRequest which resulted in a StatusBadEncodingLimitsExceeded.

The request was done like this:

```go
        // ... create client and connect
	_, err = client.ReadWithContext(ctx, &ua.ReadRequest{
		NodesToRead: []*ua.ReadValueID{{
			NodeID:      ua.MustParseNodeID("ns=0;i=2269"),
			AttributeID: ua.AttributeIDValue,
		}},
	})
	if err != nil {
		return fmt.Errorf("read with context: %w", err)
	}
```

Debugging showed that the error was created at here: https://github.com/gopcua/opcua/blob/807c2dabf202632c5fcb5c6ec00f182f3b97063c/ua/variant.go#L117-L146

The []byte that was decoded looked like this:

```
[]uint8 len: 17, cap: 65478, [140,255,255,255,255,96,227,5,236,109,213,217,1,255,255,255,255]
```

which resulted in an `m.arrayLength` of `-1`

### Looking at the Node-ID in UaExpert

image

I'm unsure how I could work around this error or fix this issue. Another problem is that the `client.State()` becomes `Disconnected` and `Reconnecting` after this issue without returning back to `Connected`.

magiconair:
Fix the Variant encoder and decoder to handle nil slices
correctly by setting the length to -1.

https://reference.opcfoundation.org/Core/Part6/v104/docs/5.2.5

Closes #678


magiconair:
This fix isn't entirely correct and the tests are too complex. Working on it.

magiconair:
@kung-foo could you have a look? Then I can merge this as well and release v0.6.1

Issue #757, opened 2024-12-09
Replace pascaldekloe/verify with stretchr/testify


magiconair:
While the `verify.Values` library has served us well it shows its age in that it doesn't show the private fields that are necessary for debugging. 

I'm going to create a PR to replace all calls of `verify.Values` with `require.Equal`. The call arguments have to be reversed since `verify.Values(t, "", got, want)` maps to `require.Equal(t, want, got)`

magiconair:
merged

PR #758, opened 2024-12-09
Issue #757: use stretchr/testify/require

magiconair:
stretchr/testify is more strict in equality testing and provides more useful debugging output. 

In a first step I've replaced all usages of `verify.Values` with `testify` and found two glitches in the variant tests.

In later steps I'll go through the `if got, want` pattern and gradually replace them with `require` as well.


kung-foo (approved):

kung-foo:
You've put this in the v0.6.2 milestone, but it looks like there are changes to the other PRs that are targeting v0.6.1 that also have testify updates. Do we need to have all of them target one milestone?

magiconair:
I changed the `CodecTest` to stretchr/testify in a previous commit since it allowed me to find a failing test with the `nil` slices. We already have two different test patterns with `got, want` and `verify.Values`. I'm trying to switch this all to `require` so that there is only one.

We don't need to have them all in one milestone and do them one step at a time as long as they're all converted at some point.

PR #759, opened 2024-12-13
fix: regression in examples introduced by #753

jessekv:
It looks like there was a mistake when updating the examples. (`err == nil` should be `err != nil`)

The broken commit: https://github.com/gopcua/opcua/pull/753/commits/f3e85a2928361e16b0b577a1255a45e628986177


jessekv:
@magiconair should I make an issue for this one as well?

magiconair:
No, that's fine. Thank you!

magiconair (approved):

Issue #616, opened 2022-11-10
log.Fatal called when a certificate fails to load


jrtitus:
This causes any process using gopcua/opcua to exit without warning and is caused by either the referenced file not existing, or a decoding failure. 

Can it be handled more cleanly like logging an error and returning? Or perhaps returning `(Option, error)` from `CertificateFile`?

https://github.com/gopcua/opcua/blob/3618f0ec0fcc2f33ad91303374404bb6daa8e44a/config.go#L268-L271

It looks like the same thing happens in the `RemoteCertificateFile`, `PrivateKeyFile`, and `setCertificate` functions as well. I also see TODO comments in `AuthAnonymous` and `AuthUsername` questioning the use of `Fatal`. My suggestion would be no.

magiconair:
Eh, `log.Fatal` in a library is never OK. Sorry about that. Will fix.

```sh
$ ack -l log.Fatal | grep -v examples
cmd/status/main.go
cmd/id/main.go
cmd/service/main.go
config.go
```

magiconair:
Ah, I remember why I did this 3y ago. No good option to bubble up errors for config options. I'll rework the config option.

jrtitus:
> Ah, I remember why I did this 3y ago. No good option to bubble up errors for config options. I'll rework the config option.

Awesome. Thank you very much for following up so quickly on this.

magiconair:
So we have two options to solve this:

1. We bubble the error up during `Dial` and/or `Connect` which does not require us to change any API
2. We change `NewClient` to return an error which breaks all existing code. Cleaner but more disruptive.
3. We do 1. until v0.5.0 where we have to break the API anyway and then update `NewClient`.

I'm for option 3. @kung-foo any opinion?

kung-foo:
3 is fine by me

jrtitus:
Hi @magiconair. Just curious - when do you plan to make the `v0.3.8` release?

magiconair:
sorry, dropped the ball here.

magiconair:
@jrtitus done.

jrtitus:
> sorry, dropped the ball here.

No worries!

> done

Thanks! :)

PR #763, opened 2025-01-10
Issue #762: Remove calls to log.Fatal

### LINKED ISSUE #762: Very dangerous log.Fatal called in several places
(using v6.0.2)
We are about to use this in production and we noticed that when passing an invalid certificate it panics. This can cause unexpected failures at runtimes, and i saw that https://github.com/gopcua/opcua/issues/616 was marked as closed but it does not seem to be.

There are three `log.Fatal` in the codebase:
- `config.go` line 325: "Failed to parse certificate"
- `server.go` line 112 and line 151

I wonder why the error cannot bubble up to the `NewClient` because that now returns an error. 

Are we missing something? Thanks in advance

danomagnum:
The function that used log.Fatal in config.go isn't exported and is only called by functions that return errors so I added an error return then pass it up.

The first log.Fatal in server.go followed the same approach as #616 to delay the error till the start() is called on the server.

The second log.Fatal in server.go should never be possible, so I changed it to a panic (which should never happen).

Fixes #762

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -109,9 +109,7 @@ func New(opts ...Option) *Server {
 		opt(cfg)
 	}
 	url := ""
-	if len(cfg.endpoints) == 0 {
-		log.Printf("No endpoints defined!")
-	} else {
+	if len(cfg.endpoints) != 0 {
 		url = cfg.endpoints[0]
 	}
 
--- a/server/server.go
+++ b/server/server.go
@@ -108,11 +108,15 @@ func New(opts ...Option) *Server {
 	for _, opt := range opts {
 		opt(cfg)
 	}
+	url := ""
 	if len(cfg.endpoints) == 0 {
-		log.Fatalf("No endpoints defined!")
+		log.Printf("No endpoints defined!")
+	} else {
+		url = cfg.endpoints[0]
 	}
+
 	s := &Server{
-		url:      cfg.endpoints[0],
+		url:      url,
 		cfg:      cfg,
 		cb:       newChannelBroker(cfg.logger),
 		sb:       newSessionBroker(cfg.logger),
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -108,11 +108,15 @@ func New(opts ...Option) *Server {
 	for _, opt := range opts {
 		opt(cfg)
 	}
+	url := ""
 	if len(cfg.endpoints) == 0 {
-		log.Fatalf("No endpoints defined!")
+		log.Printf("No endpoints defined!")
```
magiconair: I'd remove the log statement here. Either we validate in the `New` function and return an error or we check later. If we don't want to change the signature of `New` then we check later and you have a check for that.

magiconair (changes requested):
Other than that LGTM

magiconair:
LGTM

magiconair (approved):

danomagnum:
@magiconair 

Did you see what happened with the checks on this?  It failed the first time (couldn't open port 4840 in one of the conn_test tests).  It does pass on my machine, so I pushed an empty commit and when it ran again, it passed.  Some kind of race condition must exist in terms of one test closing the port fast enough before another can open it?  Pretty strange.

On the python integration tests I've started getting a different error sometimes too.  I'm sure it's something with how my network is setup, but I'm not 100% sure why it doesn't try to go out the gateway anyway.

```
=== RUN   TestClientTimeoutViaOptions
Endpoints other than open requested but private key and certificate are not set.
Listening on 0.0.0.0:4840
    timeout_test.go:55: 
                Error Trace:    /home/dan/programs/opcua/tests/python/timeout_test.go:55
                                                        /home/dan/programs/opcua/tests/python/timeout_test.go:28
                Error:          got %#v, wanted net.timeoutError
                Test:           TestClientTimeoutViaOptions
                Messages:       connect: network is unreachable
--- FAIL: TestClientTimeoutViaOptions (1.45s)
```

magiconair:
I've seen this and it happens sporadically but not often enough for me to be concerned. Would be great to understand why though. 

Released as v0.6.3

magiconair:
@danomagnum I wonder where all the thumbs-up on the issue came from. People from your company? If this was such an issue why didn't anyone raise this sooner.

danomagnum:
@magiconair I don't know.  Not sure who any of the people are.  I saw the issue about log.fatal come in and it involved the server so I went ahead and fixed it.  That issue got a bunch of attention really quickly as well.  Maybe all those people are at the same company?

PR #764, opened 2025-01-18
monitor: add modification functionality

### LINKED ISSUE #710: monitor: add support to modify subscription, monitored items, and set monitoring mode
Currently, we have `subscription.Subscribe()`, `subscription.Unsubscribe()`, `subscription.AddMonitorItems()` and `subscription.RemoveMonitorItems()` implemented in the `monitor` package, but the modification functionality is missing.

Should be able to leverage the functions implemented in `opcua.subscription`

jackchenjc:
closes #710

### DIFF
--- a/monitor/subscription.go
+++ b/monitor/subscription.go
@@ -56,7 +56,7 @@ func (m *Item) NodeID() *ua.NodeID {
 	return m.nodeID
 }
 
-// Request is a struct to manage a request to monitor a node
+// Request is a struct to manage a request to monitor a node or modify a monitored node
 type Request struct {
 	NodeID               *ua.NodeID
 	MonitoringMode       ua.MonitoringMode
@@ -221,6 +221,12 @@ func (s *Subscription) pump(ctx context.Context, notifyCh chan<- *DataChangeMess
 			}
 		}
 	}
+}
+
+// Modify modifies the subscription settings
+func (s *Subscription) Modify(ctx context.Context, params *opcua.SubscriptionParameters) error {
+	_, err := s.sub.ModifySubscription(ctx, *params)
+	return err
 }
 
 // Unsubscribe removes the subscription interests and cleans up any resources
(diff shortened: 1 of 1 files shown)


monitor/subscription.go
```diff
@@ -403,6 +409,121 @@ func (s *Subscription) RemoveMonitorItems(ctx context.Context, items ...Item) er
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if len(nodes) == 0 {
+		return nil
+	}
+
+	var toModify Item
+	toModifyReq := make([]*ua.MonitoredItemModifyRequest, 0)
+	for _, node := range nodes {
```
magiconair, line 423: This doesn't look right since something is appended when there was no match or am I missing something here? Also what happens when `node.MonitoringParameters` is `nil`? That also looks off.

Some minor style comments:
* You don't have to allocate an empty array in Go
* It's a common pattern to use early returns or continues as pre-conditions
* Prefer `req` over `request` since you're also using `resp` instead of `response`

How about this?

```go
	var modifiedItems[]*ua.MonitoredItemModifyRequest
	for _, node := range nodes {
		for _, item := range s.itemLookup {
			if item.nodeID.String() != node.NodeID.String() {
				continue
			}

			// todo(fs): what happens if node.MonitoringParameters is nil? 
			// todo(fs): Should the request still be added?
			// todo(fs): If not, it could look like this
			//
			// if node.MonitoringParameters == nil {
			//     break
			// }
			// req := &ua.MonitoredItemModifyRequest{
			//	MonitoredItemID: item.id,
			//	RequestedParameters: node.MonitoringParameters,
			// }
			// req.RequestedParameters.ClientHandle = item.handle
			// modifiedItems = append(modifiedItems, req)
			// break
			

			req := &ua.MonitoredItemModifyRequest{
				MonitoredItemID: item.id,
			}
			if node.MonitoringParameters != nil {
				req.RequestedParameters = node.MonitoringParameters
				req.RequestedParameters.ClientHandle = item.handle
			}
			modifiedItems = append(modifiedItems, req)
			break
		}
	}
```

magiconair (changes requested):

monitor/subscription.go
jackchenjc: You're right, using `continue` as pre-condition looks much better.
As for other styles, they are actually copied from the existing code (AddMonitorItems). Should I fix them as well in this PR?

I also agree that the request without `node.MonitoringParameters` should not be added. Thanks for the suggestion.

magiconair: > As for other styles, they are actually copied from the existing code (AddMonitorItems). Should I fix them as well in this PR?

You're right. I didn't check the rest of the code. Consistency is more important. Keep them as you had it and just fix the pre-condition and the skip of the empty params. Thank you.

magiconair (approved):

magiconair:
@jackchenjc Thank you! I've merged it and released v0.6.5

PR #766, opened 2025-02-03
server: use DataValue instead of Variant for node attributes

danomagnum:
By keeping the attribute values as DataValues instead of Variants it allows a user to set the timestamp, quality, or any other parameter of each attribute specifically instead of always returning a default.

The `SetAttribute` function already took a datavalue, so I don't expect there to be any real changes to user code - but now it will keep the value and not throw it away.  The big difference is users will have to be careful to set the correct flags on the datavalue where before it ignored these and set the values at the end.


magiconair:
Is this a breaking change? If this is a breaking change then we should tag this as v0.7 or something similar.

danomagnum:
I think the only thing that breaks is custom Node definitions - the attribute map went from `map[ua.AttributeID]*ua.Variant` to `map[ua.AttributeID]*ua.DataValue`  so if anyone was hand-rolling nodes instead of using helper functions that would break. it probably makes sense to mark it for the next revision.

magiconair (commented):
Shouldn't there be some tests that need adapting?

danomagnum:
All the tests use the helper functions so they're largely one step removed from the change.  A couple lines in the test server did have to change since `SetAttribute()` now takes a `*DataValue` instead of a `DataValue`

There's certainly an argument to be made that more tests are needed, but of the tests that exist there wasn't much to change.

magiconair:
Right. Was looking for `_test.go` files. Missed the `tests/go/server.go`

magiconair (approved):
This only affects the server so lets do it. I've moved some milestones around. 

LGTM

PR #767, opened 2025-02-09
client: add connState channel

### LINKED ISSUE #741: Does the Client/Connection send out 'events' or notify in case the status of the connection changes?
Does the Client/Connection send out 'events' in case the status of the connection changes?

The client keep track of the connection state:
```
const (
	// Closed, the Connection is currently closed
	Closed [ConnState](https://pkg.go.dev/github.com/gopcua/opcua#ConnState) = [iota](https://pkg.go.dev/builtin#iota)
	// Connected, the Connection is currently connected
	Connected
	// Connecting, the Connection is currently connecting to a server for the first time
	Connecting
	// Disconnected, the Connection is currently disconnected
	Disconnected
	// Reconnecting, the Connection is currently attempting to reconnect to a server it was previously connected to
	Reconnecting
)
```

Is there a way to get notified when the value changes?

magiconair:
Add support for an optional channel to receive client state changes.

Fixes #741

### DIFF
--- a/client.go
+++ b/client.go
@@ -166,6 +166,9 @@ type Client struct {
 	// atomicState of the client
 	atomicState atomic.Value // ConnState
 
+	// stateCh is an optional channel for connection state changes. May be nil.
+	stateCh chan<- ConnState
+
 	// list of cached atomicNamespaces on the server
 	atomicNamespaces atomic.Value // []string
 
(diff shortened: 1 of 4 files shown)


sruehl:
I wonder what the decision was to use a channel for that if it is called in that synced fashion. So I just want to use that to hook a log statement too but this puts me in a weird spot on when to know when to end the go func to consume that channel.
So IMHO the code should be extended as follows:
```
	if c.stateCh != nil {
		select {
		case <-ctx.Done():
		case c.stateCh <- s:
                default:
		}
	}
```
with the default branch we resolve the blocking issue by not blocking state changes. Of course the same could be done with a buffered channel but then the question is what is a proper size.
Feels like a:
```
func StateChanged(fn func(ConnState)) Option {
	return func(cfg *Config) error {
		cfg.stateFn = fn
		return nil
	}
}
```
would be simpler to use as we don't have the complexity managing the channel and go func when the current implementation requires it to be non-blocking anway.

PR #771, opened 2025-02-12
Set conn write deadline for sendAsyncWithTimeout

piggito:
When losing connection to endpoint during calls to sendAsyncWithTimeout the Write call in connection doesn't respect the timeout parameter and waits for the lifetime in connection to expire before throwing an error.

I propose to SetWriteDeadline to timeout to prevent Write call taking longer than expected.

### DIFF
--- a/uasc/secure_channel.go
+++ b/uasc/secure_channel.go
@@ -1043,9 +1043,11 @@ func (s *SecureChannel) sendAsyncWithTimeout(
 
 		// send the message
 		var n int
+		s.c.SetWriteDeadline(time.Now().Add(timeout))
 		if n, err = s.c.Write(chunk); err != nil {
 			return nil, err
 		}
+		s.c.SetWriteDeadline(time.Time{})
 
 		atomic.AddUint64(&instance.bytesSent, uint64(n))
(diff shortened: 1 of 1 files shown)


magiconair:
Thank you. I need to check the code but did you check if we always set a `timeout` or could that value be zero?

piggito:
thanks for your quick response @magiconair , as I see the `timeout` value can be as low as `RequestTimeout` which has a default of 10 seconds. We could set it to zero in client opts but then I think it is matter of how to interpret zero value in  https://github.com/gopcua/opcua/blob/main/config.go#L503 mb we can update comments and docs to interpret zero as max timeout and set uasc.MaxTimeout in code?

magiconair:
Hmm, how do other clients handle that?

piggito:
I checked https://github.com/FreeOpcUa/opcua-asyncio and I see they don't do any particular validation on timeout value but they set a default when it is None https://github.com/FreeOpcUa/opcua-asyncio/blob/master/asyncua/client/client.py#L44

magiconair (approved):

magiconair:
Looks like GitHub got confused here. Merging manually.

PR #780, opened 2025-03-19
Removed hardcoded SecurityMode in SecureChannel readChunk method.

ioansiran:
This change removes the hardcoding of SecurityMode in readChunk method. 

This hardcoding which caused any `Sign` Security Mode connection to fail due to the mismatch between given endpoint selection (for example auth-mode: Username, sec-mode: Sign, sec-method: Basic256Sha256). 
Since this line overrides the selected security mode, the server returned an EOF and immediately closed connection, regardless of authentication method and certificate validity.

To fix the concern in the previous TODO comment, I've added a check to see if there's no valid provided security method, and set a default in that case, otherwise don't override.

I'd like to add this critical bugfix to the latest version of gopcua as this library is used in production by us at [Octotronic ](https://octotronic.com/)


magiconair:
Thank you! Could you add a test as well, please? Unit or integration.

ioansiran:
> Thank you! Could you add a test as well, please? Unit or integration.

Hello. I've changed the approach a bit, and enforced a valid SecurityMode when creating the SecureChannel, and removed the override when reading chunks. This made things easier to test, since we'd want to keep config consistent throughout the lifecycle

magiconair (approved):

PR #785, opened 2025-04-01
Prevent panic on context cancellation

oliverpool:
I am using the opcua server in tests, with a short-lived context. I sometimes get a panic, because `msg` is nil.

This PR prevents this.

The diff may look large when whitespaces are not ignored, because I removed an indentation (`ctx.Err()` is more concise than `select{<-ctx.Done;default}` and more "left-aligned"). [diff ignoring whitespaces](https://github.com/gopcua/opcua/pull/785/files?diff=unified&w=1)

_BTW: thanks for this huge package!!_

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -331,41 +331,39 @@ func (s *Server) acceptAndRegister(ctx context.Context, l *uacp.Listener) {
 // monitorConnections reads messages off the secure channel connection and
 // sends the message to the service handler
 func (s *Server) monitorConnections(ctx context.Context) {
-	for {
-		select {
-		case <-ctx.Done():
-			return
-		default:
-			msg := s.cb.ReadMessage(ctx)
-			if msg.Err != nil {
-				if s.cfg.logger != nil {
-					s.cfg.logger.Error("monitorConnections: Error received: %s\n", msg.Err)
-				}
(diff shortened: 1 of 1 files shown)


magiconair:
Glad that you like it and that it is useful :) 

Regarding the PR: this doesn't look right. Contexts are checked through the `Done` channel usually in a `select` statement. 

Why don't you just add the `if msg == nil { continue }` statement here?

image

oliverpool:
> Why don't you just add the if msg == nil { continue } statement here?

This is done:
image

After doing this, I simplified the for loop from:
```go
	for {
		select {
		case <-ctx.Done():
			return
		default:
			// do something
		}
	}
```
to:
```go
	for ctx.Err() == nil {
		// do something
	}
```

This should be correct according to https://pkg.go.dev/context#Context:
> If Done is not yet closed, Err returns nil.
> If Done is closed, Err returns a non-nil error explaining why:

`ctx.Err()` is a essentially a non-blocking way to check if a context is done.

magiconair:
You are right. Since `ReadMessage()` blocks we don't create a busy loop. Just a bit wary when I see patterns I haven't seen before.

magiconair (approved):

PR #797, opened 2025-05-06
Prevent panic when browsing a gopcua server

oliverpool:
I tried the browse example against a gopcua server and got a panic, because `ua.Variant.NodeID` returned nil on `TypeIDExpandedNodeID` (seems related to https://github.com/gopcua/opcua/discussions/524).

I guess it should return the underlying node instead.

I added a test to prevent any future regression (and adjusted the default `node` value of browse to prevent questions, like https://github.com/gopcua/opcua/discussions/545)

### DIFF
--- a/examples/browse/browse.go
+++ b/examples/browse/browse.go
@@ -45,9 +45,11 @@ func join(a, b string) string {
 	return a + "." + b
 }
 
+const maxDepth = 10
+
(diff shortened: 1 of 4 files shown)


magiconair (approved):
Added a test. Otherwise, LGTM

PR #798, opened 2025-05-06
server: expose resolved URL

oliverpool:
useful when letting the OS choose a port with :0 (especially in tests)

The `srvhandshake` TODO could be fixed by this change? (however I don't know how it should behave when multiple endpoints are provided, since the listener actually listens only on one address)

https://github.com/gopcua/opcua/blob/359b38a5ff733969df3de09f23ab8e4bdd27d1de/uacp/conn.go#L305-L310

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -96,7 +96,8 @@ type security struct {
 }
 
 // New returns an initialized OPC-UA server.
-// Call Start() afterwards to begin listening and serving connections
+// At least one [EndPoint] must be provided.
+// Call [Server.Start] afterwards to begin listening and serving connections
 func New(opts ...Option) *Server {
 	cfg := &serverConfig{
 		cap:              capabilities,
@@ -236,14 +237,11 @@ func (s *Server) URLs() []string {
 }
 
 // URL returns the resolved opc endpoint that the server is listening on.
-// Useful when a :0 port was provided.
 func (s *Server) URL() string {
 	return s.l.Endpoint()
 }
 
-// Start initializes and starts a Server listening on addr
-// If s was not initialized with NewServer(), addr defaults
-// to localhost:0 to let the OS select a random port
+// Start initializes and starts a Server.
 func (s *Server) Start(ctx context.Context) error {
 	var err error
 
--- a/server/server.go
+++ b/server/server.go
@@ -209,7 +209,6 @@ func (s *Server) AddNamespace(ns NameSpace) int {
 
 	if ns.ID() == 0 {
 		return 0
-
 	}
 
 	return len(s.namespaces) - 1
@@ -236,6 +235,12 @@ func (s *Server) URLs() []string {
 	return s.cfg.endpoints
 }
 
+// URL returns the resolved opc endpoint that the server is listening on.
+// Useful when a :0 port was provided.
+func (s *Server) URL() string {
+	return s.l.Endpoint()
+}
+
 // Start initializes and starts a Server listening on addr
 // If s was not initialized with NewServer(), addr defaults
 // to localhost:0 to let the OS select a random port
@@ -256,7 +261,7 @@ func (s *Server) Start(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	log.Printf("Started listening on %v", s.URLs())
+	log.Printf("Started listening on %v", s.URL())
 
 	s.initEndpoints()
 	s.setServerState(ua.ServerStateRunning)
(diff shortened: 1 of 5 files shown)


uacp/conn.go
```diff
@@ -134,6 +135,21 @@ func Listen(ctx context.Context, endpoint string, ack *Acknowledge) (*Listener,
 	if err != nil {
 		return nil, err
 	}
+	if _, ok := strings.CutSuffix(laddr.Host, ":0"); ok {
```
magiconair: I'd prefer if you could extract this code into a separate function and write a table driven test for that.

server/server.go
```diff
@@ -236,6 +235,12 @@ func (s *Server) URLs() []string {
 	return s.cfg.endpoints
 }
 
+// URL returns the resolved opc endpoint that the server is listening on.
+// Useful when a :0 port was provided.
```
magiconair: Let's remove the "Useful..." comment and update the `server.EndPoint` option comment instead, e.g.

```
Use `0` to choose a random free port.
```

magiconair (changes requested):

server/server.go
oliverpool: done in https://github.com/gopcua/opcua/pull/798/commits/eafb6cfb105fd1cb8981b3c353b43f193fc8b520

uacp/conn.go
oliverpool: refactored in https://github.com/gopcua/opcua/pull/798/commits/91c151828d1f116ee1177f17f0e187ef062b9e21

What kind of input did you have in mind for the table-driven test? (guidance appreciated)

magiconair:
The more I look at that code the more I think that there are a number of subtle changes we should make:

* The `EndPoint` option should be `Endpoint` (deprecate the other)
* The `Endpoint` option should accept a URL instead of a `(host, port)` tuple. Maybe we can add an `EndpointHostPort` option as well which behaves like the current one. 
* The `server.New` function should be `server.NewServer` and should return an error. For example, when one of the endpoints was not a valid URL or some other part of the configuration was wrong. Deprecate `server.New` and panic on error or defer the error to `Start()` to not break existing code. 
* The `URL()` method you added is not necessary since a) a server can have multiple listening endpoints and b) there is already a `URLs()` method.
* The `replaceZeroPort` function should probably parse the URL, replace the `port` value and then format the URL again. Also, should probably be called `replacePortZero` or `findListeningPort` or something like that.

danomagnum:
Related to this, I found an issue the other day that I've started working on but not tested much yet.  We were trying to run the server inside a container which required the bind address to be different than the exposed endpoints.  We ended up having to put a 0.0.0.0 endpoint as the first endpoint to make it work.

I was working on adding a Bind() option to force a specific address and port instead of basing it off of endpoints like it does now.  I think we should probably end up supporting multiple binds but that wasn't going to be as simple of a change so I didn't pursue it yet.

https://github.com/gopcua/opcua/compare/main...danomagnum:opcua:main

magiconair:
Endpoints are the publicly visible part since they also show up in the `EndpointDescription`. `Listen` or `ListenAddr` might be a better name?

oliverpool:
Just like [http.Server.Serve](https://pkg.go.dev/net/http#Server.Serve)
I think a better approach would be to add a Serve method to accept a net.Listener. The provided listener must be wrapped in an OPC listener.

danomagnum:
> Just like [http.Server.Serve](https://pkg.go.dev/net/http#Server.Serve) I think a better approach would be to add a Serve method to accept a net.Listener. The provided listener must be wrapped in an OPC listener.

That's a good idea.  I will look into that.

oliverpool:
Ideally one should be able to call `Serve(net.Listener)` multiple times in parallel (would allow to listen on multiple interfaces at the same time).

This means that all the setup must happen during initialization (as @magiconair  wrote in https://github.com/gopcua/opcua/pull/798#issuecomment-2854550937).

So [`inithandler`](https://github.com/gopcua/opcua/blob/7c1a19ffd3b22fea5caccb68eb3444d59e82ad8b/server/service_handlers.go#L18) would be called first, so [`RegisterHandler`](https://github.com/gopcua/opcua/blob/7c1a19ffd3b22fea5caccb68eb3444d59e82ad8b/server/service_handlers.go#L96) should overwrite existing handlers instead of ignoring overwrites as it does currently.

PS: unfortunately I likely won't have much time to work with opcua in the near future.

PR #800, opened 2025-05-08
move logging to slog

### LINKED ISSUE #799: Standard log is being used instead of the custom provided one when adding a namespace to the server
It seems like the custom logger isn't always being used by the server, I have create a custom logger that can be disabled by implementing the `server.Logger` interface but whenever I add a namespace to the server the standard log is being used instead.

Looking at the code I can see that the standard Go's logging utility is being used in some cases like inside the `nodesImportNodeSet` method.

I've create a sample project below to reproduce the issue

```go
package main

import (
	"fmt"
	"os"

	"github.com/gopcua/opcua/id"
	"github.com/gopcua/opcua/server"
	"github.com/gopcua/opcua/ua"
)

type Logger struct{}

func (logger *Logger) Debug(msg string, args ...any) {
	if logEnabled {
		fmt.Print("[D] ")
		fmt.Printf(msg, args...)
		fmt.Println()
	}
}

func (logger *Logger) Error(msg string, args ...any) {
	if logEnabled {
		fmt.Print("[E] ")
		fmt.Printf(msg, args...)
		fmt.Println()
	}
}

func (logger *Logger) Info(msg string, args ...any) {
	if logEnabled {
		fmt.Print("[I] ")
		fmt.Printf(msg, args...)
		fmt.Println()
	}
}

func (logger *Logger) Warn(msg string, args ...any) {
	if logEnabled {
		fmt.Print("[W] ")
		fmt.Printf(msg, args...)
		fmt.Println()
	}
}

var endpoint = "0.0.0.0"
var port = 4840
var logEnabled bool = false //NOTE: I'm disabling the logs here, so we shouldn't see anything
var logger = new(Logger)

func main() {
	var opts []server.Option

	opts = append(opts, server.SetLogger(logger))
	opts = append(opts, server.EnableSecurity("None", ua.MessageSecurityModeNone))
	opts = append(opts, server.EnableAuthMode(ua.UserTokenTypeAnonymous))

	hostname, err := os.Hostname()
	if err != nil {
		log.Fatal("Error getting host name %v", err)
	}

	opts = append(opts,
		server.EndPoint(endpoint, port),
		server.EndPoint("localhost", port),
		server.EndPoint(hostname, port),
	)

	srv := *server.New(opts...)
	rootNs, _ := srv.Namespace(0)

	logger.Info("This shouldn't print anything until I'm turning the logs on") // This is not logging as we disabled the log
	testns := server.NewNodeNameSpace(&srv, "testns")                          // This (or the lines below) are logging even if the log is disabled
	testnsObjects := testns.Objects()
	rootNs.Objects().AddRef(testnsObjects, id.HasComponent, true)

	logEnabled = true

	logger.Info("This should be the only line appearing in the terminal")

	srv.Close()
}

```

magiconair:
This patch addresses some of the logging issues in the server but not all:

* standardize the receiver names for all server methods
* replace the `nil` checks with a `DiscardLogger`
* replace `log.Printf` with calls to the server logger

However, the `server.Logger` interface states that it is compatible with `slog.Logger` but the formatting relies on the `Printf` semantics. However, to be `slog` compatible the parameters need to passed as key/value pairs:

https://pkg.go.dev/log/slog#pkg-overview
> A log record consists of a time, a level, a message, and a set of key-value pairs, where the keys are strings and the values may be of any type. 

So the log messages have to be updated as well. I'll keep updating this patch.

The larger issue is whether this should be reflected in the client as well since that has it's own logging with the `debug` package. Not sure if I want to include that in this patch or separate this. 

Fixes #799


kung-foo:
maybe i've misunderstood the new slog package, but isn't the idea that our library should accept a `slog.Handler`? Then we use the actual `slog.Logger` instances internally, and the caller decides how the deal with (or not) the output? As is right now, our `Logger` interface is a limited version of the `slog.Logger` (i.e. missing `.With` and context related funcs).

magiconair:
Hmm, that's not how I understand it. Where would the current log level be defined? Not in the handler, is it? The handler only formats the output, or not?

https://github.com/golang/example/blob/master/slog-handler-guide/README.md#loggers-and-their-handlers

magiconair:
When I read the "Design Process" in https://go.dev/blog/slog#the-design-process then I think the split into `Logger` and `Handler` is so that existing logging libraries like zig and logrus can use the `Handler` to avoid rewriting all the logging code. 

I guess we could take that approach as well but since we already have two different logging approaches in the code for both client and server I'm OK with investing a bit of time to rewrite the log messages to conform with `log/slog`. 

The only issue is that the output of the log messages will change and I'm wondering if this is an issue?

What do you think?

danomagnum:
I think when I first changed the logging to use "slog" i was wrongly assuming it was using the same format as the classic log library for its parameters (after all, the function signature is the same).  We since migrated a lot of logging in other projects over from log to slog and found that isn't the case.  I don't think I've looked back at any of the opc logs since then.  (Just looked and confirmed getting `!BADKEY` errors in the logs presently since they aren't always key:value pairs after the message.)  If you'd like I will volunteer to go through and convert all the printf style logs to slog style key:value pairs.

Now that I've used slog more adding the `With` method would make sense in a lot of places of the server.  For example the sessions could create a child logger with the session info.  

If worried about changing the output of the log messages, we could set up a `legacyLogger` that would meet the logging interface and still output in the old fashion and use that as the default instead of a `DiscardLogger`.  This would also allow use of `With` instead of `NewPrefixLogger` in the conversion of the client code to achieve a similar effect.

magiconair:
@danomagnum don't worry about it. You've contributed the server code and pushed it over the line - finally. Logging has been an issue for a while. A good time to do an overhaul. So let's do it right.

magiconair:
Other code shouldn't be dependent on server logs but some monitoring code might be. We can call it 0.9 and announce it beforehand.

kung-foo:
For reference Uber's Zap logger (which we use at Intelecy), bridges their logger with `slog` via: https://github.com/uber-go/zap/blob/master/exp/zapslog/handler.go#L53-L62

So it seems like the approach could either be options like `.WithLogger(*slog.Logger)` or `.WithLogHandler(slog.Handler)`

magiconair:
@kung-foo I think you're right that we use the `slog.Logger` as frontend with a handler that is provided to us. I've taken a stab at this by creating an `internal/ualog` package which maps the public functions of the `log/slog` package and adds a `Fatal` helper since we need this in all the examples. We also need the ability to set the Debug log level through the command line. I've added that to `ualog` as well.

Consider this the first stab at this but that should be easier to refactor.

I've created an `slog` wrapper for the server but the `nil` checks still need to be removed and the log messages be rewritten. It wasn't that much work after all. Couple of hours.

magiconair:
There is an import cycle in the integration tests. Maybe someone can have a look...

```
❯ OPC_DEBUG=debug go test -tags integration ./...
# github.com/gopcua/opcua/uasc
package github.com/gopcua/opcua/uasc
        imports github.com/gopcua/opcua/tests/python from secure_channel_test.go
        imports github.com/gopcua/opcua from py_server.go
        imports github.com/gopcua/opcua/uasc from client.go: import cycle not allowed in test
```

magiconair:
Also, still need to do a good sweep over the changes. While we're at it I'd also like to re-format the error messages the same way, e.g. from `sent 15 bytes` to `sent message msg_size=15` or something like that.

internal/ualog/ualog.go
```diff
@@ -0,0 +1,49 @@
...
+// if [debug] is true.
+func SetDebugLogger(debug bool) {
+	logLevel := slog.LevelInfo
+	if debug {
+		logLevel = slog.LevelDebug
+	}
+	logger := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{
+		Level: logLevel,
+	}))
+	slog.SetDefault(logger)
```
kung-foo: Would this be considered unexpected behavior? If I'm using this lib, and I have my own default logger configured, this would step on it.

client.go
```diff
@@ -538,15 +537,15 @@ func (c *Client) monitor(ctx context.Context) {
 						activeSubs = 0
 						for _, subID := range subsToRepublish {
 							if err := c.republishSubscription(ctx, subID, availableSeqs[subID]); err != nil {
-								dlog.Printf("republish of subscription %d failed", subID)
+								dlog.Debug("republish of subscription failed", "sub_id", subID)
```
kung-foo: More of a style comment than a technical, but I try to use the typed `slog.Int(...)` rather than the awkward "pairs". Though the lack of `slog.Error(err)` annoys me to no end.

client_sub.go
```diff
@@ -149,11 +151,17 @@ func (c *Client) republishSubscription(ctx context.Context, id uint32, available
...
 // more messages to restore.
 func (c *Client) sendRepublishRequests(ctx context.Context, sub *Subscription, availableSeq []uint32) error {
+	dlog := ualog.With("func", "Client.sendPublishRequests")
+
 	// todo(fs): check if sub.nextSeq is in the available sequence numbers
 	// todo(fs): if not then we need to decide whether we fail b/c of data loss
 	// todo(fs): or whether we log it and continue.
 	if len(availableSeq) > 0 && !slices.Contains(availableSeq, sub.nextSeq) {
-		log.Printf("sub %d: next sequence number %d not in retransmission buffer %v", sub.SubscriptionID, sub.nextSeq, availableSeq)
+		ualog.Info("sub: next sequence number not in retransmission buffer",
```
kung-foo: should this be `dlog`?

client_sub.go
```diff
@@ -162,38 +170,38 @@ func (c *Client) sendRepublishRequests(ctx context.Context, sub *Subscription, a
 			RetransmitSequenceNumber: sub.nextSeq,
 		}
 
-		debug.Printf("Republishing subscription %d and sequence number %d",
-			req.SubscriptionID,
-			req.RetransmitSequenceNumber,
+		dlog.Debug("Republishing subscription and sequence number",
```
kung-foo: and technically anywhere we have a `ctx` we should use the `logger.XXXContext(...)` func. But that can also be done a future PR.

internal/ualog/ualog.go
```diff
@@ -0,0 +1,49 @@
...
+	Error = slog.Error
+
+	DebugContext = slog.DebugContext
+	InfoContext  = slog.InfoContext
+	WarnContext  = slog.WarnContext
+	ErrorContext = slog.ErrorContext
+)
+
+// Fatal is a convenience function for logging an error and exiting immediately.
+func Fatal(msg string, args ...any) {
```
kung-foo, line 56: Should we even have this? Isn't the general consensus that a library should _never_ purposely panic and/or call `os.Exit`?

kung-foo (commented):
some initial comments

kung-foo:
> There is an import cycle in the integration tests. Maybe someone can have a look...
> 
> ```
> ❯ OPC_DEBUG=debug go test -tags integration ./...
> # github.com/gopcua/opcua/uasc
> package github.com/gopcua/opcua/uasc
>         imports github.com/gopcua/opcua/tests/python from secure_channel_test.go
>         imports github.com/gopcua/opcua from py_server.go
>         imports github.com/gopcua/opcua/uasc from client.go: import cycle not allowed in test
> ```

fix incoming

internal/ualog/ualog.go
magiconair: The main motivation for the `internal/ualog` package is to capture some helper functions we need for this library but that we don't export. `ualog.SetDebugLogger` and `ualog.Fatal` are helpers that are used in all the examples and none of the library code. 

Once I started using `ualog.Fatal` it felt weird to use `slog.Info` in other places and I thought this creates confusion. That's why I've mapped all the `slog` globals into `ualog`. 

I'm open for suggestions.

client.go
magiconair: I don't have a strong opinion on this and am fine with the "pairs". Seems lighter and we are using type inference in Go code. I don't see why we would use explicit types for logging but I also understand this evoked strong opinions. As for `slog.Error`: could also be a helper but maybe it's not worth it.

client_sub.go
magiconair: We can but I didn't fully understand the benefit of passing on the context since we're not storing any data in it. Would this also cancel writes to the log store?

internal/ualog/ualog.go
magiconair: Only used in the examples. See the rationale in the first comment. Let's find something that works. Maybe have an `examples/internal/slogx` package or make it explicit?

client_sub.go
magiconair: Need to check

client_sub.go
kung-foo: in our case, we have an incoming context that contains otel tracing headers. that same context is passed down into read/write/browse, etc. So I would want anything in the context to be added to the log message (depending on the handler of course). Might not be as important on the server branch, but i can see things like the client session info being of interest to a logger.

client.go
kung-foo: Yeah, the RFC did have some :fire:. To me the typed attrs are cleaner and lighter since there's no (or at least less) inference happening. Plus it can be type checked, rather than relying on linters. Also I'm hoping in the future to use the `LogValuer` interface to allow us to optionally do things like log nodes, sessions, etc. as strongly typed (and consistent). For example: `dlog.DebugContext(ctx, "node added", ualog.Node(node))`

internal/ualog/ualog.go
kung-foo: ah right, forgot this is in the `internal` package. however there is still the risk that another package dev will use this. i guess we just monitor the PRs. personally i'd rather see the examples just call `ualog.Error(...)` and then `os.Exit()` themselves. But not critical now.

magiconair: That's how I started and if we go down this route then we don't need the `ualog` package. We then only need a convenience function to set the log level to DEBUG and somehow I'm too dense to figure out another way.

client.go
magiconair: I guess we wouldn't want to see each number in Go explicitly type casted (e.g. `int32(5)` everywhere) and for the same reason I see more noise than value in using the explicit helpers. But I haven't spent a lot of time with `slog`. Why don't we leave it like this for now and add them where it makes sense?

client_sub.go
magiconair: Fair enough. I'll add them.

client_sub.go
magiconair: yes. fixed

internal/ualog/ualog.go
magiconair: I think I'm with you that `ualog` in its current form will get a life of its own. Let's only keep the `Fatal` and `SetDebugLogger` as helpers for the examples, keep the package internal and use plain `slog` anywhere else.

magiconair: Just got a better idea for the `SetDebugLogger`. We only need the `TextHandler` with the right log level. In the examples we can write this which should make this clear.

```go
slog.SetDefault(slog.New(ualog.NewTextHandler(*debug)))
```

magiconair: Trimmed `ualog` down to two functions which are only used in the examples.

internal/ualog/ualog.go
magiconair: I've kept the `ualog.Fatal` as one of two helpers for the examples in `ualog`.

sruehl:
Just stumbled over this today too (but looking at debug.go). In our project we use zerolog, so far as I understand we have the ability to plug a slog later then and I can use something like https://github.com/samber/slog-zerolog to bridge gopcua over slow to zerolog if I got that right, right?

magiconair:
@sruehl that's how it should work. Feedback is welcome. The client uses the `slog.Default()` logger but maybe we can make that configurable as well. You'd provide a different handler.

sruehl:
having that configurable on the client would be awesome.

sruehl:
~So maybe remove the `debug.go` completely in favor of a ClientOption? Not sure how you want to handle `debug.Logger` but even the flags could be transformed to a option then too (or keep them as fallback for compability).~

Edit: should have looked closer but `debug.go` already looks right 🤦🏼. So then only a ClientOption would be missing?

magiconair:
Hmm, @sruehl this means injecting the logger into other components as well since we can no longer use global functions. Like in the Dialer, Listener, Conn, Subscription, etc. 

How are other libraries handling this? Can't we just use the default handler and you configure that one?

sruehl:
In plc4go we pass them down the chain either use arguments or the ctx. Default handler works too but keep in mind then  it is impossible to define different loggers for different instances (E.g. what you might want to do in zerolog is to have loggers with certain key/values attached to produce events which you might want to use to segment them in your logging analyzer. When I have to use a global library logger I can't do that anymore.).

edit: In case of "objects" the logger is then a field of that struct

sruehl:
Oh and to give you (@magiconair) a usecase which is quite vital at the moment: I have a application which produces humongous amount of logs. In that application opcua is only a small part. So if I want to filter by logs only by that part usually I filter for logs with a field `protocol:opcua`. This of course only works if every log statement in that component produces that k/v. This case could work with a default handler as all of those should have that in common. So now I have a problem with a particular server: So I add an additional filter `opcuaHost:192.168.1.202` to my filter to only view logs which a related to this particular server. I only see logs with that key which all my loggers produce but gopcua doesn't produce that key. Now you can see that the global approach brakes appart as I would have no way of assigning different loggers to different client instances. Does that make sense?

magiconair:
Yes, that makes sense. It also clarifies who is in charge of adding additional tags. I've just experimented with adding a `pkg=gopcua.Server` to the server logger but that would clearly fall into the responsibility of the code setting up the server.

```
❯ go test -tags integration -v ./tests/go -run Stats -- -debug
=== RUN   TestStats
2025/05/14 11:42:30 ERROR Duplicate reference pkg=gopcua.Server type=FromState
2025/05/14 11:42:30 ERROR Duplicate reference pkg=gopcua.Server type=HasSubtype
2025/05/14 11:42:30 ERROR Duplicate reference pkg=gopcua.Server type=HasModellingRule
2025/05/14 11:42:30 ERROR Duplicate reference pkg=gopcua.Server type=HasDescription
2025/05/14 11:42:30 ERROR Duplicate reference pkg=gopcua.Server type=HasFalseSubState
2025/05/14 11:42:30 ERROR Duplicate reference pkg=gopcua.Server type=Organizes
2025/05/14 11:42:30 ERROR Duplicate reference pkg=gopcua.Server type=HasTypeDefinition
2025/05/14 11:42:30 ERROR Duplicate reference pkg=gopcua.Server type=HasGuard
2025/05/14 11:42:30 ERROR Duplicate reference pkg=gopcua.Server type=HasTrueSubState
2025/05/14 11:42:30 ERROR Duplicate reference pkg=gopcua.Server type=HasAddIn
2025/05/14 11:42:30 ERROR Duplicate reference pkg=gopcua.Server type=HasInterface
2025/05/14 11:42:30 ERROR Duplicate reference pkg=gopcua.Server type=HasCondition
2025/05/14 11:42:30 ERROR Duplicate reference pkg=gopcua.Server type=HasComponent
2025/05/14 11:42:30 ERROR Duplicate reference pkg=gopcua.Server type=HasEncoding
2025/05/14 11:42:30 ERROR Duplicate reference pkg=gopcua.Server type=HasCause
2025/05/14 11:42:30 ERROR Duplicate reference pkg=gopcua.Server type=ToState
2025/05/14 11:42:30 ERROR Duplicate reference pkg=gopcua.Server type=HasEffect
2025/05/14 11:42:30 ERROR Duplicate reference pkg=gopcua.Server type=HasDictionaryEntry
2025/05/14 11:42:30 ERROR Duplicate reference pkg=gopcua.Server type=HasProperty
2025/05/14 11:42:30 ERROR Duplicate reference pkg=gopcua.Server type=HasEventSource
2025/05/14 11:42:30 ERROR Duplicate reference pkg=gopcua.Server type=HasNotifier
2025/05/14 11:42:30 INFO Started listening pkg=gopcua.Server url=[opc.tcp://localhost:4840]
2025/05/14 11:42:30 INFO Registered connection pkg=gopcua.Server remote_addr=[::1]:62436
2025/05/14 11:42:30 INFO Registered new channel pkg=gopcua.Server secchan_id=482782124 secchan_count=1
2025/05/14 11:42:30 ERROR monitorConnections: Unknown SecureChannel pkg=gopcua.Server secchan_id=0
2025/05/14 11:42:30 WARN Secure Channel closed pkg=gopcua.Server secchan_id=482782124
2025/05/14 11:42:30 ERROR Socket closed pkg=gopcua.Server error="accept tcp [::1]:4840: use of closed network connection"
```

magiconair:
I'll remove that again.

magiconair:
@sruehl do you want to pass in the logger or the handler? Right now, I'm setting things up to configure the handler.

sruehl:
good question: I haven't looked into slog too much as we use zerolog all over the place but from the looks of it Handler might be the better way.

sruehl:
AFAIK a handler would then look like this:

```golang
type ZerologHandler struct {
    logger zerolog.Logger
}

func (h *ZerologHandler) Log(level slog.Level, msg string, kvs ...interface{}) error {
    var zerologLevel zerolog.Level
    switch level {
    case slog.LevelDebug:
        zerologLevel = zerolog.DebugLevel
    case slog.LevelInfo:
        zerologLevel = zerolog.InfoLevel
    case slog.LevelWarn:
        zerologLevel = zerolog.WarnLevel
    case slog.LevelError:
        zerologLevel = zerolog.ErrorLevel
    default:
        zerologLevel = zerolog.TraceLevel
    }

    event := h.logger.WithLevel(zerologLevel).Str("msg", msg)

    for i := 0; i < len(kvs); i += 2 {
        if i+1 >= len(kvs) {
          break
        }
        event.Interface(kvs[i].(string), kvs[i+1])
    }

    event.Send()
    return nil
}
```
Looks good and this way you just plug your existing logger in

magiconair:
Finished reworking the server logs. 

@danomagnum I think the log messages could use an overhaul. Maybe you can have a look which ones are still needed and which ones could be removed. Would you be up for that?

magiconair:
@danomagnum could you have a look at the last commit where I make server fields private? There is a weird mix of fields in the server services which are public and private and it isn't clear why. Fields are lacking comments. Maybe you can address that.

What stands out is that you expose the `sync.Mutex` which is something that we shouldn't do. If something needs to be guarded with a lock then that should be hidden in a method. 

I think we should do a separate round of server code cleanup.

danomagnum:
@magiconair 

I'll take a look at the logs and the private/public stuff.  I agree it would probably be a good time to go back through it now since it has been a while since I worked on it in depth and I'll have a better eye for problems.

magiconair:
@danomagnum Awesome. Thank you!

magiconair:
I've pushed the first stab at a client logger and extended the `ualog` package to store a logger in the context. If there is already a standard way of doing this then let's use that. 

I've also used the `examples/accesslevel` code to see what using different loggers would look like. 

This change is getting quite big.

server/monitored_item_service.go
```diff
@@ -139,11 +146,7 @@ func (s *MonitoredItemService) ChangeNotification(n *ua.NodeID) {
 }
 
 func (s *MonitoredItemService) NextID() uint32 {
-	i := atomic.AddUint32(&s.id, 1)
-	if i == 0 {
-		i = atomic.AddUint32(&s.id, 1)
-	}
-	return i
+	return atomic.AddUint32(&s.nextID, 1)
```
danomagnum: I looked at a branch that wasn't squashed and this was added in a commit called "[Fixes from testing with UAExpert](https://github.com/danomagnum/opcua/commit/6e088e910b3c8754d269ee58413562a66c658f41#diff-e3f30690c426d1dde692c4a8eb93dfdec1079e0f8671803db4a6ec4cfa8d14e0L101)".  I don't recall exactly why I added the check - This shouldn't return a 0 until you've gone through 4 billion monitored items and I certainly didn't test that many.

I do know at some point there was a 0 in some value (maybe subscription ID?) which caused an issue so I had to explicitly skip zero.  It is possible I added this check here to prevent the potential for that same problem.

danomagnum:
One thing to note - you don't have to `with` the function name.  You can get that automatically by settin `AddSource` option in the handler.  Then you can set up a `ReplaceAttr` function to pick what parts of the function/file/line/etc... you want logged.

as an example:
```Go
opts := slog.HandlerOptions{
		AddSource: true,
		ReplaceAttr: func(groups []string, a slog.Attr) slog.Attr {
			if a.Key == slog.SourceKey {
				s := a.Value.Any().(*slog.Source)
				s.File = path.Base(s.File)
				fn := s.Function
				fns := strings.Split(s.Function, "/")
				if len(fns) > 1 {
					fn = fns[len(fns)-1]
				}
				sd := sourceDescr{
					File: s.File,
					Line: s.Line,
					Fn:   fn,
				}
				a.Value = slog.AnyValue(sd)
			}
			return a
		},
	}
```

Of course this would depend on the user to set it up when configuring the logger initially so if you always wanted the function it can stay.

danomagnum:
I've spent some time looking at the server logging in general this evening.  It makes sense to me that a secure channel logger that has been instantiated something like `s.logger = logger.With("secure_channel_id", secureChannelID, "remote_addr", c.RemoteAddr().String())` would make the most sense.  Some functions already get the sc passed in.  Others would need to be modified to take a logger so they can get the right logger.  That's not a big deal.

The secure channel's logger is private to the uasc package though.  The three ways I see are 
1. Make the logger in the secure channel public
2. Add a Logger() function to return the channel's logger
3. Keep my own mapping of appropriately configured loggers in the server and pass those into functions as they are called.

Is there one of these that is preferable or maybe there's a different way you'd prefer?

server/monitored_item_service.go
magiconair: I don't see how a `uint32` with `AddUint32` could ever be zero except for the roll-over. Good point. I'll add it back in with a comment.

magiconair: pushed

magiconair:
> One thing to note - you don't have to `with` the function name. You can get that automatically by settin `AddSource` option in the handler. Then you can set up a `ReplaceAttr` function to pick what parts of the function/file/line/etc... you want logged.
> 
> as an example:
> 
> ```go
> opts := slog.HandlerOptions{
> 		AddSource: true,
> 		ReplaceAttr: func(groups []string, a slog.Attr) slog.Attr {
> 			if a.Key == slog.SourceKey {
> 				s := a.Value.Any().(*slog.Source)
> 				s.File = path.Base(s.File)
> 				fn := s.Function
> 				fns := strings.Split(s.Function, "/")
> 				if len(fns) > 1 {
> 					fn = fns[len(fns)-1]
> 				}
> 				sd := sourceDescr{
> 					File: s.File,
> 					Line: s.Line,
> 					Fn:   fn,
> 				}
> 				a.Value = slog.AnyValue(sd)
> 			}
> 			return a
> 		},
> 	}
> ```
> 
> Of course this would depend on the user to set it up when configuring the logger initially so if you always wanted the function it can stay.

I don't think we can modify the handler the user passes in. So this would depend on the user. But we could add a helper function to `ualog` to assist that.

sruehl:
FYI: stumpled over this here: https://messagetemplates.org. TLDR: define messages using placeholders to be able to print a human readable text at the end

danomagnum:
I set up a context in the server that gets populated with the client and service type when the service handler is dispatched.  It then gets passed down in the context to any other functions.  These changes are over in the MapNamespaceRefactor branch.  Take a look and see what you think about this approach.

istyf:
What is the status of this PR? I would like to go through the server code to see if I can use it for a project, but with such a big PR "incoming" it is nice to know what to expect in the near future. :)

danomagnum:
I haven't had a chance to look at it in a little bit.  Hopefully I can get back to it now that the weather is cooling off.

istyf:
As a heads up I am working on a counter proposal based on my findings when trying to integrate the server in another project. Hoping to push a PR today or tomorrow to get your feedback.

sruehl:
@danomagnum: do you see and issues preventing us from getting this merged now?

danomagnum:
> @danomagnum: do you see and issues preventing us from getting this merged now?

No, I don't foresee any issues - I haven't had a chance to review it though.

PR #807, opened 2025-06-02
feat: add state change func

sruehl:
Addition to #767.

Rational: as #767 handles the calls more or less synchronous this provides a simpler way to hook into state changes.


sruehl:
@magiconair added some comments to #767 but could be that is drowned in the notifications.

kung-foo (approved):

PR #817, opened 2025-07-12
Fix Basic256Sha256 server (Fixes #815)

### LINKED ISSUE #815: Status of `Basic256Sha256` Encryption?
I see in the README feature support matrix that `Basic256Sha256` is "Untested". I have a working opcua server when not using encryption, but no matter what I try I am unable to get `Basic256Sha256` and certificates working. Seeing "Untested" in the README makes me think it might not work as is.

I am willing to dive in and try to fix this in the code if needed, but I would love to understand the status of `Basic256Sha256` and the history of its implementation.

If basic functionality is expected to work, I would love to see a minimal server / client implementation (gopcua implementation of a server that supports `Basic256Sha256` and gopcua client that can subscribe to it.)

jminardi:
When enabling `Basic256Sha256` on the demo server I was seeing `invalid node id type` errors.

This was because the OPC UA standard expects that all initial handshakes utilize asymmetrical encryption, so Basic256Sha256 must be enabled until the channel is properly opened and the server can read the client's requested encryption mode.

See here:
https://reference.opcfoundation.org/Core/Part4/v104/docs/5.5.2
> The OpenSecureChannel request and response Messages shall be signed with the sender's private key. These Messages shall always be encrypted.

### DIFF
--- a/examples/server/node_server/server.go
+++ b/examples/server/node_server/server.go
@@ -61,13 +61,13 @@ func main() {
 	// Set your security options.
 	opts = append(opts,
 		server.EnableSecurity("None", ua.MessageSecurityModeNone),
+		server.EnableSecurity("Basic256Sha256", ua.MessageSecurityModeSignAndEncrypt),
 		/*
 				These security modes are not implemented yet.
 			server.EnableSecurity("Basic128Rsa15", ua.MessageSecurityModeSign),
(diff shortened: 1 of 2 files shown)


jminardi:
@magiconair Is there anything you'd like me to add or update to consider merging this PR in?

kung-foo (approved):

PR #819, opened 2025-07-14
fix: avoid panic on atomic load

### LINKED ISSUE #818: Using atomics which are not set panics
see https://github.com/gopcua/opcua/blob/7c1a19ffd3b22fea5caccb68eb3444d59e82ad8b/client.go#L703
or https://github.com/gopcua/opcua/blob/7c1a19ffd3b22fea5caccb68eb3444d59e82ad8b/client.go#L691
instead of this returning nil it panics. We should add a guard check there.
Edit: apparently this happens when you create a client with `new(Client)`

Edit: Root Cause is in the comment, analysis was wrong, investigating....

sruehl:
fixes #818

### DIFF
--- a/client.go
+++ b/client.go
@@ -688,7 +688,11 @@ func (c *Client) setPublishTimeout(d time.Duration) {
 // During reconnect this value can change.
 // Make sure to capture the value in a method before using it.
 func (c *Client) SecureChannel() *uasc.SecureChannel {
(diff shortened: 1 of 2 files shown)


sruehl:
Test doesn't run on main when the guards in load are not active

magiconair (approved):

PR #822, opened 2025-08-04
remove markdown filter

kung-foo:
GitHub still creates a test requirement even if all files are excluded. But the test never runs, so the action hangs forever.


danomagnum (approved):
Looks good to me.  Sounds like this is a known issue with github actions.

PR #824, opened 2025-09-03
Account for signature padding as well as padding length byte (Fixes #781)

### LINKED ISSUE #781: EOF in readChunk when reading 1500 nodes
Moved from https://github.com/gopcua/opcua/issues/658#issuecomment-1903473439

@magiconair Hello

I use Telegraf OPC UA plugin, which uses gopcua.
I am currently encountering the same situation. Using security and OPCUA to subscribe, EOF occurs when Nodes exceeds 1000 or 1500.
the logs..
```
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] Connecting OPC UA Client to server
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uacp: connecting to opc.tcp://192.168.1.XX:48050
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uacp 1: start HEL/ACK handshake
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uacp 1: sent HELF with 60 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uacp 1: recv ACKF with 28 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uacp 1: recv &uacp.Acknowledge{Version:0x0, ReceiveBufSize:0xffff, SendBufSize:0xffff, MaxMessageSize:0xffff00, MaxChunkCount:0x100}
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 1/1: send *ua.OpenSecureChannelRequest with 132 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uacp 1: recv OPNF with 136 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 1/1: recv OPNF with 136 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 1/1: recv *ua.OpenSecureChannelResponse
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 1/1: sending *ua.OpenSecureChannelResponse to handler
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 1: received security token. channelID=2437948123 tokenID=1 createdAt=2024-01-19T06:57:35Z lifetime=1h0m0s
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 1: security token is refreshed at 2024-01-19T07:42:41Z (45m0s). channelID=2437948123 tokenID=1
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 1: security token expires at 2024-01-19T08:12:35Z. channelID=2437948123 tokenID=1
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 1/2: send *ua.GetEndpointsRequest with 97 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uacp 1: recv MSGF with 3690 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 1/2: recv MSGF with 3690 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 1/2: recv *ua.GetEndpointsResponse
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 1/2: sending *ua.GetEndpointsResponse to handler
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 1: Close()
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 1/3: send *ua.CloseSecureChannelRequest with 57 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uacp 1: close
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 1: readChunk EOF
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] Configuring OPC UA connection options
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] Loading cert/key from /etc/telegraf/ua.pem//etc/telegraf/ua_key.pem
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] security policy from configuration 
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uacp: connecting to opc.tcp://192.168.1.XX:48050
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uacp 2: start HEL/ACK handshake
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uacp 2: sent HELF with 60 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uacp 2: recv ACKF with 28 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uacp 2: recv &uacp.Acknowledge{Version:0x0, ReceiveBufSize:0xffff, SendBufSize:0xffff, MaxMessageSize:0xffff00, MaxChunkCount:0x100}
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/1: send *ua.OpenSecureChannelRequest with 1905 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uacp 2: recv OPNF with 1901 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2: setting securityPolicy to http://opcfoundation.org/UA/SecurityPolicy#Basic256Sha256
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/1: recv OPNF with 1901 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/1: recv *ua.OpenSecureChannelResponse
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/1: sending *ua.OpenSecureChannelResponse to handler
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2: received security token. channelID=2437948124 tokenID=1 createdAt=2024-01-19T06:57:35Z lifetime=1h0m0s
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2: security token is refreshed at 2024-01-19T07:42:41Z (45m0s). channelID=2437948124 tokenID=1
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2: security token expires at 2024-01-19T08:12:35Z. channelID=2437948124 tokenID=1
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/2: send *ua.CreateSessionRequest with 1584 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uacp 2: recv MSGF with 5408 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/2: recv MSGF with 5408 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/2: recv *ua.CreateSessionResponse
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/2: sending *ua.CreateSessionResponse to handler
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/3: send *ua.ActivateSessionRequest with 464 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uacp 2: recv MSGF with 144 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/3: recv MSGF with 144 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/3: recv *ua.ActivateSessionResponse
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/3: sending *ua.ActivateSessionResponse to handler
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/4: send *ua.ReadRequest with 144 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uacp 2: recv MSGF with 480 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/4: recv MSGF with 480 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/4: recv *ua.ReadResponse
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/4: sending *ua.ReadResponse to handler
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] Connected to OPC UA Server
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] Creating OPC UA subscription
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/5: send *ua.CreateSubscriptionRequest with 128 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uacp 2: recv MSGF with 112 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/5: recv MSGF with 112 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/5: recv *ua.CreateSubscriptionResponse
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/5: sending *ua.CreateSubscriptionResponse to handler
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] Subscribed with subscription ID 2432947172
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/6: send *ua.CreateMonitoredItemsRequest with 65536 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/6: send *ua.CreateMonitoredItemsRequest with 20544 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2/7: send *ua.PublishRequest with 112 bytes
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2: readChunk EOF
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uacp 2: close
2024-01-19T14:57:41+08:00 D! [inputs.opcua_listener] uasc 2: Close()
2024-01-19T14:57:41+08:00 E! [telegraf] Error running agent: starting input inputs.opcua_listener: failed to start monitoring items: EOF
```
Is there currently a solution to this problem?

jminardi:
This does fix the problem for me, although I am not 100% sure it's the correct fix. Here is my thinking:

I was able to reproduce the oversize message bug with an opcua server simulator. It happened on the very first message that needed to be split up into chunks. 

The `maxBodySize` calculation seems to be missing an allowance for the size of the signature padding. Signature padding is added in sign and encrypt, but not subtracted from the body size available. This means that any message that uses signAndEncrypt will fail if it's larger than the available message size and needs to be chunked. That's why the workaround of subscribing to less nodes works, as it lets the message fit inside one chunk.

By accounting for the worse case padding of 16 bytes + 1 byte for the padding size byte, the messages are small enough to fit, and I was able to subscribe to 10s of thousands of nodes with 1 client and 1 request.

It is possible that this is a red herring, and any "fix" that reduced the `maxBodySize` would have worked, but from my investigation, it does seem that the signature padding isn't accounted for so this feels at least plausible.

---
_Maintainer note (JeremyTheocharis): verified correct against OPC UA Part 6 v1.05 §6.7.2.5 and the current open62541 and OPC Foundation .NET reference implementations; see the review for details. Merging for v0.9.0._

Fixes #781

### DIFF
--- a/uasc/secure_channel_instance.go
+++ b/uasc/secure_channel_instance.go
@@ -137,11 +137,13 @@ func (c *channelInstance) SetMaximumBodySize(chunkSize int) {
 	sequenceHeaderSize := 8
 	headerSize := 12
 	symmetricAlgorithmHeader := 4
+	signaturePadding := c.algo.PlaintextBlockSize() + 1
 
 	// this is the formula proposed by OPCUA - source node-opcua
+	// Modified to reserve worst-case padding + padding length byte
 	maxBodySize :=
 		c.algo.PlaintextBlockSize()*
-			((chunkSize-headerSize-symmetricAlgorithmHeader-c.algo.SignatureLength()-1)/c.algo.BlockSize()) -
+			((chunkSize-headerSize-symmetricAlgorithmHeader-c.algo.SignatureLength()-signaturePadding)/c.algo.BlockSize()) -
 			sequenceHeaderSize
 	c.maxBodySize = uint32(maxBodySize)
 
--- a/uasc/secure_channel_instance.go
+++ b/uasc/secure_channel_instance.go
@@ -137,19 +137,34 @@ func (c *channelInstance) SetMaximumBodySize(chunkSize int) {
 	sequenceHeaderSize := 8
 	headerSize := 12
 	symmetricAlgorithmHeader := 4
-	signaturePadding := c.algo.PlaintextBlockSize() + 1
 
-	// this is the formula proposed by OPCUA - source node-opcua
-	// Modified to reserve worst-case padding + padding length byte
+	// signAndEncrypt appends one PaddingSize byte (two if ExtraPaddingSize is
+	// present, i.e. for signatures longer than 256 bytes) before padding the
+	// plaintext to a block boundary.
+	paddingSizeBytes := 1
+	if c.algo.RemoteSignatureLength() > 256 {
+		paddingSizeBytes = 2
+	}
+
+	// OPC UA Part 6, 6.7.2.5 defines
+	//
+	//   MaxBodySize = PlainTextBlockSize *
+	//       Floor((MessageChunkSize - HeaderSize - 1) / CipherTextBlockSize) -
+	//       SequenceHeaderSize - SignatureSize
+	//
+	// where the -1 reserves room for the PaddingSize byte. Subtracting it
+	// inside the Floor only takes effect when chunkSize-headerSize is an exact
+	// multiple of the cipher block size. For any other chunk size (e.g. the
+	// default 65535) a maximum-size body pads out to one cipher block more
+	// than fits in the chunk. Reserving the PaddingSize byte(s) outside the
+	// Floor instead is exact: a body of maxBodySize fills the chunk to the
+	// last whole cipher block and one more byte no longer fits, for every
+	// chunk size (see TestMaxBodySizeFitsChunk).
 	maxBodySize :=
 		c.algo.PlaintextBlockSize()*
-			((chunkSize-headerSize-symmetricAlgorithmHeader-c.algo.SignatureLength()-signaturePadding)/c.algo.BlockSize()) -
-			sequenceHeaderSize
+			((chunkSize-headerSize-symmetricAlgorithmHeader)/c.algo.BlockSize()) -
+			sequenceHeaderSize - c.algo.SignatureLength() - paddingSizeBytes
 	c.maxBodySize = uint32(maxBodySize)
-
-	// this is the formula proposed by ERN - source node-opcua
-	// maxBlock := (chunkSize - headerSize) / c.algo.BlockSize()
-	// c.maxBodySize = c.algo.PlaintextBlockSize()*maxBlock - sequenceHeaderSize - c.algo.SignatureLength() - 1
 }
 
 // signAndEncrypt encrypts the message bytes stored in b and returns the
@@ -179,8 +194,18 @@ func (c *channelInstance) signAndEncrypt(m *Message, b []byte) ([]byte, error) {
 		if extraPadding {
 			paddingBytes = 2
 		}
-		paddingLength := plaintextBlockSize - ((len(b[headerLength:]) + c.algo.SignatureLength() + paddingBytes) % plaintextBlockSize)
+		// The PaddingSize byte(s) and Padding must fill the plaintext up to a
+		// block boundary. If it is already block-aligned no padding is needed
+		// beyond the PaddingSize byte(s) themselves (cf. open62541
+		// ua_securechannel_crypto.c, padding calculation).
+		remainder := (len(b[headerLength:]) + c.algo.SignatureLength() + paddingBytes) % plaintextBlockSize
+		paddingLength := 0
+		if remainder != 0 {
+			paddingLength = plaintextBlockSize - remainder
+		}
 
+		// appends paddingLength Padding bytes plus one PaddingSize byte,
+		// each holding the paddingLength value (OPC UA Part 6, 6.7.2.5)
 		for i := 0; i <= paddingLength; i++ {
 			b = append(b, byte(paddingLength))
 		}
--- a/uasc/secure_channel_instance_test.go
+++ b/uasc/secure_channel_instance_test.go
@@ -1,0 +1,140 @@
+package uasc
+
+import (
+	"bytes"
+	"strconv"
+	"testing"
+
+	"github.com/gopcua/opcua/ua"
+	"github.com/gopcua/opcua/uapolicy"
+	"github.com/stretchr/testify/require"
+)
+
+func symmetricInstance(t *testing.T, uri string) *channelInstance {
+	t.Helper()
+
+	nonce := bytes.Repeat([]byte{0x5a}, 32)
+	algo, err := uapolicy.Symmetric(uri, nonce, nonce)
+	require.NoError(t, err, "error: create symmetric algorithm")
+
+	return &channelInstance{
+		sc:   &SecureChannel{cfg: &Config{SecurityMode: ua.MessageSecurityModeSignAndEncrypt}},
+		algo: algo,
+	}
+}
+
+// symmetricChunk returns the raw bytes of a MSG chunk (header, symmetric
+// security header, sequence header, body) and the matching Message needed
+// by signAndEncrypt.
+func symmetricChunk(body []byte) ([]byte, *Message) {
+	m := &Message{
+		MessageHeader: &MessageHeader{
+			Header:                  NewHeader(MessageTypeMessage, ChunkTypeFinal, 1),
+			SymmetricSecurityHeader: NewSymmetricSecurityHeader(1),
+			SequenceHeader:          NewSequenceHeader(1, 1),
+		},
+	}
+
+	b := make([]byte, 0, 24+len(body))
+	b = append(b, 'M', 'S', 'G', 'F')
+	b = append(b, 0, 0, 0, 0) // message size, set by signAndEncrypt
+	b = append(b, 1, 0, 0, 0) // secure channel id
+	b = append(b, 1, 0, 0, 0) // security token id
+	b = append(b, 1, 0, 0, 0) // sequence number
+	b = append(b, 1, 0, 0, 0) // request id
+	b = append(b, body...)
+	return b, m
+}
+
+var symmetricPolicyURIs = []string{
+	ua.SecurityPolicyURIBasic128Rsa15,
+	ua.SecurityPolicyURIBasic256,
+	ua.SecurityPolicyURIBasic256Sha256,
+	ua.SecurityPolicyURIAes128Sha256RsaOaep,
+	ua.SecurityPolicyURIAes256Sha256RsaPss,
+}
+
+// TestMaxBodySizeFitsChunk reproduces the bug from
+// https://github.com/gopcua/opcua/issues/781: a body of exactly maxBodySize
+// must produce an encrypted chunk that does not exceed the negotiated chunk
+// size, for every security policy and in particular for the default chunk
+// size of 65535 which is not a multiple of the AES block size.
+func TestMaxBodySizeFitsChunk(t *testing.T) {
+	chunkSizes := []int{4096, 8192, 65521, 65535, 65536}
+
+	for _, uri := range symmetricPolicyURIs {
+		for _, chunkSize := range chunkSizes {
+			t.Run(testName(uri, chunkSize), func(t *testing.T) {
+				c := symmetricInstance(t, uri)
+				c.SetMaximumBodySize(chunkSize)
+
+				body := bytes.Repeat([]byte{0xab}, int(c.maxBodySize))
+				b, m := symmetricChunk(body)
+				want := make([]byte, len(b)-16)
+				copy(want, b[16:])
+
+				cipher, err := c.signAndEncrypt(m, b)
+				require.NoError(t, err, "error: message encrypt")
+				require.LessOrEqual(t, len(cipher), chunkSize, "encrypted chunk exceeds negotiated chunk size")
+
+				// a body one byte larger must no longer fit, i.e. maxBodySize
+				// is not needlessly conservative
+				b2, m2 := symmetricChunk(append(body, 0xab))
+				cipher2, err := c.signAndEncrypt(m2, b2)
+				require.NoError(t, err, "error: message encrypt")
+				require.Greater(t, len(cipher2), chunkSize, "maxBodySize is more conservative than necessary")
+
+				// round trip
+				mc := new(MessageChunk)
+				_, err = mc.Decode(cipher)
+				require.NoError(t, err, "error: message decode")
+
+				plain, err := c.verifyAndDecrypt(mc, cipher)
+				require.NoError(t, err, "error: message decrypt")
+				require.Equal(t, want, plain, "plaintext not equal")
+			})
+		}
+	}
+}
+
+// TestPaddingBlockAligned reproduces the second bug discussed in
+// https://github.com/gopcua/opcua/pull/824: when the plaintext including the
+// PaddingSize byte is already block-aligned, no padding bytes must be added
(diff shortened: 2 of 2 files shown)


jminardi:
I've rebuilt the telergaf docker container with this branch of gopcua and pushed it here in case anyone wants to try it to see if it fixes their problems: ghcr.io/jminardi/telegraf:fix-781

jminardi:
@reindlt confirmed that this fix solved the oversize message chunk issue in their use case

jminardi:
@magiconair or @kung-foo is there anything either of you would like to see in this PR?

I previously was only using a simulated opcua server (using node-opcua), but I recently tested this against kepware and it fixes the issue there as well.

jminardi:
@JeremyTheocharis I know you might be busy getting situated as a new maintainer, but if you had a chance to review this PR it would be helpful.

JeremyTheocharis:
let me check it out :)

JeremyTheocharis:
i am still looking at it, it is not trivial. maybe some thoughts already:
1. i have compared reference implementations. .NET seems to have the same behjaviour as gopcua, but open62541 seems to have the correct behaviour as far as i did in my first analysis
2. i think this PR will probably fix the bug, but I need to understand first whether gopcua is not spec comliant in that regard

does anyone have some thoughts on this?

jminardi:
Hi @JeremyTheocharis thanks for digging into this! I just spent some time comparing the OPC UA spec as written to the `gopcua` implementation here. Looking at this with fresh eyes after a few weeks gave me another idea. Here is the relevant definition and implementation:

OPC UA spec:
https://reference.opcfoundation.org/Core/Part6/v104/docs/6.7.2.5
```
MaxBodySize = PlainTextBlockSize * Floor(
    (MessageChunkSize – HeaderSize - 1)/CipherTextBlockSize
  ) – SequenceHeaderSize – SignatureSize;
```

Current `gopcua` implementation:
```go
maxBodySize := c.algo.PlaintextBlockSize()*
  ((chunkSize-headerSize-symmetricAlgorithmHeader-c.algo.SignatureLength()-1)/c.algo.BlockSize()) -
  sequenceHeaderSize
```

Comparing them directly, here is how they match up:

| OPC UA Spec               | `gopcua` implementation                                                                          |
|-----------------------|----------------------------------------------------------------------|
| PlainTextBlockSize      | `c.algo.PlaintextBlockSize()`                                                                      |
| Floor()                          | go's integer division                                                                                     |
| MessageChunkSize    | `chunkSize`                                                                                                  |
| HeaderSize - 1             | `headerSize-symmetricAlgorithmHeader-c.algo.SignatureLength()-1` |
| CipherTextBlockSize   | `c.algo.BlockSize()`                                                                                     | 
| SequenceHeaderSize | `sequenceHeaderSize `                                                                              | 
| SignatureSize              | _missing from implementation?_                                                                | 

The two main differences are the HeaderSize and SignatureSize.

The OPC UA spec says: "The HeaderSize includes the MessageHeader and the SecurityHeader." 

When I initially looked at this code and the spec and made this PR, I had assumed the "SecurityHeader" mentioned above was `symmetricAlgorithmHeader` and `c.algo.SignatureLength()`, plus my new addition of `signaturePadding`.

However I now believe a different (better?) interpretation is that "SecurityHeader" is just `symmetricAlgorithmHeader`, and `c.algo.SignatureLength()` is in the wrong place. The fix would be to move `c.algo.SignatureLength()` out from the numerator and use it instead where the spec mentions "SignatureSize" at the end of the formula. This would bring the `gopcua` implementation more in line with the formula defined in the spec. Here is what I am thinking:

```go
maxBodySize := c.algo.PlaintextBlockSize() *
  ((chunkSize-headerSize-symmetricAlgorithmHeader-1)/c.algo.BlockSize()) -
  sequenceHeaderSize - c.algo.SignatureLength()
```

This change would result in a smaller `maxBodySize` which I think would work. Does this seem reasonable as an alternative approach? How does it compare to the other implementations you've been researching?

JeremyTheocharis:
I haven't forgotten you, just being overwhelmed here on our side :)

jminardi:
No worries, I know this is a tough area to fix. The "good" news is that its completely broken and not working now so anything we do can't make it worse 😉

Any initial thoughts on my proposal above? I can try actually implementing it to confirm it works if you think it'll possibly be merged.

JeremyTheocharis:
Hi @jminardi !

Sorry for taking so long to review this. It took me over the past 2 months around 5-7h. It is currently my personal highlight on "time spend compared to lines of code".

I think the proposal you are doing is good. There is a second bug though, let me use the words of my LLM:

The padding logic (lines 180-187) incorrectly calculates `paddingLength` when data is already block-aligned:
- Current: `paddingLength = plaintextBlockSize - (remainder % plaintextBlockSize)`
- When remainder = 0: gives `paddingLength = 16`, loop adds 17 bytes
- Should give: `paddingLength = 0`, loop adds 1 byte (just the size byte)

**Reference** (open62541 correct implementation):
https://github.com/open62541/open62541/blob/master/src/ua_securechannel_crypto.c#L260-281

The fix is to add an explicit zero-check:

```go
remainder := (len(b[headerLength:]) + c.algo.SignatureLength() + paddingBytes) % plaintextBlockSize
paddingLength := 0
if remainder != 0 {
    paddingLength = plaintextBlockSize - remainder
}
```

Is there a way on how we can reproduce the (two) bugs in a test?

JeremyTheocharis:
Is it something you still want to work on? Otherwise, I can proceed to make the necessary changes but also don't want to "steal" your idea

reindlt:
Hi, as I stumbled upon this issue again what needs to be done to finish this off? I think we are close to a solution here.

JeremyTheocharis:
> Hi, as I stumbled upon this issue again what needs to be done to finish this off? I think we are close to a solution here.

https://github.com/gopcua/opcua/pull/824#issuecomment-3737371952 see also my previous comment there

jminardi:
Hi @JeremyTheocharis sorry for the long delay here. I've just been running the code as its been working for me, but I wanted to circle back and try to get the actually correct fix merged in.

Two parts to it:

For the padding fix I used your zero-remainder check exactly as you wrote it. The old code added a full extra block (17 bytes) when the plaintext was already block-aligned, and your fix drops that to the one size byte. 

For the maxBodySize calc, I started with the approach I proposed above (moving `SignatureSize` out to the end of the formula to line it up with the spec), but when I wrote a test for it I found it was still 1 byte too big at the default chunk size.

The culprit is the `- 1` (the PaddingSize byte) that the spec keeps *inside* the Floor:

```
MaxBodySize = PlainTextBlockSize * Floor((MessageChunkSize - HeaderSize - 1)/CipherTextBlockSize) - SequenceHeaderSize - SignatureSize
```

That `-1` only actually buys you anything when `MessageChunkSize - HeaderSize` happens to land exactly on a cipher-block boundary. For any other chunk size (the default 65535 isn't a multiple of 16) a full-size body still pads out to one block past what fits in the chunk, so it goes over on the wire by a byte. This lines up with what you found about open62541 being the correct one: it reserves the PaddingSize byte outside the Floor, and that turns out to be exact for every chunk size. So the formula I landed on is:

```go
maxBodySize := c.algo.PlaintextBlockSize() *
    ((chunkSize - headerSize - symmetricAlgorithmHeader) / c.algo.BlockSize()) -
    sequenceHeaderSize - c.algo.SignatureLength() - paddingSizeBytes
```

(`paddingSizeBytes` is 1, or 2 for signatures longer than 256 bytes.)

To answer your question about getting both bugs into a test: yes. I added `TestMaxBodySizeFitsChunk`, which encrypts a body of exactly `maxBodySize` and asserts the encrypted chunk fits, and that one byte more does *not* fit (so we're not being needlessly conservative), across all the symmetric policies at a handful of chunk sizes including 65535. And `TestPaddingBlockAligned` covers the padding case.

Hopefully this is good to merge now, let me know if there is anything else you'd like to see.

JeremyTheocharis (approved):
Thanks @jminardi, and thanks @reindlt for field-testing this on a real encrypted setup.

I went through the padding math against OPC UA Part 6 v1.05 §6.7.2.5 and cross-checked it line by line against the current open62541 and OPC Foundation .NET reference implementations. It's correct and matches both.

The root cause of #781 is in `SetMaximumBodySize`: it subtracts the signature and the PaddingSize byte *inside* the Floor division, which only reserves the space correctly when the chunk size is an exact multiple of the cipher block size. For any other size, including the default 65535, a maximum-size body comes out up to a block too big. `signAndEncrypt` adds a full extra block of padding even when the plaintext is already block-aligned, making the chunk larger still. The result is a chunk over the negotiated MaxChunkSize, which is the EOF/timeout people hit past ~1000 nodes on an encrypted channel.

The fix is correct for two reasons. First, it floors the cipher region to whole blocks before reserving signature + PaddingSize byte(s) outside the Floor; that matches open62541 `setBufPos` and .NET `WriteSymmetricMessage`, which reserve only the size byte(s) rather than a full block because the whole-block floor already absorbs the variable padding. Second, the zero-check (`paddingLength = remainder == 0 ? 0 : blockSize - remainder`) gives an already-aligned plaintext no extra block, the same as open62541 `padChunk` and .NET's guarded padding. Worst-case overhead is now one block, not one block plus one byte.

I verified it two ways:
- `TestPaddingBlockAligned` and `TestMaxBodySizeFitsChunk` fail on `main` (the block-aligned case comes out 16 bytes too large) and pass with this change. The existing asymmetric round-trip test still exercises the 2-byte ExtraPaddingSize path for >2048-bit keys.
- End to end against the OPC Foundation .NET reference server: a SignAndEncrypt read of 1455 nodes fails on `main` (the server drops the oversized chunk without an error, so the client times out) and succeeds with this change.

This resolves #781. I'll reopen that issue and add the `Fixes #781` reference to the PR description so the merge links and closes it.

PR #825, opened 2025-09-09
set openSecureChannelRequest AuthenticationToken to nil on server side

nraake:
Some opc client libraries don't send the openSecureChannelRequest with a nil AuthenticationToken (e.g. open62541, which is used by Qt OPC UA: https://github.com/open62541/open62541/blob/10b66c8176cbd8d181931c7fe703038b887b6218/src/client/ua_client_connect.c#L608) With the current implementation, the client-server connection will be interrupted after 75% of the requested channel lifetime, leading to crashes or loss of data. This change improves compatibility of the gopcua server with applications building on Qt or open62541.

### DIFF
--- a/uasc/secure_channel.go
+++ b/uasc/secure_channel.go
@@ -740,9 +740,7 @@ func (s *SecureChannel) handleOpenSecureChannelRequest(reqID uint32, svc ua.Requ
 	}
 
 	// Part 6.7.4: The AuthenticationToken should be nil. ???
-	if req.RequestHeader.AuthenticationToken.IntID() != 0 {
-		return ua.StatusBadSecureChannelTokenUnknown
-	}
+	req.RequestHeader.AuthenticationToken = nil
 
 	s.cfg.Lifetime = req.RequestedLifetime
 	s.cfg.SecurityMode = req.SecurityMode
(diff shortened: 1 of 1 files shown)


jminardi:
The comment leads here: https://reference.opcfoundation.org/Core/Part6/v104/docs/6.7.4 which does indicate that "The AuthenticationToken in the RequestHeader shall be set to null." Doesn't that mean this is technically a bug in the `open62541` library since its not following the spec?

That being said I do agree its good to be liberal in what you accept, so this may make sense to change. It just feels odd to adjust a client's request like this before processing it. Did you try just removing the early return and continuing the processing with a non-null AuthenticationToken?

nraake:
@jminardi regarding your remarks:

- yes, the fact that the `open62541` client sends a non-null auth token on OpenSecureChannel requests is a deviation from the specifications of the opc ua protocol, so it's technically a bug.
- you're absolutely right that it seems wrong to change the request
- I have tested removing the check/early return, and it works just as well

I will change the pr accordingly. Should we maybe raise a debug level log message? Or just gloss over the token completely?

jminardi:
@nraake I think you have a bug in your most recent commit (`==` should be `!=`). But to minimize changes with the existing code your changes should just be this:

```
	// Part 6.7.4: The AuthenticationToken should be nil. ???
	if req.RequestHeader.AuthenticationToken.IntID() != 0 {
		debug.Printf("Received non-nil AuthenticationToken on OpenSecureChannelRequest")
	}
```

This becomes a one line change to print a debug message instead of return an error.

PR #839, opened 2025-12-02
Fix: Republish loop doesn't advance sequence numbers

mvandergrift:
### Problem
The `sendRepublishRequests` function was requesting republished messages but  never advancing `sub.nextSeq` or `sub.lastSeq`. This causes an infinite loop requesting the same sequence number repeatedly. In addition, republished data was never being sent to `notifySubscription`.

### Root Cause
After successfully receiving a `RepublishResponse`, the code checked the status  but didn't process the notification or update sequence tracking, so the next  iteration would request the same sequence again. `StatusBadMessageNotAvailable` would never be returned from the server, so the loop would never exit.

### Solution
- Process the republished notification via `notifySubscription`
- Advance `sub.lastSeq` and `sub.nextSeq` to track progress
- Add failsafe check when sequence is no longer in `availableSeq` buffer. This avoids a situation where the server fails to return `StatusBadMessageNotAvailable` on a call for an out-of-bounds `retransmitSequenceNumber`

### Testing
This bug manifests deep in subscription state management during reconnection  scenarios. While end-to-end testing would require complex OPC UA server  infrastructure, the logic fix is straightforward: we must advance the sequence counter after successful republish.

### DIFF
--- a/client_sub.go
+++ b/client_sub.go
@@ -206,7 +206,21 @@ func (c *Client) sendRepublishRequests(ctx context.Context, sub *Subscription, a
 				debug.Printf("Republishing subscription %d failed: %v", req.SubscriptionID, status)
 				return status
 			}
+
+			// Process the republished notification and advance sequence number
+			if res.NotificationMessage != nil {
+				c.notifySubscription(ctx, sub, res.NotificationMessage)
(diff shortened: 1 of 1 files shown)


JeremyTheocharis (approved):

JeremyTheocharis:
<3

Issue #841, opened 2025-12-03
Server crashes on write requests to non-existing namespace


semihbkgr:
When a client sends a **WriteRequest** targeting a node in a **non-existing namespace**, the OPC UA server crashes with a `nil pointer dereference` panic.

## Expected behavior
I am **not an expert on OPC UA internals**, but I believe the expected behavior in this case should be:
- The server returns a meaningful OPC UA error code such as:
  - `Bad_NodeIdUnknown`
  - `Bad_NodeNotInView`
- The server should *not crash* under any circumstances when the client sends an invalid namespace.

This is consistent with how the server already behaves when the client writes to a non-existing node **inside an existing namespace**.

## Observed behavior
The server panics:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x58]

github.com/gopcua/opcua/server.(*AttributeService).Write(...)
    server/attribute_service.go:89
```

On the client side, the error becomes:

```
Write failed: EOF
```

This occurs when writing to a node such as:

```
ns=132;i=101
```

(where namespace 132 does not exist).

## Steps to reproduce
1. Start a server using gopcua/opcua.
2. Send a WriteRequest to a node with a non-existing namespace, e.g.:  
   `ns=132;i=101`
3. The server immediately crashes.

## Cause
In `AttributeService.Write`:

```go
ns, err := s.srv.Namespace(int(n.NodeID.Namespace()))
if err != nil {
    status[i] = ua.StatusBadNodeNotInView
}
status[i] = ns.SetAttribute(n.NodeID, n.AttributeID, n.Value)
```

If `Namespace()` returns an error, then `ns == nil`, but the code still calls:

```go
ns.SetAttribute(...)
```

→ causing a nil pointer dereference and server crash.

Note: The `Read` implementation *does* handle missing namespaces correctly by using `continue`, but `Write` does not.

## Suggested fix
Add a `continue` when namespace lookup fails:

```go
ns, err := s.srv.Namespace(int(n.NodeID.Namespace()))
if err != nil || ns == nil {
    status[i] = ua.StatusBadNodeNotInView
    continue
}
status[i] = ns.SetAttribute(n.NodeID, n.AttributeID, n.Value)
```

## Additional notes
- Writing to a non-existing **node inside an existing namespace** works correctly and produces:
  ```
  BadNodeIDUnknown (0x80340000)
  ```
[...]

PR #842, opened 2025-12-03
Fix server crash on Write requests to non-existing namespace

semihbkgr:
Fix crash in `AttributeService.Write` when handling Write requests targeting a non-existing namespace.

See: #841


istyf (approved):
/lgtm

Fixes #841

JeremyTheocharis (approved):

PR #844, opened 2025-12-05
do not return bad attribute id when setting attrs that are not the value attribute

### LINKED ISSUE #843: SetAttribute on server nodes returns bad attribute error for all attributes ids except attributeidvalue
In Node's SetAttribute only the case for AttributeIDValue returns a nil error, the default case returns a bad attribute error after updating the attributes map.

Error here: https://github.com/gopcua/opcua/blob/0d83d1cc736a5db59bfbb31a167d3e9c6b09c845/server/node.go#L216:L218

istyf:
Fixes #843

### DIFF
--- a/server/node.go
+++ b/server/node.go
@@ -201,21 +201,22 @@ func (n *Node) Attribute(id ua.AttributeID) (*AttrValue, error) {
 		return nil, ua.StatusBadAttributeIDInvalid
 	}
 }
+
 func (n *Node) SetAttribute(id ua.AttributeID, val *ua.DataValue) error {
-	switch {
-	case id == ua.AttributeIDValue:
+
+	switch id {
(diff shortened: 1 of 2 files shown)


istyf:
@JeremyTheocharis Can you please take a look at this?

JeremyTheocharis (approved):
Reviewed as maintainer. The fix correctly stops `Node.SetAttribute` from reporting a Bad status after it has already stored a non-Value attribute. Verified against OPC UA Part 4 §5.10.4 and the open62541 and OPC Foundation .NET reference servers: a permitted non-Value attribute write returns Good. Added an integration regression that is red on the previous behavior (Bad_AttributeIdInvalid) and green with this change. Thanks @istyf.

PR #847, opened 2025-12-10
allow Node to work on a client interface instead of a concrete implementation

### LINKED ISSUE #846: Is the Client.Node function impossible to mock for testing?
## Problem Statement

If I'm not mistaken, given

- `Node()` is a member of `Client`
- the `opcua.Node` type does not export `c`
 
https://github.com/gopcua/opcua/blob/0d83d1cc736a5db59bfbb31a167d3e9c6b09c845/client.go#L991-L993

When attempting to mock the `Client` with something like

```go
type OpcuaClient interface {
    Node(id *ua.NodeID) *opcua.Node
    // ... other functions ...
}
```

if the `client.Node()` function must be called, `c` can never be set properly because the mocked function will not mimic the original's behavior. This becomes a problem when `node.Attributes()` is called, which eventually leads to a panic during a call to `client.Send()` where the `client` ends up being `nil` due to an incomplete mocked return value.

```go
mockClient.EXPECT().Node(gomock.Any()).Return(&opcua.Node{ID: ua.MustParseNodeID("s=test")})
```

## Possible Solution (not tested)

Export `C` from `opcua.Node` so a mock could be realized where the mocked client could be reused.

#### Using `go.uber.org/mock/gomock`
```go
mockClient.EXPECT().Node(gomock.Any()).Return(&opcua.Node{ID: ua.MustParseNodeID("s=test"), C: mockClient})
```

#### Using `github.com/vektra/mockery` with `github.com/stretchr/testify/mock`
```go
mockClient.On("Node", mock.Anything).Return(&opcua.Node{ID: ua.MustParseNodeID("s=test"), C: mockClient})
```

##  ❓ 

Is there another way to do this? Any other suggestions are welcome.

istyf:
This change improves the testability of Node by allowing the injected client to be a mock/fake/stub/whateveryouneed.

Fixes #846

### DIFF
--- a/client.go
+++ b/client.go
@@ -122,6 +122,16 @@ func (a bySecurityLevel) Len() int           { return len(a) }
 func (a bySecurityLevel) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
 func (a bySecurityLevel) Less(i, j int) bool { return a[i].SecurityLevel < a[j].SecurityLevel }
 
+type ClientInterface interface {
+	Browse(context.Context, *ua.BrowseRequest) (*ua.BrowseResponse, error)
+	BrowseNext(context.Context, *ua.BrowseNextRequest) (*ua.BrowseNextResponse, error)
+
+	NodeFromExpandedNodeID(*ua.ExpandedNodeID) *Node
+
+	Read(context.Context, *ua.ReadRequest) (*ua.ReadResponse, error)
+	Send(context.Context, ua.Request, func(ua.Response) error) error
+}
+
 // Client is a high-level client for an OPC/UA server.
 // It establishes a secure channel and a session.
 type Client struct {
--- a/client.go
+++ b/client.go
@@ -126,10 +126,16 @@ type ClientInterface interface {
 	Browse(context.Context, *ua.BrowseRequest) (*ua.BrowseResponse, error)
 	BrowseNext(context.Context, *ua.BrowseNextRequest) (*ua.BrowseNextResponse, error)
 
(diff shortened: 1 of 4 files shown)


istyf:
Made the same type of changes to Subscription.

JeremyTheocharis (commented):
I like the direction. Making `Node` and `Subscription` testable without a real connection is worth doing. One thing before merging.

**Blocker: `RegisterSubscription_NeedsSubMuxLock` exports an anti-pattern twice over: the `_NeedsSubMuxLock` naming, plus a public method that requires a lock external callers don't hold.**

Two fixes, your pick:

**(A) Keep it private, off the interface.** Restore `registerSubscription_NeedsSubMuxLock` (the name it had before this PR), remove from `ClientInterface`, and restructure `Subscription.recreate_create` so the orchestrator in `Client.recreateSubscription` calls the helper directly on `*Client`.

**(B) Add a wrapper that locks.** Keep `registerSubscription_NeedsSubMuxLock` unexported, add an exported `RegisterSubscription` on `ClientInterface` that takes `subMux.Lock()` internally and delegates. Heads up: `recreateSubscription` at `client_sub.go:93` currently holds `subMux.Lock()` across the whole function, so routing through the wrapper would deadlock. You'd need to drop the outer lock too.

istyf:
Option A clearly seems like the better option here. I have removed RegisterSubscription_NeedsSubMuxLock from the client interface and unexpected it again. Please have another look.

JeremyTheocharis (approved):

PR #853, opened 2026-03-18
fix: handle certificate chains returned by OPC UA servers

diericd:
Handle certificate chains returned by OPC UA servers.

Some servers return their `ServerCertificate` as an X.509 chain (leaf + intermediate(s)) rather than a single DER certificate. Parsing assumed a single cert, so connecting to a chain-returning server failed. This parses the chain and uses the leaf certificate for the thumbprint and public key.

The `Sign`-mode handshake fix originally bundled here landed separately in #866; this PR is now scoped to the certificate-chain change.

Adds tests for chain, single-cert, malformed, and empty certificate inputs.


JeremyTheocharis:
would it be possible to add a test that uses the wincc certificate that was failing for you? you could create a new folder called testdata/certs and then put your cert in there.

this would make it easier for me in reviewing :)

as soon i have that, i would take a closer look into the code changes you did.

diericd:
Hi @JeremyTheocharis, I've added some tests that shoud fail without the proposed fixes and the certificate presented by the wincc server.

diericd:
ping?

JeremyTheocharis:
I have found a regression in the existing codebase, so i am currently fixing it, will then rebase your PR and add a slight modification and an additional test to it and then merge it. Nothing to do from your side.

JeremyTheocharis:
Thanks @diericd — and thanks for diagnosing the `Sign`-mode break along the way.

I've landed the complete §6.7.4 OPN-encryption fix separately in #866 (now merged), so I've taken this PR down to just its certificate-chain change, which is the valuable part. With maintainer-edit enabled I rebased onto `main`, dropped the `SecurityMode` override commit (superseded by #866), and added a few tests covering single-cert, malformed, and empty inputs alongside your chain tests. Your commits are otherwise unchanged.

The fix is correct: a server may return its certificate as a chain (leaf + intermediates), and we now parse the chain and use the leaf for the thumbprint and public key. Verified against the chain test data and the full crypto suite. Merging for v0.9.0 — thanks for the contribution.

PR #855, opened 2026-04-13
fix restoreSession closing the session it just re-activated

### LINKED ISSUE #854: restoreSession closes the session it just re-activated, deleting all subscriptions
**Description**

  When a connection drops and the auto-reconnect logic takes the restoreSession path, the session is activated successfully on the server but then immediately closed with DeleteSubscriptions: true. This makes subscription
  transfer/restore impossible, leaving the client unable to recover without a full restart.

  **Root Cause**

  ActivateSession() has an internal callback that always calls c.CloseSession(ctx) before setting the new session (client.go:921). This is correct for the recreateSession path where old and new sessions differ, but in
  restoreSession the "previous" and "new" session are the same object.

  The sequence in restoreSession:
  1. s := c.Session() - gets existing session with active subscriptions
  2. c.ActivateSession(ctx, s) - re-activates it on the server
  3. Inside the callback, c.CloseSession(ctx) is called - this sends CloseSessionRequest{DeleteSubscriptions: true} for s (the session we just restored)
  4. c.setSession(s) - sets the now-closed session back on the client
  5. restoreSubscriptions fails because subscriptions were deleted server-side

  **Prior Fix**

  This was partially fixed for the recreateSession path in #700 by adding c.setSession(nil) before calling ActivateSession, which causes the CloseSession call inside the callback to no-op (since closeSession(nil) returns early). The restoreSession path was not updated with the same fix.

  **Fix**

Add c.setSession(nil) before the ActivateSession call in the restoreSession action, matching what recreateSession already does. The session object is still held by the local variable s and gets re-set on the client inside the ActivateSession callback via c.setSession(s).

skartikey:
During reconnect, the restoreSession path calls ActivateSession with the existing session. ActivateSession's internal callback always calls CloseSession before setting the new session. Since the old and new session are the same object, this sends CloseSessionRequest with DeleteSubscriptions: true, destroying all subscriptions on the server.

Clear the session reference before calling ActivateSession, matching the fix applied to recreateSession in #700. The session object is still held by the local variable and gets re-set inside the callback.

Fixes #854

### DIFF
--- a/client.go
+++ b/client.go
@@ -430,6 +430,10 @@ func (c *Client) monitor(ctx context.Context) {
 							continue
 						}
 
+						// clear the session from the client to prevent
+						// ActivateSession from closing it via CloseSession
+						c.setSession(nil)
+
 						dlog.Printf("trying to restore session")
 						if err := c.ActivateSession(ctx, s); err != nil {
 							dlog.Printf("restore session failed: %v", err)
--- a/client.go
+++ b/client.go
@@ -445,6 +445,10 @@ func (c *Client) monitor(ctx context.Context) {
 							continue
 						}
 
+						// clear the session from the client to prevent
+						// ActivateSession from closing it via CloseSession
+						c.setSession(nil)
+
 						dlog.Printf("trying to restore session")
 						if err := c.ActivateSession(ctx, s); err != nil {
 							dlog.Printf("restore session failed: %v", err)


skartikey:
Adding context from a production report. A user running Telegraf 1.38.2 with inputs.opcua_listener against a Siemens OpenPCS7 V9.2 server hit this exact bug. The connection dropped after ~25 minutes of low data-change activity, and the auto-reconnect sequence played out exactly as described in #854:                                                                                                                   
                                                                                                
  restoreSession -> ActivateSession (success) -> CloseSession (!!!) -> "session restored"                                                                                                                          
  -> ReadRequest -> StatusBadSessionIDInvalid                                                                                                                                                                      
  -> recreateSession (success) -> TransferSubscriptions -> ERRF/EOF                                                                                                                                                
  -> restoreSubscriptions -> broken pipe                                                                                                                                                                           
  -> "no subscriptions to resume"                                                                                                                                                                                  
                                                                                                                                                                                                                   
  After this, the client was permanently disconnected with no further reconnection attempts. The fix here matches what #700 did for recreateSession and is the minimal correct change.                             
                  
  One question: should `closeSession(nil)` also be guarded with a debug log (e.g., `dlog.Printf("skipping close of nil session")`) to make future troubleshooting easier, or is the current silent return-early sufficient?

JeremyTheocharis:
<3

Thank you! I have no opinion on the debug log. Good report by the way, allowed me to easily review it.

JeremyTheocharis (approved):

skartikey:
@JeremyTheocharis May I know the plan for merging this PR? Are we waiting for a second review? Happy to rebase it onto master as well since the branch looks out of date.

PR #856, opened 2026-04-20
fix(monitor): return partial results from AddMonitorItems on per-node failures

sruehl:
## Problem

`AddMonitorItems` returns an error on the first node with a non-OK status code, discarding all successfully created monitored items in the same batch. This means a single stale or invalid node ID (e.g. `StatusBadNodeIDUnknown`) blocks monitoring of all other valid nodes in the chunk.

This is particularly problematic in production environments where:
- Devices are reconfigured and some node IDs become stale
- Large batches of nodes are monitored in chunks
- The all-or-nothing behavior causes cascading failures across the entire subscription

## Fix

`AddMonitorItems` now:
1. Registers all successful items in `itemLookup` and `handles`
2. Cleans up handles for failed items
3. Returns successfully created items alongside `errors.Join` that lists per-node failures

## New types

```go
// ItemError represents a single monitored item that failed to be created.
type ItemError struct {
    NodeID *ua.NodeID
    Status ua.StatusCode
}
```

## Usage

```go
	items, err := sub.AddMonitorItems(ctx, requests...)
	if err != nil {
	    // items contains the successfully created monitors (nil if all failed)

	    // Check for a specific status code across all failures:
	    if errors.Is(err, ua.StatusBadNodeIDUnknown) { ... }

	    // Inspect the first failing item:
	    var ie *monitor.ItemError
	    if errors.As(err, &ie) { _ = ie.NodeID; _ = ie.Status }

	    // Iterate over all failures:
	    if u, ok := err.(interface{ Unwrap() []error }); ok {
	        for _, e := range u.Unwrap() {
	            var ie *monitor.ItemError
	            if errors.As(e, &ie) { /* ie.NodeID, ie.Status */ }
	        }
	    }
	}
```

## Backward compatibility

This is backward compatible:
- Callers that only check `err != nil` still see the error on partial failure
- Callers that inspect the returned `[]Item` will now get the successfully created ones
- Total failures (service-level errors, transport errors) still return `nil, err` as before

### DIFF
--- a/monitor/subscription.go
+++ b/monitor/subscription.go
@@ -2,6 +2,7 @@ package monitor
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"sync/atomic"
 
@@ -17,6 +18,45 @@ var (
 	// ErrSlowConsumer is returned when a subscriber does not keep up with the incoming messages
 	ErrSlowConsumer = errors.New("slow consumer. messages may be dropped")
 )
+
+// ItemError represents a single monitored item that failed to be created.
+// AddMonitorItems returns an errors.Join of *ItemError when one or more items
+// could not be created on the server. Callers can inspect failures with the
+// standard errors helpers:
+//
+//	items, err := sub.AddMonitorItems(ctx, requests...)
+//	if err != nil {
+//	    // items contains the successfully created monitors (nil if all failed)
+//
+//	    // Check for a specific status code across all failures:
+//	    if errors.Is(err, ua.StatusBadNodeIDUnknown) { ... }
+//
+//	    // Inspect the first failing item:
+//	    var ie *monitor.ItemError
+//	    if errors.As(err, &ie) { _ = ie.NodeID; _ = ie.Status }
+//
+//	    // Iterate over all failures:
+//	    if u, ok := err.(interface{ Unwrap() []error }); ok {
+//	        for _, e := range u.Unwrap() {
+//	            var ie *monitor.ItemError
+//	            if errors.As(e, &ie) { /* ie.NodeID, ie.Status */ }
+//	        }
+//	    }
+//	}
+type ItemError struct {
+	NodeID *ua.NodeID
+	Status ua.StatusCode
+}
+
+func (e *ItemError) Error() string {
+	return fmt.Sprintf("%s: %s", e.NodeID, e.Status)
+}
+
+// Unwrap exposes the underlying status code so that errors.Is can match on
+// values such as ua.StatusBadNodeIDUnknown.
+func (e *ItemError) Unwrap() error {
+	return e.Status
+}
 
 // ErrHandler is a function that is called when there is an out of band issue with delivery
 type ErrHandler func(*opcua.Client, *Subscription, error)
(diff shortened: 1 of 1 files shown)


sruehl:
@JeremyTheocharis / @kung-foo can you please have a look and merge this?

JeremyTheocharis:
Currently looking at it

monitor/subscription.go
```diff
@@ -18,6 +20,37 @@ var (
 	ErrSlowConsumer = errors.New("slow consumer. messages may be dropped")
 )
 
+// ItemError represents a single monitored item that failed to be created.
+type ItemError struct {
+	NodeID *ua.NodeID
+	Status ua.StatusCode
+}
+
+// PartialError is returned by AddMonitorItems when some (but not all) items
```
led0nk: This here needs adjustment, or better the guard of `AddMonitorItems`

It only checks on `len(failedItems) > 0`, which would also take place if everything fails. Personally i'd prefer an adjustment of the guard there.

monitor/subscription.go
```diff
@@ -18,6 +20,37 @@ var (
...
+//	var pe *monitor.PartialError
+//	if errors.As(err, &pe) {
+//	    // items contains the successfully created monitors
+//	    // pe.Items contains the failures with per-node status codes
+//	}
+type PartialError struct {
+	Items []ItemError
+}
+
+func (e *PartialError) Error() string {
```
led0nk: I'd also suggest adding a `Unwrap` function. Otherwise this would break if someone uses `errors.Is(err, ua.StatusBadNodeIDUnknown)`, which currently works.

With that in addition one could check on things like that.

led0nk (commented):
hello @sruehl , i work for umh as well and found those 2 nits here. I already spoke with @JeremyTheocharis and this is already approved. Would be great to fix them, but not particularly neccessary.

Thanks for your contribution!

monitor/subscription.go
sruehl: fixed locally

monitor/subscription.go
sruehl: what should the return error be like? I agree that the comment is misleading (if not wrong).

sruehl: would it be cleaner to just use errors.Join() instead of the PartialError?

led0nk: i think there are 2 potential designs:

either we could make the docstring correct - and we're fine 
-> on this we would need to check on the caller-side if `len(partialErrors) == len(nodeIDs)` to throw a general error

or we check in this function with the `len(nodeID/items)` and if all items basically throw an error, we could do a errors.Join with all the errors we got there to give a full error back, which includes the corresponding nodeIDs + errors.

I think for now it would be just fine to correct the docstring, but it's on your side, what you prefer

sruehl:
@led0nk I redesigned it to get rid or the PartialError wrapper and utilizing errors.Join. This way we stay compatible.

monitor/subscription.go
sruehl: I reworked that now to use error.Join

led0nk (approved):
nice, looks good

JeremyTheocharis (approved):

PR #857, opened 2026-04-20
feat: adding pre-resolver to allow the caller to translate hosts to IPs via non-DNS

shaunco:
We come across many OPC-UA servers that are only accessible via IP, but the endpoint discovery returns endpoints with hostnames (think "myserver"), despite the local DNS servers not being able to resolve "myserver".

This change adds a pre-resolver hook to allow the caller to do non-DNS resolution of server returned hostnames.


sruehl:
@JeremyTheocharis do you see any issues with merging this? Once that is on main I could ditch my fork pin.

PR #861, opened 2026-06-02
uasc: chunk large server responses across all message writes

### LINKED ISSUE #860: Clients can't browse gopcua server with too many nodes in one folder
This is somewhat correlating to #781 and #824 

When gopcua server sends BrowseResponse that has Message Size over 64KB neither UaExpert nor gopcua client can process it and end up resetting connection (connection closed on client side).

The message it seems is sent via uasc/message.go:
```
func (m *Message) Encode() ([]byte, error) {
	chunks, err := m.EncodeChunks(math.MaxUint32)
	if err != nil {
		return nil, err
	}
	return chunks[0], nil
}

func (m *Message) EncodeChunks(maxBodySize uint32) ([][]byte, error) {
```

1. message size limit is not set (i.e. MaxUint32)
2. setting limit here does not work since only first chunk is returned via chunks[0], at best MSG + C (ChunkTypeIntermediate) would be sent

istyf:
SendMsgWithContext and sendResponseWithContext were still encoding and writing server messages as a single buffer, which broke larger responses even though request sending already supported chunked writes.

Factor the chunked write path into a shared helper, use it for both server response send paths, and keep per-chunk sequence numbers/signing consistent with the existing async request flow.

Add a regression test that verifies a large ReadResponse is written as multiple UASC chunks with increasing sequence numbers and a final chunk marker.

Fixes #860

### DIFF
--- a/uasc/secure_channel.go
+++ b/uasc/secure_channel.go
@@ -813,6 +813,7 @@ func (s *SecureChannel) handleOpenSecureChannelRequest(reqID uint32, svc ua.Requ
 	if err != nil {
 		return err
 	}
+	instance.SetMaximumBodySize(int(s.c.SendBufSize()))
 
 	instance.state = channelActive // todo(fs): is this correct?
 	// s.setState(secureChannelOpen)
--- a/uasc/secure_channel.go
+++ b/uasc/secure_channel.go
@@ -11,7 +11,6 @@ import (
 	"crypto/x509"
 	"encoding/binary"
 	"io"
-	"log"
 	"math"
 	"net"
 	"strings"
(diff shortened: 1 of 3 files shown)


fuantaji:
This works. 

Tho for me Message Size for chunk messages is 4120. Could have been split into large chunks I think.

UaExpert works until Reassembled UA Message exceeds 1MB in size.

istyf:
4120? That's peculiar. Let me investigate a bit more ...  Done.

Thanks a lot for the quick feedback! 4120 pointed me to this piece of code https://github.com/gopcua/opcua/blob/main/uasc/message.go#L133:L138

It debug logs a warning about maxBodySize == 0 and falls back to a body size of 4096 (which when the message header size 24 is added ends up being exactly 4120).

I believe that the issue here is that the OpenSecureChannel request handler does not properly set the maximum body size once the symmetric runtime crypto is established. I will push a fix for that and would appreciate if you could retest with that fix.

Regarding the 1MB failure my guess is that it happens because the server currently does not support paging/continuation via BrowseNext and that UA Expert sends a max message size that the server currently ignores. I think that deserves its own issue and fix.

fuantaji:
Yep,
```
Message Type: MSG
Chunk Type: C
Message Size: 65534
```
now chunks are big

istyf:
@JeremyTheocharis Issue fixed and confirmed. Can you give this a 👍 or 👎 ?

jminardi:
Hi @istyf, nice work here. I'd independently hit this exact thing on my own server branch and landed on almost the same two changes: the `instance.SetMaximumBodySize(int(s.c.SendBufSize()))` after OpenSecureChannel is identical to yours, and I had a rougher version of the chunked response writer. Yours is cleaner than mine was, so I support merging this in.

On the 1MB / `BrowseNext` limit @fuantaji ran into, I've got continuation-point + BrowseNext support working on my server branch and plan on submitting that as a followup PR ASAP.

JeremyTheocharis:
Will look at it, but currently prioritizing doing the client fixes first (found there a regression)

JeremyTheocharis (approved):
Reviewed as maintainer. This fixes #860: the server now splits responses larger than one MessageChunk into multiple UASC chunks instead of truncating to a single chunk. Verified the server send path reuses the same maxBodySize/sign-and-encrypt path as the client (so the #781/#824 padding fix applies), per-chunk sequence numbers are monotonic, and each encrypted chunk stays within the negotiated chunk size across all symmetric policies (added a regression test on the production send path; `go test -race` clean). Cross-checked framing/sequence/size against open62541 and the OPC Foundation .NET stack. Negotiated MaxMessageSize/MaxChunkCount enforcement on send and round-trip/Sign-only test coverage are marked as TODOs for follow-up (out of scope for #860). Thanks @istyf.

PR #862, opened 2026-06-03
fix(uasc): derive ReceiverCertificateThumbprint on the server OPN path

mschm:
When a SecureChannel opens, readChunk() captures the peer's certificate
from the inbound OpenSecureChannel into cfg.RemoteCertificate but never
derives cfg.Thumbprint. Server-driven OPN responses therefore go out with
an empty ReceiverCertificateThumbprint, and strict clients (e.g. Prosys,
UaExpert) reject every Basic256Sha256 Sign / SignAndEncrypt channel with a
"certificate thumbprint mismatch".

Derive Thumbprint = SHA1(RemoteCertificate) on the shared readChunk path,
mirroring the client-side derivation in config.go (SecurityFromEndpoint).
This is gated by the existing SecurityPolicyURI != None guard plus a
defensive length check, so the None policy keeps emitting an empty
thumbprint as required. The client direction is unaffected: cfg.Thumbprint
is already initialised at session-open time from the endpoint's
ServerCertificate, so re-deriving here is a no-op there.

Per OPC UA Part 6 §6.7.2, the next outgoing asymmetric chunk must carry
ReceiverCertificateThumbprint = SHA1(receiverCertificate). This complements
the server-side Basic256Sha256 fix in #817, which forces SignAndEncrypt on
the OPN path but does not populate the thumbprint.

### DIFF
--- a/uasc/secure_channel.go
+++ b/uasc/secure_channel.go
@@ -479,6 +479,13 @@ func (s *SecureChannel) readChunk() (*MessageChunk, error) {
 		s.cfg.SecurityPolicyURI = m.SecurityPolicyURI
 		if m.SecurityPolicyURI != ua.SecurityPolicyURINone {
 			s.cfg.RemoteCertificate = m.AsymmetricSecurityHeader.SenderCertificate
+			// Re-derive the receiver thumbprint from the peer certificate
+			// learned in the inbound OPN. No-op for the client path, where
+			// Thumbprint is already set from the endpoint's ServerCertificate
+			// at session open.
+			if len(s.cfg.RemoteCertificate) > 0 {
+				s.cfg.Thumbprint = uapolicy.Thumbprint(s.cfg.RemoteCertificate)
+			}
 			debug.Printf("uasc %d: setting securityPolicy to %s", s.c.ID(), m.SecurityPolicyURI)
 
 			remoteCert, err := x509.ParseCertificate(s.cfg.RemoteCertificate)


uasc/secure_channel.go
```diff
@@ -479,6 +480,18 @@ func (s *SecureChannel) readChunk() (*MessageChunk, error) {
...
+			// OPC UA Part 6 §6.7.2 — the next outgoing asymmetric chunk
+			// on this channel must carry
+			// ReceiverCertificateThumbprint = SHA1(remoteCertificate).
+			// On the server side this is the SHA1 of the client cert
+			// learned from the inbound OPN; on the client side it is
+			// already initialised at session-open time from the
+			// endpoint's ServerCertificate, so re-deriving here is a
+			// no-op for the client direction.
+			if len(s.cfg.RemoteCertificate) > 0 {
+				thumb := sha1.Sum(s.cfg.RemoteCertificate)
```
istyf: I would prefer a call to uapolicy.Thumbprint instead of duplicating the functionality here. It also keeps the sha1 import confined to the policy/crypto layer.

uasc/secure_channel.go
```diff
@@ -479,6 +480,18 @@ func (s *SecureChannel) readChunk() (*MessageChunk, error) {
 		s.cfg.SecurityPolicyURI = m.SecurityPolicyURI
 		if m.SecurityPolicyURI != ua.SecurityPolicyURINone {
 			s.cfg.RemoteCertificate = m.AsymmetricSecurityHeader.SenderCertificate
+			// OPC UA Part 6 §6.7.2 — the next outgoing asymmetric chunk
```
istyf: I would trim this comment to a short statement about re-deriving not affecting the client path. Quoting the spec for every change gets a bit wordy.

istyf (changes requested):

istyf:
@JeremyTheocharis I can confirm this issue and that the PR seems to fix it. The current thumbprint code does not handle certificate chains properly though, but I think that should be a separate issue.

mschm:
Thanks @istyf — addressed both points: the thumbprint now uses `uapolicy.Thumbprint` (so the `crypto/sha1` import is gone from the `uasc` layer), and I trimmed the comment to a short note about the client path. Agreed the certificate-chain handling is out of scope here and better tracked as a separate issue.

istyf (approved):
/lgtm

jminardi:
I've been running a similar fix in my server fixes branch and just got around to opening PR #867 

I removed my fix from that PR and support merging in this one if its approved

PR #863, opened 2026-06-03
fix(uasc): chunk server responses to honor the negotiated SendBufferSize

mschm:
The two server-side response send paths (SendMsgWithContext,
sendResponseWithContext) encoded and wrote each response as a single
message and never split it into chunks. Responses larger than the
channel's negotiated SendBufferSize were emitted as one oversized chunk,
which strict clients (e.g. Prosys, UaExpert) reject with
"chunk too large (max = 65535)".

Unify both paths with the request-side sendAsyncWithTimeout chunk loop by
extracting sendChunkedMessage: encode via m.EncodeChunks(maxBodySize),
advance the SequenceNumber on each subsequent chunk, sign and encrypt per
chunk, and write each chunk separately. The instance lock is held across
the multi-chunk emission in both paths so concurrent sends cannot
interleave chunks on the wire and sequence-number allocation stays
monotonic.

Per OPC UA Part 6 §7.1.4, every chunk must fit the receiver's advertised
SendBufferSize. maxBodySize is derived from the negotiated buffer size
during handleOpenSecureChannelRequest.

### DIFF
--- a/uasc/secure_channel.go
+++ b/uasc/secure_channel.go
@@ -1074,38 +1074,20 @@ func (s *SecureChannel) SendMsgWithContext(ctx context.Context, instance *channe
 		}
 	}
 
-	// we need to get a lock on the sequence number so we are sure to send them in the correct order.
+	// Hold the instance lock across the multi-chunk emission
+	// so concurrent sends on the same channel cannot interleave chunks on
+	// the wire and so sequence-number allocation stays monotonic.
+	instance.Lock()
+	defer instance.Unlock()
+
 	// encode the message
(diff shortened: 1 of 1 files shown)


istyf:
Thank you so much for your contribution, but this issue already has a fix waiting for review in #861 .

mschm:
Thanks @istyf — #861 covers the same two server send paths (`SendMsgWithContext` / `sendResponseWithContext`) via the shared chunk helper and is further along, so I'll defer to it and close this in its favor. Happy to review #861.

PR #864, opened 2026-06-03
fix(client): inherit channel SecurityPolicyURI when UserTokenPolicy URI is empty

mschm:
Per OPC UA Part 4 §7.36, a UserTokenPolicy whose SecurityPolicyURI is empty or
null inherits the SecureChannel's SecurityPolicy. SecurityFromEndpoint and the
EncryptUserPassword / NewUserTokenSignature paths already handled this, but used
a strict "" comparison that missed whitespace-only values advertised by some
servers (e.g. legacy node-opcua endpoints).

Treat a whitespace-only per-token SecurityPolicyURI as empty and fall back to
the channel's SecurityPolicyURI across SecurityFromEndpoint, EncryptUserPassword
and NewUserTokenSignature, with the Part 4 §7.36 contract documented in the doc
comments. Adds regression tests for the username-token / empty-policy-URI case.

### DIFF
--- a/config.go
+++ b/config.go
@@ -348,7 +348,12 @@ func setCertificate(cert []byte, cfg *Config) error {
 }
 
 // SecurityFromEndpoint sets the server-related security parameters from
-// a chosen endpoint (received from GetEndpoints())
+// a chosen endpoint (received from GetEndpoints()).
+//
+// A UserTokenPolicy with an empty SecurityPolicyURI inherits the
+// SecureChannel's SecurityPolicy (OPC UA Part 4 §7.36), so this falls back to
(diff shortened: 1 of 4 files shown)


mschm:
Apologies for the noise here — on a closer look this isn't needed. gopcua already inherits the channel's SecurityPolicy when a UserTokenPolicy's SecurityPolicyURI is empty/null (which is the case real servers, including node-opcua, actually produce). My change only added whitespace-only trimming on top, and I couldn't tie that to any real endpoint — the spec treats empty/null as the inherit case, not whitespace. Closing in favor of the existing behavior. Thanks for the quick, helpful reviews on the other PRs!

PR #865, opened 2026-06-11
server: implement Browse continuation points and BrowseNext

jminardi:
This implements continuation points for the server's View service, so a client can page through a node that has a large number of references using `Browse` + `BrowseNext`.

This came up in #861 where @fuantaji ran into a wall browsing a large address space from UaExpert (the reassembled response exceeds UaExpert's limit), and @istyf suggested it deserved its own fix. Now when a client sets `RequestedMaxReferencesPerNode`, the server now returns that many references plus a continuation point, and `BrowseNext` walks the rest.

How it works:

- `Browse` now honors `RequestedMaxReferencesPerNode`. If a node has more references than the limit, it returns the first page and a continuation point. If the client doesn't set a limit (`0`), behavior is unchanged and everything comes back in one response with no continuation point, so there's zero overhead for clients that don't need this.
- `BrowseNext` resumes from a continuation point, returns the next page, and hands back a new continuation point until the references are exhausted. It also handles `ReleaseContinuationPoints`.

In my implementation the continuation point is stateless, it encodes the original `BrowseDescription` plus an offset, so `BrowseNext` re-browses the node and skips to that offset. The upside is the server keeps no per-session browse state; the trade-off is it re-walks the node's references each round.

The `server` package didn't have any tests yet, so I added some:

- continuation-point encode/decode round-trip, plus rejection of malformed points with `BadContinuationPointInvalid`
- a full `Browse` + `BrowseNext` paging loop over 250 references at page size 10, asserting every reference comes back exactly once (matches an unpaged browse)
- `ReleaseContinuationPoints`
- an invalid continuation point reported per-result without failing the whole call

### DIFF
--- a/server/namespace_map.go
+++ b/server/namespace_map.go
@@ -1,6 +1,8 @@
 package server
 
 import (
+	"maps"
+	"slices"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


jminardi:
@JeremyTheocharis any chance you can take a look at this?

JeremyTheocharis (approved):

JeremyTheocharis:
Yes, ready to be merged! Thank you @jminardi and @istyf !

currently figuring out why the tests are not running though...

PR #867, opened 2026-06-16
server: RegisterNodes identity, node attribute fixes

jminardi:
A few small server-side fixes I hit while getting the gopcua server working with real clients (telegraf, plus encrypted UaExpert / node-opcua). They're independent of each other but all small, so I bundled them. Two are in `server`, one is in `uasc`.

1. RegisterNodes/UnregisterNodes as identity (`server`)

The server returned `BadServiceUnsupported` for `RegisterNodes`, which breaks clients that call it on connect. `RegisterNodes` is an optimization hint, and when a server doesn't optimize node access the spec says it *shall* return the NodeIds from the request unchanged (Part 4, 5.8.5), so that's what this does. `UnregisterNodes` keeps no state to release and succeeds.

2. Node attribute fixes (`server`)

Two unchecked type assertions on node attributes, both currently panic:

- `NewFolderNode` / `NewVariableNode` stored a `uint32` NodeClass in the `Description` attribute, which holds a `LocalizedText` so calling `Node.Description()` on any folder or variable node panics today.
- `Node.DataType()` asserted `*ua.ExpandedNodeID` on the stored attribute, and this one is reachable remotely: the Write service stores client-supplied values without validation (`Node.SetAttribute`), and the DataType *attribute* is spec-typed `NodeId` so a single client write plants a `*ua.NodeID` that panics the server on the next Browse that resolves a reference to that node. `DataType()` now accepts either form and falls back to a null id otherwise.

(There are a few more accessors with the same unchecked-assert pattern, e.g. `BrowseName()`. I kept this PR to the two that affected me today and can do a follow-up sweep if you want one.)

~~3. OPN ReceiverCertificateThumbprint (`uasc`)~~

~~The server never computed the client certificate thumbprint, so its OpenSecureChannel response went out with an empty `ReceiverCertificateThumbprint`. Spec-compliant clients reject that (Part 6, 6.7.2). This computes the SHA-1 of the client certificate when a server channel receives the OPN request (gated on the channel kind, so client channels are untouched).~~

Removed, PR #862 also implemented the Thumbprint fix and was reviewed/approved already, so it should be merged instead of the commit in this PR.

Tests

The `server` package didn't have any tests yet; these run fully in-process. Both attribute tests panic on current `main`: `Description()` of a fresh variable node, and `DataType()` after a client-style `SetAttribute` write. RegisterNodes has an identity round-trip test asserting the ids come back unchanged, in order, for both services.


JeremyTheocharis:
nice, currently working through the PRs here. @istyf

jminardi:
Removed my thumbprint fix since it has already been addressed in #862

server/node.go
```diff
@@ -89,7 +89,7 @@ func NewFolderNode(nodeID *ua.NodeID, name string) *Node {
 			ua.AttributeIDNodeClass:     DataValueFromValue(uint32(ua.NodeClassObject)),
 			ua.AttributeIDBrowseName:    DataValueFromValue(attrs.BrowseName(name)),
 			ua.AttributeIDDisplayName:   DataValueFromValue(attrs.DisplayName(name, name)),
-			ua.AttributeIDDescription:   DataValueFromValue(uint32(ua.NodeClassObject)),
+			ua.AttributeIDDescription:   DataValueFromValue(attrs.DisplayName(name, name)),
```
istyf, line 92: The second argument to DisplayName here is the locale to set for the LocalizedText instance that is being created. Name is obviously rarely the correct value to pass for a locale, but since it is called this way in a lot of places already I think we can live with it for a short while longer until we (soon) get around to cleanup the type management and the attribute helpers.

server/node.go
```diff
@@ -117,7 +117,7 @@ func NewVariableNode(nodeID *ua.NodeID, name string, value any) *Node {
 				ua.AttributeIDNodeClass:     DataValueFromValue(uint32(ua.NodeClassVariable)),
 				ua.AttributeIDBrowseName:    DataValueFromValue(attrs.BrowseName(name)),
 				ua.AttributeIDDisplayName:   DataValueFromValue(attrs.DisplayName(name, name)),
-				ua.AttributeIDDescription:   DataValueFromValue(uint32(ua.NodeClassVariable)),
+				ua.AttributeIDDescription:   DataValueFromValue(attrs.DisplayName(name, name)),
```
istyf, line 120: same here

server/node.go
```diff
@@ -139,7 +139,7 @@ func NewVariableNode(nodeID *ua.NodeID, name string, value any) *Node {
 			ua.AttributeIDNodeClass:     DataValueFromValue(uint32(ua.NodeClassVariable)),
 			ua.AttributeIDBrowseName:    DataValueFromValue(attrs.BrowseName(name)),
 			ua.AttributeIDDisplayName:   DataValueFromValue(attrs.DisplayName(name, name)),
-			ua.AttributeIDDescription:   DataValueFromValue(uint32(ua.NodeClassVariable)),
+			ua.AttributeIDDescription:   DataValueFromValue(attrs.DisplayName(name, name)),
```
istyf, line 142: same here

istyf (approved):
This looks like a good step towards more robust type management in the server. Lots of work remaining in this area though.

jminardi:
@istyf thanks for reviewing. And sorry I should have dug into the `DisplayName` arguments rather than just copy/paste from another line, but I agree it can all be fixed in a separate PR. @JeremyTheocharis I am happy to make that PR after this is merged if you'd like, or leave it up to someone else.

jminardi:
@JeremyTheocharis this was approved by istyf, are there any other changes you'd like to see before approving?

JeremyTheocharis (approved):

PR #871, opened 2026-06-19
feat: add PKCS#8 private key support

### LINKED ISSUE #870: Feat: support PKCS#8 private keys in PrivateKeyFile / loadPrivateKey
Currently `PrivateKeyFile` (and the underlying `loadPrivateKey` function) only accepts **PKCS#1** encoded RSA private keys (`-----BEGIN RSA PRIVATE KEY-----`). Any key in **PKCS#8** format (`-----BEGIN PRIVATE KEY-----`) is silently rejected with the unhelpful error:

```
opcua: Failed to decode PEM block with private key
```

### Why this matters

PKCS#8 is the **modern default** for private key generation:

| Tool / Platform | Default output format |
|---|---|
| `openssl genpkey -algorithm RSA` | PKCS#8 |
| `openssl genrsa` + `openssl pkcs8 -topk8 -nocrypt` | PKCS#8 |
| Java (`keytool`, `KeyPairGenerator`) | PKCS#8 |
| .NET / BouncyCastle | PKCS#8 |
| OpenSSL 3.x (default since 2021) | PKCS#8 |

Users who generate their OPC UA client certificates with modern tooling, or who receive keys from Java/.NET-based PKI infrastructure, will be blocked with no clear explanation. PKCS#1 is still supported everywhere, so this is purely an **additive gap** in the library.

The same limitation affects DER-encoded key files: they are always parsed with `x509.ParsePKCS1PrivateKey` only, so a PKCS#8 DER key also fails.

stefan-kolb:
Closes #870

### Problem

`PrivateKeyFile` / `loadPrivateKey` only accepted PKCS#1 RSA keys (`BEGIN RSA PRIVATE KEY`). PKCS#8 keys (`BEGIN PRIVATE KEY`) — the default output of OpenSSL 3.x, Java, and .NET — were rejected with:

```
opcua: Failed to decode PEM block with private key
```

### Changes

**`config.go`**

Rewrote `loadPrivateKey` to handle both formats, and extracted two focused helpers:

`parsePKCS8RSAPrivateKey` explicitly rejects non-RSA PKCS#8 keys (EC, Ed25519 …) since OPC UA only supports RSA.

**`config_test.go`**

Three new table-driven test cases added to `TestOptions`:

| Test name | Expectation |
|---|---|
| `PrivateKeyFile("key-pkcs8.pem")` | succeeds, correct `*rsa.PrivateKey` loaded |
| `PrivateKeyFile("key-pkcs8.der")` | succeeds via DER fallback path |
| `PrivateKeyFile("key-pkcs8-ec.pem")` | fails: `"opcua: PKCS#8 private key is not an RSA key"` |

Test data is hardcoded (fresh keys, not derived from the existing `keyPEM`) so the two formats are independently verifiable.

### Backwards compatibility

Full — PKCS#1 keys continue to work exactly as before. No public API changes.

### DIFF
--- a/config.go
+++ b/config.go
@@ -270,18 +270,51 @@ func loadPrivateKey(filename string) (*rsa.PrivateKey, error) {
 		return nil, errors.Errorf("Failed to load private key: %s", err)
 	}
 
-	derBytes := b
 	if strings.HasSuffix(filename, ".pem") {
 		block, _ := pem.Decode(b)
-		if block == nil || block.Type != "RSA PRIVATE KEY" {
+		if block == nil {
 			return nil, errors.Errorf("Failed to decode PEM block with private key")
 		}
-		derBytes = block.Bytes
+		switch block.Type {
+		case "RSA PRIVATE KEY":
+			// PKCS#1 – the classic "openssl genrsa" output
+			pk, err := x509.ParsePKCS1PrivateKey(block.Bytes)
+			if err != nil {
+				return nil, errors.Errorf("Failed to parse PKCS#1 private key: %s", err)
+			}
+			return pk, nil
+		case "PRIVATE KEY":
+			// PKCS#8 – the modern "openssl genpkey" default
+			return parsePKCS8RSAPrivateKey(block.Bytes)
+		default:
+			return nil, errors.Errorf("Failed to decode PEM block with private key: unsupported type %q", block.Type)
+		}
 	}
 
-	pk, err := x509.ParsePKCS1PrivateKey(derBytes)
+	return parseRSAPrivateKeyDER(b)
+}
+
+// parseRSAPrivateKeyDER tries to parse der as a PKCS#1 RSA key first and,
+// if that fails, as a PKCS#8 key. This is necessary for DER-encoded files
+// because there is no block-type header to identify the format.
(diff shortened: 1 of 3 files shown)


stefan-kolb:
@kung-foo @JeremyTheocharis I created a possible implementation for my feature request #870. Would love, if you can take a look at this. Thanks, Stefan

JeremyTheocharis:
@stefan-kolb Thanks for this. PKCS#8 support is a real gap and the change looks correct and backwards-compatible. I pushed a maintainer conformance test to your branch (`tests/go/private_key_pkcs8_test.go`) that drives a real `SignAndEncrypt` handshake with a PKCS#8 key (PEM and DER), red on `main` / green on your branch. I'll open a separate follow-up PR for some loader-hardening I noticed (content-sniff instead of extension dispatch, PEM fullchain handling) so it doesn't bulk up your feature PR.

JeremyTheocharis (approved):
LGTM — maintainer conformance test added, CI green.

stefan-kolb:
@JeremyTheocharis Thanks for merging this so quickly! When can we expect a new release with the included functionality? We'd love to use it in production soon 😄If I can help in any way, just tell me.

PR #877, opened 2026-06-30
fix restoreSubscriptions leaving the client connected when a subscription fails to recreate

### LINKED ISSUE #876: restoreSubscriptions leaves the client Connected with no subscriptions when recreateSubscription fails
## What happens

During auto-reconnect, when `transferSubscriptions` fails and the client falls back to `restoreSubscriptions`, a subscription that cannot be recreated (e.g. the server returns `StatusBadNodeIDUnknown` because a monitored node is no longer valid) leaves the client permanently broken: it reports `Connected` but has zero active subscriptions and a dead publish loop. No data is ever delivered again and there are no further reconnection attempts.

Debug log:

```text
debug: client: monitor: action: transferSubscriptions
debug: client: monitor: sub 6647: transfer subscription failed
debug: sub 6647: recreate_delete: subscription deleted
debug: sub 6647: recreate_create: recreated as subscription 8581
debug: client: monitor: recreate subscripitions failed: The node id refers to a node that does not exist in the server address space. StatusBadNodeIDUnknown (0x80340000)
debug: client: monitor: no subscriptions to resume
```

## Root cause

In `client.go`, the `restoreSubscriptions` case sets `action = recreateSession` when `recreateSubscription` returns an error, intending to rebuild the session and retry:

```go
for _, subID := range subsToRecreate {
    if err := c.recreateSubscription(ctx, subID); err != nil {
        dlog.Printf("recreate subscripitions failed: %v", err)
        action = recreateSession
        continue
    }
    activeSubs++
}

c.setState(ctx, Connected)
action = none
```

The two lines after the loop run unconditionally, so `action = none` immediately overrides the `action = recreateSession` set on failure and `setState(ctx, Connected)` marks the client healthy. The intended retry never runs; the failed subscriptions are silently dropped and the client sits in `Connected` with `activeSubs == 0`.

This is the same class of bug as #854 / #855 (a reconnect path that silently abandons subscriptions), but on the `restoreSubscriptions` path rather than `restoreSession`.

## Fix

Only mark the client `Connected` and clear the action once every subscription was actually restored; otherwise keep `action = recreateSession` so the outer reconnect loop rebuilds the session and retries.

## Environment

gopcua v0.9.0, observed via Telegraf inputs.opcua_listener against a Siemens OpenPCS7 OPC UA server after an intermittent network drop.

skartikey:
Resolves #876.

During auto-reconnect, the `restoreSubscriptions` case set `action = recreateSession` when `recreateSubscription` failed, but then ran `c.setState(ctx, Connected)` and `action = none` unconditionally right after the loop. The unconditional `action = none` overrode the retry, so a subscription that could not be recreated (e.g. the server returns `StatusBadNodeIDUnknown` for a node that disappeared during the outage) was silently dropped and the client was left in `Connected` state with zero active subscriptions and a dead publish loop. No data was delivered again and no further reconnection was attempted.

This change marks the client `Connected` and clears the action only when every subscription was actually restored. If a subscription failed to recreate, `action` stays `recreateSession`, so the outer reconnect loop rebuilds the session and retries.

It also waits `ReconnectInterval` before retrying. Without the delay, a subscription that can never be restored (a permanently removed node) made the `recreateSession` -> `transferSubscriptions` -> `restoreSubscriptions` cycle hot-loop and hammer the server, since that path has no backoff of its own. The wait reuses the same interval as the `createSecureChannel` Dial loop and returns early if the context is cancelled.

The retry only helps if the subscription is still there to retry, and it was not. `recreateSubscription` forgot the subscription before creating its replacement, so a rejected `CreateSubscription` dropped it from `c.subs` for good, the next round had nothing to transfer, and the client came up `Connected` with zero subscriptions anyway. The last commit forgets the old id only once the replacement exists. `recreate_create` assigns the new id on success, so the old id captured by the caller is still the right key to forget, and a failure now leaves the subscription registered for the next attempt.

To make any of this testable, the `restoreSubscriptions` body moved out of the `monitor` switch into `(*Client).republishOrRecreateSubscriptions`, which returns the next reconnect action. Subscriptions already reach the client through `ClientInterface`, so the tests inject a stub server rather than needing a real one. `TestRepublishOrRecreateSubscriptions` covers nothing to restore, a successful recreate, a failed recreate, one of two subscriptions failing, a republish falling back to a recreate, and a failed create keeping the subscription registered. Reverting either fix breaks the corresponding cases.

Note this makes a subscription that can never be recreated retry forever at `ReconnectInterval` rather than being dropped once. That matches the `createSecureChannel` Dial loop, and it is the point of the change, but it is a behaviour change worth calling out.

Same class of bug as #854 / #855, on the `restoreSubscriptions` path instead of `restoreSession`.

### DIFF
--- a/client.go
+++ b/client.go
@@ -585,6 +585,17 @@ func (c *Client) monitor(ctx context.Context) {
 						if action == restoreSubscriptions {
 							c.setState(ctx, Connected)
 							action = none
+						} else {
+							// a subscription failed to recreate and action is
+							// recreateSession. back off before retrying so we don't
+							// hot-loop hammering the server when a subscription can
+							// never be restored (e.g. a monitored node was permanently
+							// removed from the server's address space).
+							select {
+							case <-ctx.Done():
+								return
+							case <-time.After(c.cfg.sechan.ReconnectInterval):
+							}
 						}
 
 					case abortReconnect:
--- a/client.go
+++ b/client.go
@@ -576,8 +576,16 @@ func (c *Client) monitor(ctx context.Context) {
 							activeSubs++
 						}
 
-						c.setState(ctx, Connected)
-						action = none
+						// only mark the client connected once every subscription has
+						// been restored. if a subscription failed to recreate, action
+						// was set to recreateSession above so the outer loop rebuilds
+						// the session and retries. running these two lines
+						// unconditionally dropped the failed subscriptions and left the
+						// client in Connected state with a dead publish loop.
+						if action == restoreSubscriptions {
+							c.setState(ctx, Connected)
+							action = none
+						}
 
 					case abortReconnect:
 						dlog.Printf("action: abortReconnect")
--- a/client.go
+++ b/client.go
@@ -558,44 +558,21 @@ func (c *Client) monitor(ctx context.Context) {
 						// Assume that subsToRecreate and subsToRepublish have been
 						// populated in the previous step.
 
-						activeSubs = 0
-						for _, subID := range subsToRepublish {
-							if err := c.republishSubscription(ctx, subID, availableSeqs[subID]); err != nil {
-								dlog.Printf("republish of subscription %d failed", subID)
-								subsToRecreate = append(subsToRecreate, subID)
-							}
-							activeSubs++
-						}
-
-						for _, subID := range subsToRecreate {
-							if err := c.recreateSubscription(ctx, subID); err != nil {
-								dlog.Printf("recreate subscripitions failed: %v", err)
-								action = recreateSession
-								continue
-							}
-							activeSubs++
-						}
-
-						// only mark the client connected once every subscription has
-						// been restored. if a subscription failed to recreate, action
-						// was set to recreateSession above so the outer loop rebuilds
-						// the session and retries. running these two lines
-						// unconditionally dropped the failed subscriptions and left the
-						// client in Connected state with a dead publish loop.
-						if action == restoreSubscriptions {
+						action, activeSubs = c.republishOrRecreateSubscriptions(ctx, subsToRepublish, subsToRecreate, availableSeqs)
+						if action == none {
 							c.setState(ctx, Connected)
-							action = none
-						} else {
-							// a subscription failed to recreate and action is
-							// recreateSession. back off before retrying so we don't
-							// hot-loop hammering the server when a subscription can
-							// never be restored (e.g. a monitored node was permanently
-							// removed from the server's address space).
-							select {
-							case <-ctx.Done():
-								return
-							case <-time.After(c.cfg.sechan.ReconnectInterval):
-							}
+							break
+						}
+
+						// at least one subscription could not be recreated and action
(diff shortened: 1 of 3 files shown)


skartikey:
@JeremyTheocharis friendly ping on this one when you get a chance. 

A quick recap of what it fixes: when a subscription cannot be recreated during auto reconnect, for example because a monitored node id no longer exists in the server address space and the server answers StatusBadNodeIDUnknown, restoreSubscriptions gives up but leaves the client in the Connected state with no active subscriptions and a dead publish loop. From the outside the client looks perfectly healthy, so callers have nothing to react to, never reconnect, and data collection stops until the process is restarted. The full log trace is in #876.

Someone else ran into the same thing independently: jminardi reported on #876 that collection stops and never resumes whenever Kepware or the network goes down and comes back, which matches this state machine path exactly.

Downstream this is influxdata/telegraf#19186 in the opcua_listener plugin. Telegraf pins v0.9.0, so we cannot pick the fix up until this merges and a new tag goes out.

JeremyTheocharis:
Hi, sorry for the late response. I was on vacation and on conferences :) 

Would it be possible for you to add a test for this and ideally reference either the OPC UA spec and dielly also other reference implementations on how they solve it? With this we can start to build up our own test suite and start preventing regressions over time. Thank you!

skartikey:
@JeremyTheocharis No worries, thanks for coming back to it. Added tests and rebased on main.

To make the reconnect step testable I moved it out of the big `monitor` switch into `(*Client).republishOrRecreateSubscriptions`, which returns the next reconnect action. Subscriptions already reach the client through `ClientInterface`, so the test just injects a stub server.

`TestRepublishOrRecreateSubscriptions` has five cases, the important one being the exact sequence from the log in #876: the server creates the new subscription but returns Bad_NodeIdUnknown for the monitored item. Before this fix the step reported success there. If I remove the fix, three of the five cases fail.

On the references you asked for:

Part 4 5.13.2 (CreateMonitoredItems) says a bad NodeId is a per-item result, not a subscription failure: "Monitored Nodes can be removed from the AddressSpace after the creation of a MonitoredItem. This does not affect the validity of the MonitoredItem but a Bad_NodeIdUnknown shall be returned in the Publish response." UA-.NETStandard (`MonitoredItem.SetCreateResult`), node-opcua (`recreateSubscriptionAndMonitoredItem`) and open62541 (`ua_client_subscriptions.c`) all handle it that way, per item, keeping the subscription.

Part 4 5.13.7 (TransferSubscriptions) has nothing normative on what a client does when restore fails, so that part is our call. For the retry delay I followed node-opcua, whose `repair_client_session` waits 2000 ms before retrying instead of looping immediately.

JeremyTheocharis (approved):
Verified: this fixes the dead-assignment bug in the reconnect state machine where a failed recreateSubscription set `action = recreateSession` only to be overwritten by `action = none`, reporting Connected with a dead subscription. The per-item restore keeps the subscription registered until the replacement exists. Both existing unit tests and a python integration fixture pass.

PR #873, opened 2026-06-23
server: normalise node data type attribute reads to nodeid before returning them

istyf:
This PR adds DataType node attribute read normalisation so that the value that is currently internally held as either a NodeID or an ExtendedNodeID is presenteds to clients as a NodeID as the standard mandates.

The fix avoids mutating the stored attribute, preserves the existing special handling for NodeId/EventNotifier, and simplifies the common error handling path.

Fixes #849

PR #878, opened 2026-07-01
fix(server): apply AccessLevel only to the Value attribute

mschm:
`NodeNameSpace.Attribute()` gated **every** attribute read behind
`AccessLevelTypeCurrentRead`. A Variable that denies current-read therefore
returned `Bad_UserAccessDenied` for all of its attributes — `BrowseName`,
`NodeClass`, `DisplayName`, and the rest — leaving the node impossible to
browse or describe.

### Why this is wrong

`AccessLevel` and `UserAccessLevel` are Variable-class attributes that describe
how the **Value** attribute may be accessed. OPC UA Part 3 §5.6.2:

> The AccessLevel Attribute is used to indicate how the Value of a Variable can
> be accessed (read/write) and if it contains current and/or historic data.

They do not govern the other attributes. A Read of a non-Value attribute
returns its own operation-level `StatusCode` (Part 4 §5.10.2) and must not be
denied on AccessLevel grounds.

This is the read-side counterpart to the non-Value **write** behavior already
covered by `tests/go/write_nonvalue_attribute_test.go`, whose documentation
states: *"AccessLevel governs the Value attribute only."*

### Change

Gate the access check on `attr == ua.AttributeIDValue`, so a write-only node
stays browsable and describable while its `Value` remains access-protected.
Adds a table test asserting non-Value reads succeed and the `Value` read still
returns `Bad_UserAccessDenied`.

### Note

This touches `NodeNameSpace.Attribute()`, which #873 also refactors. The two
changes are orthogonal (access gate vs. DataType-read normalization); happy to
rebase onto whichever lands first.

### DIFF
--- a/server/namespace_node.go
+++ b/server/namespace_node.go
@@ -122,7 +122,10 @@ func (as *NodeNameSpace) Attribute(id *ua.NodeID, attr ua.AttributeID) *ua.DataV
 		}
 	}
 
-	if !n.Access(ua.AccessLevelTypeCurrentRead) {
+	// OPC UA Part 3 §5.6.2: AccessLevel restricts the Value attribute only.
+	// All other attributes (NodeClass, BrowseName, DisplayName, …) must
+	// remain readable regardless of the access level.
+	if attr == ua.AttributeIDValue && !n.Access(ua.AccessLevelTypeCurrentRead) {
 		return &ua.DataValue{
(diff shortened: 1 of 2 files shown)


JeremyTheocharis:
@istyf ?

JeremyTheocharis (approved):
Verified the spec-Value-only gate and ran the gate (green with fix, red with it reverted). Branch current, 4 CI cells green. Merging.

PR #881, opened 2026-07-12
monitor: give each request its own MonitoringParameters copy

### LINKED ISSUE #880: monitor: batched AddMonitorItems maps every item to the last node when Requests share a MonitoringParameters
In `monitor/subscription.go` (v0.9.0), **AddMonitorItems** writes ClientHandle through the caller's MonitoringParameters pointer:

```go
if node.MonitoringParameters != nil {
    request.RequestedParameters = node.MonitoringParameters // aliases caller's struct
    request.RequestedParameters.ClientHandle = handle        // mutates it in place
}
```

If several Requests share one `*ua.MonitoringParameters`, each iteration overwrites **ClientHandle** on that same struct, so all requests end up with the last handle. The server then reports every change under that one handle and the client resolves them all to the last node — every notification appears to come from the last node.

Same pattern in **ModifyMonitorItems**. (The Subscribe/AddNodes path is unaffected since it leaves MonitoringParameters nil.)

## Reproduce the issue

Server with a few fast nodes, e.g. opc-plc:

```
docker run --rm -p 50000:50000 mcr.microsoft.com/iotedge/opc-plc:latest \
  --pn=50000 --autoaccept --unsecuretransport --fn=3 --vfr=50 --ft=UInt
```

```
go mod init repro
go get github.com/gopcua/opcua@v0.9.0
```

main.go - Subscribe to 3 nodes via **AddMonitorItems**, sharing one ***ua.MonitoringParameters**:

```go
package main

import (
      "context"
      "log"
      "time"

      "github.com/gopcua/opcua"
      "github.com/gopcua/opcua/monitor"
      "github.com/gopcua/opcua/ua"
)

func main() {
      ctx := context.Background()

      c, err := opcua.NewClient("opc.tcp://localhost:50000")
      if err != nil {
              log.Fatal(err)
      }
      if err := c.Connect(ctx); err != nil {
              log.Fatal(err)
      }
      defer c.Close(ctx)

      m, err := monitor.NewNodeMonitor(c)
      if err != nil {
              log.Fatal(err)
      }

      ch := make(chan *monitor.DataChangeMessage, 16)
      sub, err := m.ChanSubscribe(ctx, &opcua.SubscriptionParameters{Interval: time.Second}, ch)
      if err != nil {
              log.Fatal(err)
      }
      defer sub.Unsubscribe(ctx)

      // One Request per node, all SHARING a single *ua.MonitoringParameters.
      params := &ua.MonitoringParameters{}
      ids := []string{"ns=3;s=FastUInt1", "ns=3;s=FastUInt2", "ns=3;s=FastUInt3"}
      reqs := make([]monitor.Request, len(ids))
      for i, id := range ids {
              reqs[i] = monitor.Request{
                      NodeID:               ua.MustParseNodeID(id),
                      MonitoringMode:       ua.MonitoringModeReporting,
                      MonitoringParameters: params, // shared pointer
              }
      }
      if _, err := sub.AddMonitorItems(ctx, reqs...); err != nil {
              log.Fatal(err)
      }

      const total = 30
      counts := map[string]int{}
      for i := 0; i < total; i++ {
              counts[(<-ch).NodeID.String()]++
      }
      for _, id := range ids {
              log.Printf("%-16s : %d", id, counts[id])
      }
      log.Printf("distinct NodeIDs seen: %d (expected %d)", len(counts), len(ids))
}
```

Expected: notifications spread across all three nodes.

Actual — every notification resolves to the last node:

ns=3;s=FastUInt1  : 0
ns=3;s=FastUInt2  : 0
ns=3;s=FastUInt3  : 30
distinct NodeIDs seen: 1 (expected 3)

## Suggested fix

Copy before mutating (shallow copy suffices — only ClientHandle is written); same for ModifyMonitorItems:

```go
if node.MonitoringParameters != nil {
    params := *node.MonitoringParameters
    params.ClientHandle = handle
    request.RequestedParameters = &params
}
```

jackchenjc:
AddMonitorItems and ModifyMonitorItems assigned the caller's *ua.MonitoringParameters into the request and then wrote ClientHandle through that pointer. When several requests shared one instance, each iteration overwrote ClientHandle on the same struct, leaving every item with the last handle so all notifications resolved to the last node. Copy the parameters per request instead.

Fixes #880

### DIFF
--- a/monitor/subscription.go
+++ b/monitor/subscription.go
@@ -345,8 +345,10 @@ func (s *Subscription) AddMonitorItems(ctx context.Context, nodes ...Request) ([
 		request.MonitoringMode = node.MonitoringMode
 
 		if node.MonitoringParameters != nil {
-			request.RequestedParameters = node.MonitoringParameters
-			request.RequestedParameters.ClientHandle = handle
+			// copy so each request owns its ClientHandle; sharing would map all items to the last node
+			params := *node.MonitoringParameters
+			params.ClientHandle = handle
+			request.RequestedParameters = &params
 		}
 		toAdd = append(toAdd, request)
 	}
@@ -484,11 +486,13 @@ func (s *Subscription) ModifyMonitorItems(ctx context.Context, nodes ...Request)
 				break
 			}
 
+			// Copy so each request owns its parameters (see AddMonitorItems).
+			params := *node.MonitoringParameters
+			params.ClientHandle = item.handle
 			request := &ua.MonitoredItemModifyRequest{
 				MonitoredItemID:     item.id,
-				RequestedParameters: node.MonitoringParameters,
+				RequestedParameters: &params,
 			}
-			request.RequestedParameters.ClientHandle = item.handle
 			toModify = append(toModify, request)
 			break
 		}
--- a/monitor/subscription_test.go
+++ b/monitor/subscription_test.go
@@ -1,0 +1,256 @@
+package monitor
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"slices"
+	"testing"
+	"time"
+
+	"github.com/gopcua/opcua"
+	"github.com/gopcua/opcua/server"
+	"github.com/gopcua/opcua/ua"
+	"github.com/stretchr/testify/require"
+)
+
+// notificationTimeout caps how long a test waits for notifications. The server
+// pushes an initial value as soon as a monitored item is created, so a passing
+// run needs a small fraction of this. The cap is what keeps a failing run
+// affordable: the notifications a regression loses never arrive, so the test
+// waits out the whole deadline on every required CI context.
+const notificationTimeout = 5 * time.Second
+
+// testNodes are the variable nodes the loopback server exposes. Each one holds a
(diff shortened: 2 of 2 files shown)


JeremyTheocharis:
I am currently working on this and taking this as an example to start adding tests and in a second OPR refactoring the code as well a little bit ot make it easier to test and more maintainable. I hope that is OK.

JeremyTheocharis (approved):
Thanks for this, and for the issue — the reproduction made it straightforward to verify.

I pushed the regression test onto the branch rather than asking you for one. `monitor/` had no test file at all, so it needed a server fixture first, which is most of the new code.

What the test pins: with the parameter copy reverted, every notification resolves to the last node monitored and all three values arrive under `batch_c`. It also catches a permuted handle-to-NodeID mapping, where each value is delivered under some other node — the same wrong attribution, which is why the assertion compares node-to-value pairs instead of counting distinct nodes. A second test covers the nil-parameters path through `AddNodeIDs`, which nothing else exercised.

Green on all four contexts, both Windows legs included.

PR #885, opened 2026-07-22
server: fix data race in MapNamespace.Attribute (missing read lock)

Lacuno:
# server: fix data race in MapNamespace.Attribute (missing lock)

## Bug

`MapNamespace.SetValue` and `MapNamespace.SetAttribute` both take `ns.Mu.Lock()`
before reading or writing `ns.Data`. `MapNamespace.Attribute` does not take any
lock at all, even though it reads `ns.Data` in two places:

- the `AttributeIDValue` case: `v, ok := ns.Data[key]`
- the `AttributeIDDataType` case: `v := ns.Data[key]`
- a debug-log statement that reads the whole map: `ns.srv.cfg.logger.Debug("'%s' Data at read: %v", ns.name, ns.Data)`

Any application that publishes tag updates from a goroutine via `SetValue`
(the documented way to update a `MapNamespace`) while the server concurrently
serves an OPC UA Read request (`AttributeService.Read` -> `NameSpace.Attribute`)
hits a classic writer-locks/reader-doesn't data race on the `Data` map.

Confirmed under `go test -race`: a background goroutine calling `SetValue`
on a tag every couple of seconds, racing an OPC UA client `Read`, reliably
reproduces:

```
WARNING: DATA RACE
Write at ... by goroutine ...:
  github.com/gopcua/opcua/server.(*MapNamespace).SetValue()
      server/namespace_map.go:43 +0x...
      (writes ns.Data[key] = value under ns.Mu.Lock())

Previous read at ... by goroutine ...:
  github.com/gopcua/opcua/server.(*MapNamespace).Attribute()
      server/namespace_map.go:194 +0x...
      (reads ns.Data[key], no lock held)
```

(Exact line numbers shift release to release; this is the shape of the race
on `master` as of this patch, and unchanged since v0.8.0.)

**Both v0.8.0 and v0.9.0 have this bug.** v0.9.0 additionally introduces an
unlocked server-side write, `s.Data[key] = v` in the write-attribute path
(`namespace_map.go:352` in that tag), which is a second, separate unlocked
access on top of the same underlying problem -- not covered by this patch,
since this patch targets `master`'s current `Attribute`/read-side gap, but
worth a maintainer's attention when triaging this report.

## Fix

Take `ns.Mu.RLock()` for the full duration of `Attribute`, mirroring the
locking `SetValue` and `SetAttribute` already do on the write side. This is
the minimal fix: `Attribute` doesn't call any other `MapNamespace` method
that itself takes `ns.Mu` (`Objects()`/`Root()` don't lock), so a
whole-function `RLock` cannot self-deadlock, and readers can still run
concurrently with each other (RWMutex).

`Browse` and `SetAttribute` already lock correctly and are unchanged.

## Testing

- `go build ./server/` -- clean.
- `go vet ./server/` -- clean.
- `go test ./server/ -run '.*' -count=1` -- no test files exist in
  `server/` on this checkout (`ok ... [no test files]`), so this patch adds
  no regression coverage of its own beyond the downstream repro described
  above; happy to add a `-race` regression test exercising concurrent
  `SetValue`/`Attribute` calls if that's wanted before merge.

## Scope

Single file, `server/namespace_map.go`, additive lock/unlock only -- no
behavior change for callers, no API change.


JeremyTheocharis:
@istyf

PR #887, opened 2026-07-30
treat a rejected monitored item as an item level result when recreating a subscription

### LINKED ISSUE #886: recreate_monitoredItems drops every monitored item when one item returns a bad status
## What happens

When a subscription is recreated during auto-reconnect and the server rejects one monitored item, the subscription loses all of its monitored items, including the ones the server accepted. It stays registered under its new id but monitors nothing, so it never delivers data again.

## Root cause

`Subscription.recreate_monitoredItems` clears `s.items` before sending (subscription.go:465) and only refills it after a fully good response (subscription.go:490). Any single bad `MonitoredItemCreateResult.StatusCode` returns at subscription.go:486, before the refill, so `s.items` is left empty. Same for the transport error at subscription.go:481.

## Reproduction

With the stub client from #877, a subscription with one monitored item, and a server that accepts `CreateSubscription` but returns `Bad_NodeIdUnknown` for the item:

```
sub.items before recreate:   1
sub.items after recreate:    0
client subscription ids:     [8581]   (new id, still registered)
```

## Why this matters

This is the second half of #876. #877 makes the reconnect loop retry with a new session instead of reporting `Connected`, but the retry now finds a subscription with zero items, sends nothing, returns nil, and the client comes up `Connected` monitoring nothing. Same outcome, one layer down.

## Expected behaviour

A bad status for one item should be item-level, not a subscription-level failure. Part 4, 5.13.2 (CreateMonitoredItems) returns one status per item and says:

> Monitored Nodes can be removed from the AddressSpace after the creation of a MonitoredItem. This does not affect the validity of the MonitoredItem but a Bad_NodeIdUnknown shall be returned in the Publish response.

The other stacks do it that way: UA-.NETStandard records a per-item result in `MonitoredItem.SetCreateResult`, node-opcua applies each result in `recreateSubscriptionAndMonitoredItem` and only warns, open62541 deletes just the failed MonitoredItem in `ua_client_subscriptions.c`.

So `recreate_monitoredItems` should keep the accepted items, record the failed ones per item, and let the subscription continue instead of returning the first bad status for the whole restore.

Found while writing the reconnect tests in #877. Present on main.

skartikey:
Resolves #886. Stacked on #877, so please merge that one first. Only the last two commits belong to this PR.

`recreate_monitoredItems` cleared `s.items` before sending `CreateMonitoredItems` and only refilled it after a fully good response, returning the first bad `MonitoredItemCreateResult` status as the error for the whole restore. A single node that had gone from the server's address space therefore dropped every monitored item on the subscription, including the ones the server had just accepted. The recreated subscription stayed registered under its new id but monitored nothing, so it never delivered data again.

Part 4, 5.13.2 defines that status as an operation level result for that item: "Monitored Nodes can be removed from the AddressSpace after the creation of a MonitoredItem. This does not affect the validity of the MonitoredItem but a Bad_NodeIdUnknown shall be returned in the Publish response." UA-.NETStandard (`MonitoredItem.SetCreateResult`), node-opcua (`recreateSubscriptionAndMonitoredItem`) and open62541 (`ua_client_subscriptions.c`) all record it against the individual item and keep the subscription.

This change builds the restored item map locally and swaps it in at the end, so a failed request leaves the previous items in place for the next reconnect attempt to send. A rejected item is dropped and the rest are restored. Rejected items are not kept, because `s.items` is keyed by the server assigned `MonitoredItemID`, which is zero for a rejected item, so keeping them would collide and would hand `Unmonitor` and `ModifyMonitoredItems` an id the server does not have.

The drop is reported with `log.Printf`, not the debug logger. Losing a monitored item across a reconnect means the caller silently stops receiving that node, so it has to be visible without turning debug logging on. `sendRepublishRequests` already warns at that level for its retransmission buffer check.

It also guards the result count. The old code indexed `res.Results` against the requested items and would have panicked on a short response.

`TestRecreateMonitoredItems` covers four cases: accepted items survive alongside a rejected one, an all-rejected response still succeeds with an empty item set, a request error leaves the previous items in place, and a short result list returns an error instead of panicking. All four fail against the previous implementation.

The reconnect tests from #877 needed one adjustment. Their recreate failed on a rejected monitored item, which is exactly what no longer fails a recreate, so they now fail it at `CreateSubscription` instead. They still pin the state machine bug they were written for.

Together with #877 this closes the loop on #876: the reconnect loop retries instead of reporting `Connected` after a failed restore, the subscription survives to be retried, and a restore no longer throws away the monitored items that did come back.

Not covered here: `Monitor`, the initial create path, stores a rejected item under id 0, so two rejected items collide there. That is pre-existing and separate, happy to take it in a follow-up if you want it.

### DIFF
--- a/client_sub_test.go
+++ b/client_sub_test.go
@@ -91,7 +91,8 @@ func recreateResponder(newSubID uint32, itemStatus ua.StatusCode) func(ua.Reques
 
 // rejectingRecreateResponder answers DeleteSubscriptions and then rejects
 // CreateSubscription with a service level status, so the recreate fails before
-// the replacement subscription exists.
+// the replacement subscription exists. A monitored item the server rejects is
+// not enough to fail a recreate, see recreate_monitoredItems.
 func rejectingRecreateResponder(status ua.StatusCode) func(ua.Request, func(ua.Response) error) error {
 	return func(req ua.Request, h func(ua.Response) error) error {
 		switch req.(type) {
@@ -172,9 +173,8 @@ func TestRepublishOrRecreateSubscriptions(t *testing.T) {
 		require.Equal(t, []uint32{4711}, c.SubscriptionIDs())
 	})
 
-	// This is the #876 case. The server accepts the new subscription but
-	// rejects the monitored item because the node is gone from its address
-	// space, so recreateSubscription fails. The step has to hand
+	// This is the #876 case: the subscription cannot be brought back, here
+	// because the server refuses to create it. The step has to hand
 	// recreateSession back to the reconnect loop instead of none, otherwise
 	// the loop marks the client Connected with no live subscription and never
 	// tries again.
@@ -182,7 +182,7 @@ func TestRepublishOrRecreateSubscriptions(t *testing.T) {
 		c, err := NewClient("opc.tcp://example.com:4840")
 		require.NoError(t, err)
 
-		stub := &stubClient{send: recreateResponder(8581, ua.StatusBadNodeIDUnknown)}
+		stub := &stubClient{send: rejectingRecreateResponder(ua.StatusBadTooManySubscriptions)}
 		newTestSubscription(t, c, stub, 6647)
 
 		action, activeSubs := c.republishOrRecreateSubscriptions(context.Background(), nil, []uint32{6647}, nil)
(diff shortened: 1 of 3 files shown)


JeremyTheocharis (approved):
Verified against #876/#886: a per-item rejection no longer drops the other items, and the two-pass self-concealing failure is fixed (recreate now returns an operation-level result per item). Confirmed against OPC UA Part 4 §5.13.2.4 (Table 65 — CreateMonitoredItems Operation Level Result Codes) and cross-checked with UA-.NETStandard and open62541. Merging.

Follow-up worth tracking (not blocking): a rejected item is dropped from the subscription and never re-attempted on a later reconnect, so a node only transiently absent is lost for the life of the Subscription. UA-.NETStandard retains and re-sends on the next restore; worth considering as a follow-up.

Issue #748, opened 2024-11-04
Unable to write empty string to value: StatusBad(0x80000000)


jacobfrantz1:
I'm currently using this package in production with several machines that have the "Siemens Sinumerik One" control w/ opcua server.

I ran into the issue that attempting to write an empty string to a node with a string value results in a StatusBad(0x80000000). 

After some debugging, I found that for empty strings, the client sends a null value for the length of the string in the following lines: https://github.com/gopcua/opcua/blob/708f669296f58deba4d100507115e02166c0f0f9/ua/buffer.go#L273C1-L280C1. I've got my application working with a fork that writes a zero to the string length instead of the null value for empty strings.

My question is: is this package working as intended, and the servers I'm using are at fault, or is this a problem, and it isn't possible to write an empty string with the package as is.

magiconair:
Good point. We don't have any problem with this and we have a couple of thousand deployments on both Siemens and Beckhoff PLCs. @kung-foo also doesn't seem to have a problem with this. So I don't expect this to be a wide spread problem.

However, you are correct that there is the distinction between `null` and `empty` which OPC/UA allows and gopcua treats them both as `null`. (https://reference.opcfoundation.org/Core/Part6/v104/docs/5.2.2.4)

The challenge is that both server and client language have to cooperate since both can support either both or only one of them. If the string node isn't nullable you can't write a NULL string. If your programming language doesn't support null strings you have to pick one encoding. I think that's why I chose null strings at the time. 

You can write your own `Encode` and `Decode` method for your struct. Check `ua/datatypes.go` for examples. It looks like `WriteByteString` does what you want.

PR #891, opened 2026-08-10
fix: handle empty string encoding correctly

freedom-winds:
## Summary

Fix encoding of empty OPC UA strings.

## Problem

`Buffer.WriteString("")` encoded an empty string using the OPC UA null
string sentinel (`-1`). OPC UA distinguishes between:

- null string: length `-1`
- empty string: length `0`

This caused empty strings in structures such as LocalizedText and
QualifiedName to be serialized as null values.

## Changes

- Encode empty strings with a zero length.
- Update the empty-string codec expectation.
- Add a wire-level regression test.

## Validation

- `go test ./ua`
- `patch --dry-run -p1 < fix-empty-string-encoding.patch`


JeremyTheocharis:
I think we should close this PR.

First of all the tests are failing. 

And then both encodings are conformant: 

> Part 6 §5.1.11 scopes strings in explicitly: "null, empty and zero length arrays are semantically the same for all DataEncodings", and "applications shall treat null and empty arrays as equal". It also names this exact situation — "not all DevelopmentPlatforms will be able to preserve the distinction". Go's string can't be nil, so gopcua has to pick one encoding, and -1 isn't wrong. open62541 and UA-.NETStandard both accept either form on decode and flatten them.

This is related to the issue in #748 right? So ideally what we should do instead: 

> If you want to pick it up, the change that fixes it is much narrower: only encodeValue's case string in ua/variant.go, which covers the value a user actually writes and leaves the protocol fields alone. Open it against #748 rather than as a general encoding change, and bring a capture or a named device with it, since the argument there is interoperability and only server evidence settles that.

PR #897, opened 2026-08-20
client: store conn atomically to fix Close/Dial data race

barisvelioglu-maestrohub:
### The race

`Client.conn` is a plain field. `Dial()` writes it, and during an auto-reconnect `Dial()` runs on the `monitor()` goroutine — while `Close()` reads the same field on the caller's goroutine. Nothing synchronises the two:

```
Write  Client.Dial()   client.go:598  <- monitor() goroutine
Read   Client.Close()  client.go:645  <- caller
```

The sequence that triggers it is ordinary rather than exotic: a server drops, `AutoReconnect` starts re-dialling, and whoever owns the client stops it while that is in flight.

### How it shows up

Under `-race` it reports reliably. Without the detector the failure mode is a torn interface read on a `*uacp.Conn` — closing a stale connection, or a nil dereference in `Close()`.

I hit this downstream in an OPC UA connector whose compliance suite kills the server, lets the client reconnect, then disconnects in test cleanup. It reproduced on 100% of `-race` runs before this change and is clean over 5 repeats after.

### The fix

Store the connection in an `atomic.Value` behind `conn()` / `setConn()`.

This is the treatment `atomicSechan` and `atomicSession` already get in the same struct, for the same reason — the connection was the one piece of shared state still held in a plain field. Keeping the shape identical to its neighbours seemed better than introducing a mutex alongside two atomics.

`conn()` returns `nil` when nothing has been stored, so the existing `if c.conn != nil` check in `Close()` keeps working unchanged.

### Scope

- One file, no exported API change, no behaviour change.
- `go test ./... -race` passes on `main` with this applied.
- Also verified on the `v0.8.0` tag, which is where I found it.

Happy to adjust the approach if you would rather guard it with a mutex, or fold `conn` into the existing secure-channel accessor pattern differently.


JeremyTheocharis:

PR #898, opened 2026-08-20
Fix subscription reconnect and publish loop

### LINKED ISSUE #895: Publish loop can stay paused forever after an automatic reconnect
### What happened

After an automatic reconnect, the publish loop can stay paused **forever**. The client keeps
reporting `Connected`, `Read` calls keep working, but no `PublishRequest` is ever sent again, so
no `DataChangeNotification` ever arrives. There is no error returned anywhere and no way for the
caller to observe the state through the public API.

We hit this in production against KEPServerEX V6. One channel sat in this state for **38 minutes**
before we noticed; a packet capture confirmed **zero `PublishRequest`** on that connection while a
sibling connection in the same process was publishing normally once per second.

### Affected versions

`v0.9.0`, `v0.9.1`, and current `master` (`79bc0dc`) — the code paths below are identical in all three.

### Two ways to reach the stuck state

**1. Reconnect completes with `activeSubs == 0`**

`client.go` (reconnect tail):

```go
switch {
case activeSubs > 0:
    c.resumeSubscriptions(ctx)
default:
    dlog.Printf("no subscriptions to resume")   // <- nothing is resumed
}
```

`publish()` returned an error earlier (in our case `io.EOF`, because the server closed the TCP
connection), so `monitorSubscriptions` called `pauseSubscriptions`. The reconnect state machine then
rebuilds TCP / secure channel / session **successfully**, but if the server has already dropped the
subscriptions, republish fails and recreate does not make up for it, so `activeSubs` ends at 0 and
no resume is ever sent. The publish loop stays in the inner wait loop indefinitely.

Debug log from the field (one such event, all within the same second):

```
publish: error: notif 534/1 not on server anymore: BadSequenceNumberUnknown
sub: pause
uasc 5: readChunk EOF
uasc 5: Close()
client: monitor: trying to recreate secure channel
uacp 6: start HEL/ACK handshake
client: monitor: no subscriptions to resume     <-- stuck from here on
```

**2. A `resume` delivered while the loop is not paused is discarded**

`client_sub.go` `monitorSubscriptions`:

```go
case <-c.resumech:
    dlog.Print("resume")
    // ignore since not paused
```

`pausech` and `resumech` are both buffered, and the two signals can be in flight at the same time —
for example `Subscription.Cancel()` followed by `Client.Subscribe()`:

* `Cancel` → `forgetSubscription_NeedsSubMuxLock` → when `len(c.subs) == 0` → `pauseSubscriptions`
* `Subscribe` → on success → `c.resumech <- struct{}{}`

If the outer `select` happens to pick `resumech` first it is discarded as "not paused", and the
`pausech` token that is still queued then parks the loop with no resume left to release it. The
selection is random, so this is roughly a coin flip on every cancel-then-subscribe sequence.

We ran into this while implementing a workaround for (1) and had to order our own code as
"create the new subscription *before* cancelling the old one" to avoid ever having both signals
outstanding.

### Why it is hard to notice

None of the usual liveness checks fire:

* `Client.State()` stays `Connected` — the reconnect really did succeed
* `Read` / `Browse` and every other request keep working — only the publish loop is parked
* no error is delivered to the subscription's notification channel
* nothing is logged after the initial `no subscriptions to resume` unless `OPC_DEBUG=debug` is set

### Root of the design problem

`pausech` / `resumech` carry **edge** signals, but what the loop needs is a **level**: "should I be
publishing right now?". Any lost or reordered edge desynchronises the loop from reality, and there
is no way back because the only `resumeSubscriptions` call sites are `Subscribe()` and the tail of
the reconnect state machine.

### Workaround we are using

Detect the reconnect from outside the library and force a fresh subscription:

```go
opcua.StateChangedFunc(func(s opcua.ConnState) {
    // remember "left Connected and came back" -- do not block here,
    // setState calls this synchronously and blocks on stateCh
})
// ... later, from our own goroutine: create a new subscription (which emits resume),
//     then cancel the old one
```

This works, but it means every caller has to know about the internal pause/resume mechanics.

### Happy to help

We have been running the workaround in production for a while and can reproduce the stuck state on
demand. If you can indicate which direction you would accept — for example deriving the paused state
from `len(c.subs)` instead of counting edges, or surfacing the parked state to the caller — we are
glad to send a PR with tests.

stefan-kolb:
Confirmed working (logs anonymized from staging system):
```
# --- publish loop healthy before the outage ---
uasc 3/16: send *ua.PublishRequest with 110 bytes
uasc 3/16: recv *ua.PublishResponse
debug: publish: notif: 7

# --- server times out the in-flight publish; channel torn down ---
uasc 3/17: recv MSGF with 84 bytes
uasc 3/17: err: The operation timed out. StatusBadTimeout (0x800A0000)
debug: publish: error: ignoring: The operation timed out. StatusBadTimeout (0x800A0000)
debug: client: monitor: disconnected
debug: publish: eof: pausing publish loop
debug: sub: pause

# --- auto-reconnect: rebuild secure channel ---
debug: client: monitor: auto-reconnecting
debug: client: monitor: action: createSecureChannel
uasc 4/1: send *ua.OpenSecureChannelRequest
uasc 4/1: recv *ua.OpenSecureChannelResponse
debug: client: monitor: secure channel recreated

# --- same session is reactivated: this is the restoreSession path ---
debug: client: monitor: action: restoreSession
uasc 4/2: send *ua.ActivateSessionRequest
uasc 4/2: recv *ua.ActivateSessionResponse
debug: client: monitor: session restored
uasc 4/3: send *ua.ReadRequest            # UpdateNamespaces
uasc 4/3: recv *ua.ReadResponse
debug: client: monitor: namespaces updated

# --- THE FIX: restoreSession now routes through transferSubscriptions ---
gopcua reconnect: restoreSession OK -> transferSubscriptions (1 subscriptions: [6])
debug: client: monitor: action: transferSubscriptions
uasc 4/4: send *ua.TransferSubscriptionsRequest
uasc 4/4: recv *ua.TransferSubscriptionsResponse

gopcua reconnect: restoreSubscriptions (republish=[6] recreate=[])
client_sub.go: republishing subscription 6
uasc 4/5: send *ua.RepublishRequest
uasc 4/5: err: The requested notification message is no longer available. StatusBadMessageNotAvailable (0x807B0000)
client_sub.go: Republishing subscription 6 OK

# --- connected AND publish loop resumed (was left paused before the fix) ---
state="Connected"
debug: client: monitor: resuming 1 subscriptions
gopcua reconnect: done, activeSubs=1 (publish loop RESUMED)
uasc 4/6: send *ua.PublishRequest

# --- subscription keeps delivering afterwards ---
uasc 4/6: recv *ua.PublishResponse   # SubscriptionID 6
uasc 4/7: send *ua.PublishRequest
uasc 4/7: recv *ua.PublishResponse   # SubscriptionID 6
```

Closes #895

### DIFF
--- a/client.go
+++ b/client.go
@@ -475,7 +475,17 @@ func (c *Client) monitor(ctx context.Context) {
 						}
 						dlog.Printf("namespaces updated")
 
-						action = restoreSubscriptions
+						// Recover subscriptions via transferSubscriptions, exactly
+						// like recreateSession does. ActivateSession reactivated the
+						// same session, but after the secure channel was rebuilt real
+						// servers stop delivering notifications until the
+						// subscription is re-bound with TransferSubscriptions; a
+						// plain Republish is not enough and leaves the client
+						// Connected with a silently dead subscription.
+						// transferSubscriptions also populates
+						// subsToRepublish/subsToRecreate/availableSeqs and recreates
+						// subscriptions the server can no longer transfer.
+						action = transferSubscriptions
 
 					case recreateSession:
 						dlog.Printf("action: recreateSession")
--- a/client.go
+++ b/client.go
@@ -166,11 +166,21 @@ type Client struct {
 	// for all active subscriptions.
 	pendingAcks []*ua.SubscriptionAcknowledgement
 
-	// pausech pauses the subscription publish loop
-	pausech chan struct{}
+	// publishPaused is the single source of truth for whether the subscription
+	// publish loop should be running. pauseSubscriptions/resumeSubscriptions
+	// only ever store true/false here.
+	//
+	// It replaces the former pausech/resumech edge-triggered channels: both
+	// were read by the same select in monitorSubscriptions, so a resume racing
+	// a pause could be picked first and discarded, leaving the pause to park
+	// the loop forever. A level has no such race. See
+	// https://github.com/gopcua/opcua/issues/895.
+	publishPaused atomic.Bool
 
-	// resumech resumes subscription publish loop
-	resumech chan struct{}
+	// wakech nudges the publish loop to re-check publishPaused. It only carries
+	// a "something may have changed" hint, never state, so wake-ups are safe to
+	// coalesce or drop.
+	wakech chan struct{}
 
 	// mcancel stops subscription publish loop
 	mcancel func()
@@ -216,8 +226,7 @@ func NewClient(endpoint string, opts ...Option) (*Client, error) {
 		sechanErr:   make(chan error, 1),
 		subs:        make(map[uint32]*Subscription),
 		pendingAcks: make([]*ua.SubscriptionAcknowledgement, 0),
-		pausech:     make(chan struct{}, 2),
-		resumech:    make(chan struct{}, 2),
+		wakech:      make(chan struct{}, 1),
 		stateCh:     cfg.stateCh,
 		stateFunc:   cfg.stateFunc,
 	}
--- a/client_sub.go
+++ b/client_sub.go
@@ -48,7 +48,7 @@ func (c *Client) Subscribe(ctx context.Context, params *SubscriptionParameters, 
 	stats.Subscription().Add("Count", 1)
 
 	// start the publish loop if it isn't already running
-	c.resumech <- struct{}{}
+	c.resumeSubscriptions(ctx)
 
 	sub := &Subscription{
 		SubscriptionID:            res.SubscriptionID,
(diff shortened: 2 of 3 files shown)


stefan-kolb:
Please treat this as a draft for now. I'm still seeing flaky behavior from time to time :-(

stefan-kolb:
@daihua2 Can you please take a look and confirm that this also solves your issue?

@JeremyTheocharis Confirmed to work for my problems and environment. Please also see the logs in the description. Open for any changes if required!

stefan-kolb:
The trigger for this reconnect was caused by https://github.com/gopcua/opcua/issues/908 in our staging system.

JeremyTheocharis:

Issue #805, opened 2025-05-26
Unexpected client disconnection when connecting to Kepware V6


amunozh:
Problem:
Unexpected disconnection after one minute when subscribing to Kepware OPC Server. The disconnection happens one minute after creating the subscription, but only when I subscribe to Kepware Devices with nodes with low frequency changes. If the connection is created to nodes with frequent changes the disconnection does not happen (or at least not yet). 

Logs: 
```
2025-05-26 20:32:07.297 | level=INFO ts=2025-05-26T18:32:07.297579069Z app=device-opcua source=sub-reader.go:42 msg="Subscription created for device: 19" |  
-- | -- | --
  |   | 2025-05-26 20:32:07.297 | debug: publish: notif: 1 |  
  |   | 2025-05-26 20:32:07.297 | debug: publish: pendingAcks=[{"SubscriptionID":2931,"SequenceNumber":1}] |  
  |   | 2025-05-26 20:32:07.297 | debug: publish: PublishRequest: {"RequestHeader":null,"SubscriptionAcknowledgements":[{"SubscriptionID":2931,"SequenceNumber":1}]} |  
  |   | 2025-05-26 20:32:07.298 | debug:  secure_channel_instance.go:89 got sequence number 1012 |  
  |   | 2025-05-26 20:32:07.298 | debug:  secure_channel.go:1053 uasc 4/1011: send *ua.PublishRequest with 74 bytes |  
  |   | 2025-05-26 20:32:07.339 | debug:  conn.go:384 uacp 4: recv MSGF with 18096 bytes |  
  |   | 2025-05-26 20:32:07.339 | debug:  secure_channel.go:350 uasc 4/1011: recv MSGF with 18096 bytes |  
  |   | 2025-05-26 20:32:07.339 | debug:  secure_channel.go:289 uasc 4/1011: recv *ua.PublishResponse |  
  |   | 2025-05-26 20:32:07.339 | debug:  secure_channel.go:304 uasc 4/1011: sending *ua.PublishResponse to handler |  
[...]

PR #899, opened 2026-08-20
client: stop acking notifications the server no longer retains

daihua2:
## Problem

Acknowledging a sequence number the server has already dropped from its retransmission
queue yields `BadSequenceNumberUnknown`. That is harmless per spec, but some servers
(observed with **KEPServerEX V6**) close the TCP connection in response.

Combined with a client that then recreates the subscription this becomes a
self-sustaining loop:

```
recreate subscription -> publish -> stale ack -> server closes TCP -> recreate ...
```

We measured a steady **~12 s period** against a live server, for hours. Nothing surfaced
to the application: `State()` reports `Connected` between the drops and `Read()` works,
so none of the usual liveness checks fire.

Related: #805 (closed as not planned) describes the same symptom against Kepware V6
without identifying the acknowledgement as the trigger.

## Changes

**1. Track `AvailableSequenceNumbers` and skip acks the server no longer retains.**

`Client.availSeqs` records, per subscription, what the most recent `PublishResponse`
reported as still available; `sendPublishRequest` filters the pending acknowledgements
against it. Part 4 defines `AvailableSequenceNumbers` precisely so the client knows what
may still be acknowledged or republished, so this is the conformant behaviour rather
than a workaround.

The filter is deliberately conservative — it only drops when the server reports an
**empty** set for that subscription. "The number is simply not listed" is *not* treated
as proof of absence: a server may list it only in a later response, and dropping on that
basis would leave it unacknowledged forever and let its retransmission queue grow.

**2. Latch `Client.ackRejected` on `BadSequenceNumberUnknown`.**

While set, no acknowledgements are sent at all. This is needed *on top of* (1) because
some servers advertise a sequence number in the `AvailableSequenceNumbers` of the very
response that carries it, then drop it before the acknowledgement arrives — the
advertisement is stale, so filtering on it alone cannot help.

The latch is **per client, not per subscription**: one such acknowledgement is enough to
make those servers drop the connection, so a freshly recreated subscription would send
one again and restart the loop.

It is **self-correcting** — any acknowledgement answered with `StatusOK` clears it, so a
server returning one transient `BadSequenceNumberUnknown` resumes normal acknowledging
on the next successful one.

The `availSeqs` entry is dropped together with the subscription, so the map does not grow
without bound on clients that recreate subscriptions on every reconnect.

## Compatibility

No API change. For a server that never rejects an acknowledgement and never reports an
empty retransmission queue, behaviour is byte-for-byte identical to before: the latch
never engages and the filter never drops anything.

## Testing

`client_sub_ack_test.go` covers:

- `AvailableSequenceNumbers` is **replaced**, not merged, on each response
- acks are **not** dropped merely because a number is absent from a non-empty list
- acks **are** dropped when the server reports an empty set
- acks pass through before any response has been seen for that subscription
- the latch sets on `BadSequenceNumberUnknown`, suspends every ack while set, and clears
  on the next `StatusOK`
- the `availSeqs` entry is released with the subscription

Full package suite passes.

Beyond unit tests this has run in production for a week against KEPServerEX V6 across
three workloads — idle source, ~28k values per 300 s, and switching between the two
(the switch is where the loop reproduced most reliably). Result: zero EOF, zero
`BadSequenceNumberUnknown`, zero internal reconnects.

## Note

This is orthogonal to #895 / #898. Those address the publish loop staying paused after a
reconnect (recovery); this addresses what provokes the disconnect in the first place
(prevention). We run both fixes together — either one alone still leaves a failure mode.


JeremyTheocharis:

PR #900, opened 2026-08-21
fix: NewPrefixLogger to honor a redirected Logger output

stefan-kolb:
NewPrefixLogger hardcoded os.Stderr instead of deriving its writer from the exported Logger, so prefixed loggers (e.g. the client's 'client: monitor: ' dlog) silently ignored any redirection of debug output, even though Logger is an exported *log.Logger and can already be redirected via Logger.SetOutput(w) or replaced entirely. Now NewPrefixLogger uses Logger.Writer(), so all debug output (Printf, ToJSON, and NewPrefixLogger loggers) consistently honors whatever destination Logger currently points to.

Specific use case for that is for our production setup and for debugging #895, where I needed to redirect to another log writer, i.e., syslog in my case to collect all log messages from the gopcua library.

### DIFF
--- a/debug/debug.go
+++ b/debug/debug.go
@@ -25,6 +25,13 @@ var Flags = os.Getenv("OPC_DEBUG")
 var Enable bool = FlagSet("debug")
 
 // Logger logs the debug messages when debug logging is enabled.
+//
+// It defaults to writing to os.Stderr but, since it is an exported
+// *log.Logger, applications can redirect it at runtime with
+// Logger.SetOutput(w) (or replace it entirely, e.g. Logger = log.New(...)),
+// for example to route opcua debug logs through the app's own logging
+// framework. NewPrefixLogger honors the same destination since it derives
+// its writer from Logger.Writer().
 var Logger = log.New(os.Stderr, "debug: ", 0)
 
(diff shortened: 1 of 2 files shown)


stefan-kolb:
@JeremyTheocharis Tested and ready for review.

JeremyTheocharis:
thank you <3

JeremyTheocharis (approved):
Good catch — `Printf` honoured `Logger` and `NewPrefixLogger` didn't, which made half the debug output unredirectable. Verified: reverting your line turns `TestLoggerSetOutputRedirectsPrefixLogger` red.

I pushed two things on top. A merge of main, since #873 landed and the branch was behind. And a nil guard: `Logger` is exported and assignable, so `Logger = nil` used to write to os.Stderr and would now panic — it falls back to os.Stderr again, with a test that gates it. Only reachable with debug logging on, since the `!Enable` early return comes first.

One thing I left alone, your call rather than mine: `Logger.Writer()` is read at construction, so a `SetOutput` after the client is built doesn't reach an existing prefix logger. That's 2 of the 9 call sites — `client: monitor: ` and `sub: `, both built once when the goroutines start. The other 7 are rebuilt per operation and do pick up a later redirect. Worth a doc line or a lazy lookup if you think it matters.

PR #903, opened 2026-09-03
uasc: stop discarding errors the server sends

### LINKED ISSUE #613: Why does it always return EOF when receiving an ERR type of return packet？
Obviously the server did not disconnect the connection, but just returned an error.
I can't even tell from the log what type of error it is
Is it necessary to judge the length equal to 0 in readChunk?
```
	// read a full message from the underlying conn.
	b, err := s.c.Receive()
	if err == io.EOF || len(b) == 0 {
		return nil, io.EOF
	}
```

```
debug: uacp: connecting to opc.tcp://BA-SERVER1
debug: uacp 3: start HEL/ACK handshake
debug: uacp 3: sent HELF with 52 bytes
debug: uacp 3: recv ACKF with 28 bytes
debug: uacp 3: server has no chunk limit. Using 512
debug: uacp 3: recv &uacp.Acknowledge{Version:0x0, ReceiveBufSize:0xffff, SendBufSize:0xffff, MaxMessageSize:0x1000000, MaxChunkCount:0x200}
debug: uasc 3/1: send *ua.OpenSecureChannelRequest with 132 bytes
debug: uacp 3: recv OPNF with 135 bytes
debug: uasc 3/1: recv OPNF with 135 bytes
debug: uasc 3/1: recv *ua.OpenSecureChannelResponse
debug: uasc 3/1: sending *ua.OpenSecureChannelResponse to handler
debug: uasc 3: received security token. channelID=14 tokenID=1 createdAt=2022-10-27T03:44:34Z lifetime=1h0m0s
debug: uasc 3: security token expires at 2022-10-27T04:59:34Z. channelID=14 tokenID=1
debug: uasc 3: security token is refreshed at 2022-10-27T04:28:05Z (45m0s). channelID=14 tokenID=1
debug: uasc 3/2: send *ua.CreateSessionRequest with 259 bytes
debug: uacp 3: recv ERRF with 16 bytes
debug: uasc 3: readChunk EOF
debug: uasc 3: Close()
debug: uacp 3: close
```

This is the information in wireshark
image

mjk-automation-ltd:
readChunk tested len(b) == 0 before the *uacp.Error branch, and Receive returns (nil, err) when it decodes an ERR message - so every error a server sent became a bare io.EOF and the branch written to preserve it was unreachable.

The dispatcher then dropped what survived: an ERR answering OpenSecureChannel arrives with request id 0 while the pending request is id 1, so popHandler missed and the status was discarded. The caller waited for a reply the server had already refused to send, until the socket closed.

A PLC that does not trust the client certificate answers with BadSecurityChecksFailed. Before this, that reached the caller as EOF.

Fixes #613


JeremyTheocharis:

PR #904, opened 2026-09-03
uasc: a service fault is the caller's answer, not a channel loss

dkb5191:
## What

A response whose header carries a Bad `ServiceResult` is delivered to the request's handler as before, but it reaches the client's reconnect monitor only when the status says the secure channel or the session is gone.

## Why

`uasc.SecureChannel.Receive` sets `msg.Err` to the header's `ServiceResult` for any response that is not Good, and `dispatcher` pushes that onto the client's error channel **before** it hands the response to the request's handler. `Client.monitor` skips only `Bad_NoSubscription`; with `AutoReconnect` on (the default) everything else lands in the `default` arm, `createSecureChannel`, which calls `conn.Close()` right there and dials again. Meanwhile `sendRequestWithTimeout` selects between the handler channel and `disconnected`, so the caller whose request drew the fault reads whichever Go picks.

So a server refusing **one request** on an open channel — `Bad_TooManyOperations`, `Bad_ServiceUnsupported` from a server lacking a service set, `Bad_UserAccessDenied` — tears the secure channel down, pauses and republishes every subscription on it, flips the `StateChangedCh` through `Disconnected`/`Reconnecting`, and hands the caller `io.EOF` some of the time instead of the status it was sent. We measured about one run in 150 against an in-process gopcua server answering `Call` with its default `ServiceFault`:

```
uasc 801/11: err: The server does not support the requested service. StatusBadServiceUnsupported (0x800B0000)
uasc 801/11: sending *ua.ServiceFault to handler
client: monitor: disconnected
client: monitor: auto-reconnecting
client: monitor: action: createSecureChannel
uacp 801: close
uasc 801: readChunk EOF
    call: EOF
```

## How

`MessageBody` records whether `Err` came from a response header (`serviceFault`). `dispatcher` forwards such an error to the monitor only when `channelLost(status)` holds: `Bad_SecureChannelIdInvalid`, `Bad_SecureChannelClosed`, `Bad_SecureChannelTokenUnknown`, `Bad_SessionIdInvalid`, `Bad_SessionClosed`, `Bad_SessionNotActivated`, `Bad_SubscriptionIdInvalid`, `Bad_CertificateInvalid`, `Bad_ConnectionClosed` — the statuses the monitor already has explicit arms for, plus the closed/unknown variants. Transport errors, decoding errors and UACP `ERR` messages are not response headers and reach the monitor exactly as before.

The fix sits in the dispatcher rather than the monitor because the monitor cannot tell a header status from a UACP `ERR` status, and an `ERR` is a channel-level message the server closes the socket after.

## Tests

`uasc/service_fault_test.go` covers the classification. Downstream, an end-to-end test against a gopcua server answering `Call` and `Browse` with a `ServiceFault` behind a connection-counting TCP relay fails 3/3 on v0.8.0 (`write tcp …: use of closed network connection` under the next request) and passes 10/10 with this change, with a subscription on the same connection still publishing afterwards.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Ate5MW1ccYPYM87sCRqKAD


JeremyTheocharis:

PR #907, opened 2026-09-14
fix restoreSession leaving the client connected with dead subscriptions

### LINKED ISSUE #879: restoreSession leaves client Connected with 0 active subscriptions (subsToRepublish/subsToRecreate never populated)
## What happens

After a transport-level disconnect where the reconnect flow takes the **`restoreSession`** path (session reactivated rather than recreated — the common case for a brief network blip where the server-side session survives), the client ends up reporting state `Connected` but silently has zero active subscriptions and a permanently dead publish loop. No further data is ever delivered and no further reconnection attempt is made, since the client believes it's healthy.

This is the same general failure family as #854 and #876 (a reconnect path that silently abandons subscriptions), but a distinct root cause from both.

## Root cause

In `client.go`, the `restoreSession` case, on success, goes straight to:

```go
action = restoreSubscriptions
```

(around line 469), without ever populating `subsToRepublish` / `subsToRecreate`.

Compare this to the sibling `transferSubscriptions` case (around lines 513-549), which explicitly resets and populates both slices before also transitioning to `restoreSubscriptions`:

```go
availableSeqs = map[uint32][]uint32{}
subsToRecreate = nil
subsToRepublish = nil

res, err := c.transferSubscriptions(ctx, subIDs)
switch {
case errors.Is(err, ua.StatusBadServiceUnsupported):
    subsToRepublish = nil
    subsToRecreate = subIDs
case err != nil:
    subsToRepublish = nil
    subsToRecreate = subIDs
default:
    for i := range res.Results {
        // ... populates subsToRecreate / subsToRepublish per result
    }
}
action = restoreSubscriptions
```

`restoreSubscriptions` itself documents this assumption in a comment (around lines 558-559):

```go
// try to republish the previous subscriptions from the server
// otherwise restore them.
// Assume that subsToRecreate and subsToRepublish have been
// populated in the previous step.

activeSubs = 0
for _, subID := range subsToRepublish {
    ...
    activeSubs++
}
for _, subID := range subsToRecreate {
    ...
    activeSubs++
}

c.setState(ctx, Connected)
action = none
```

When arriving via `restoreSession`, that assumption is violated: both slices are empty (or stale from a previous iteration), so both loops execute zero times, `activeSubs` stays `0`, and the code falls straight through to `c.setState(ctx, Connected)` / `action = none` regardless — marking the client healthy while the publish loop has nothing to resume.

## Steps to reproduce

1. Connect a client with `AutoReconnect` enabled (the default) and create a subscription with at least one monitored item receiving live data-change notifications.
2. Simulate a transport-level reset that the *server-side session survives* — e.g. force a TCP RST on the client's connection (via a proxy, `SO_LINGER=0`, etc.) rather than restarting the OPC-UA server process itself. A full server restart correctly takes the `recreateSession` → `transferSubscriptions` path and is unaffected by this bug.
3. Observe via debug logging (`dlog`/`gdebug.Enable = true`) that the reconnect takes the `restoreSession` path, logs "session restored" and "namespaces updated", then "action: restoreSubscriptions".
4. Observe that `resumeSubscriptions` is never meaningfully invoked with live subscriptions (or that no further `PublishResponse` notifications ever arrive), even though the client's connection state is `Connected`.

## Suggested fix

Populate `subsToRepublish` / `subsToRecreate` before leaving the `restoreSession` case — e.g. `subsToRecreate = c.SubscriptionIDs()` (recreate-only, since republish/AvailableSequenceNumbers info isn't available on this path the way `TransferSubscriptions` provides it) — or simply have `restoreSession` fall through to `transferSubscriptions` instead of jumping straight to `restoreSubscriptions`, so the existing population logic there is reused.

## Environment

gopcua v0.9.0

cpruijsen:
Fixes #879.

On the `restoreSession` path, `monitor` in `client.go` moved to `restoreSubscriptions` with
`subsToRepublish` and `subsToRecreate` both still empty, so `republishOrRecreateSubscriptions`
iterated two empty slices, returned `none`, and the loop set the client `Connected`. Nothing was
republished, nothing was recreated, and the publish loop stayed paused. The client reports healthy,
delivers no data, and never reconnects again because it believes there is nothing wrong.

The sibling `transferSubscriptions` case populates both slices before the same transition. This
one populates `subsToRepublish` with the registered subscription ids:

```go
subsToRepublish = c.SubscriptionIDs()
```

Republish rather than recreate, because the distinction is the point of the `restoreSession` path:
the server-side session survived the blip and still owns its subscriptions, so republishing also
recovers the notifications missed while the connection was down. Recreating would silently drop
them. `subsToRecreate` stays nil deliberately;
`republishOrRecreateSubscriptions` already appends a failed republish to it and recreates that
subscription, so a subscription the server did drop still comes back, one round later, through the
existing fallback rather than through a new branch here.

Same family as #854 and #876, and the third distinct cause in it, after #855 fixed this path
closing the session it had just reactivated.

**No test.** The defect is in `monitor`'s state machine, which the unit harness in
`client_sub_test.go` cannot reach: that harness drives `republishOrRecreateSubscriptions` directly
through `stubClient`, and both functions this fix composes are already covered there. A test that
fails without this change has to observe the reconnect loop itself, which needs the live server the
`tests/go` suite uses. #855 landed the same shape, `client.go` only and no test, for the same
reason.

`go build ./...`, `go vet ./...` and `go test ./...` are clean: 11 packages, no failures. The two
`ua/buffer.go` vet warnings are on `main` already and are untouched here.


aschiffmann (approved):
We hit this on **gopcua v0.9.1** against a **Siemens OPC UA** server (AutoReconnect, one subscription with live data-change items).

## Repro of #879

After a transport-level `StatusBadTimeout` (~30s idle Publish), reconnect took `restoreSession`. The session came back; `subsToRepublish` / `subsToRecreate` were still empty; the client went `Connected` and **did not** resume the publish loop. Notifications stopped until process restart. Same sequence on two fresh processes.

```
sechanErr StatusBadTimeout (0x800A0000) state=Connected nsubs=1 session=true
restoreSession client_subs=1 republish=0 recreate=0
restoreSubscriptions client_subs=1 republish=0 recreate=0
NOT resuming subscriptions activeSubs=0 client_subs=1 state=Connected
```

That matches the issue: `restoreSession` jumps to `restoreSubscriptions` without populating the slices.

## This PR

We applied the same change as here — after a successful restore, `subsToRepublish = c.SubscriptionIDs()` — and left AutoReconnect to take that path on the next timeout.

```
debug: client: monitor: action: restoreSession
debug: client: monitor: session restored
debug: client: monitor: namespaces updated
         restoreSession filled subsToRepublish=1 client_subs=1
         restoreSubscriptions client_subs=1 republish=1 recreate=0
         republish sub=<id> seqs=0 err=<nil>
debug: client: monitor: resuming 1 subscriptions
debug: client: monitor: resumed 1 subscriptions
```

- Republish succeeded with **empty** `availableSeqs` (`err=<nil>`).
- `resumeSubscriptions` ran with `activeSubs=1`.
- Data-change notifications continued after the reconnect (including a later process that completed while one of these restores ran).

This looks correct for the `restoreSession` hole. Please merge.

The reconnect itself is the idle-Publish `BadTimeout` → `sechanErr` issue — [#908](https://github.com/gopcua/opcua/issues/908). Separate; does not block this PR.


[gopcua-pr907-logs.txt](https://github.com/user-attachments/files/32303038/gopcua-pr907-logs.txt)

stefan-kolb:
Isn't that the same problem I fixed in https://github.com/gopcua/opcua/pull/898?
Just adding the subscriptions like in this PR did not work for me, if I remember correctly (see https://github.com/gopcua/opcua/pull/898/changes#diff-4b667feae66c9d46b21b9ecc19e8958cf4472d162ce0a47ac3e8386af8bbd8cfR478-R488).
I tested the behavior in our faulty staging system.
**In my case the reason for the reconnect also were `StatusBadTimeout` responses, see https://github.com/gopcua/opcua/issues/908**

cpruijsen:
Different fix, though they overlap and yours is the bigger one. #898 does two
things: routes `restoreSession` through `transferSubscriptions`, and replaces
`pausech`/`resumech` with `publishPaused`, which is #895.

I think the second one is why adding the ids alone didn't work for you. On a
reconnect `monitor` pauses at `client.go:375` and resumes at `client.go:606`,
and the publish loop queues its own pause from the error path at
`client_sub.go:452`. If the loop was inside `publish()` while that happened, it
comes back to the outer `select` with a pause and a resume both queued, and
`case <-c.resumech` there is the "ignore since not paused" branch, which throws
it away. The pause then parks the loop for good.

Queuing that sequence on main, the resume is swallowed 146 times out of 200. So
with only my change applied it would look unchanged in about three runs out of
four, which matches what you saw.

@aschiffmann did get this working on a Siemens server though, republish
returning `err=<nil>` and the loop resuming. And `availableSeqs` only
drives a log line in `sendRepublishRequests`, the republish walks from
`sub.nextSeq` either way.

No strong view on which lands. If #898 goes in whole this is redundant. I'd just
rather the #895 part not wait on the `transferSubscriptions` question, since it
eats the resume on every reconnect path.

stefan-kolb:
I'm not 100% sure in which state I tested the simpler solution similar to yours, but I think it already included the fix for #895.
But hey, I'm only a human 😄

JeremyTheocharis:

JeremyTheocharis:
Anything on the overall discussion in #910?

